PR #16, opened 2026-08-13
feat(telemetry): add the OpenTelemetry pipeline and core Berd events

matt2e:
## Summary

Berd had no product analytics. This adds the full renderer-to-gateway telemetry path plus the first catalog of Berd events, instrumented at the points where each user action actually commits rather than where it is merely requested.

**Pipeline**

- An OTel logs pipeline in `src/shared/telemetry` (`client.ts`, `exporter.ts`, `identity.ts`) batches records and hands the serialized OTLP/HTTP JSON body to the Tauri layer, so no renderer code holds a credential or an endpoint.
- `src-tauri/src/commands/telemetry.rs` owns auth and transport: anonymous `POST /v1/bootstrap` with a validated installation id, token TTL read from the response, a single 401 re-auth retry, and `POST /v1/logs` carrying `x-berd-schema-version: berd-otlp-logs-v1` against the gateway's `berd` service and `berd.telemetry` scope.
- The client stays inert unless enabled: `development` never reaches the wire, startup gates are bounded so a stalled bootstrap cannot wedge the app, dropped records are counted, and batch size and attribute lengths are capped under the gateway's body limit. Empty optional attributes are omitted rather than sent as empty values.
- The staging/prod gateway hosts are not decided yet, so the endpoint defaults to a dummy `.invalid` host in the real endpoint's shape. Swapping in a real host is pure configuration — `VITE_OTLP_LOGS_ENDPOINT`, `ALLOWED_OTEL_LOGS_HOSTS`, the CSP `connect-src`, and the pinned test values — with no code-path changes (see the `TODO(telemetry-otel)` in `vite.config.ts`).

**Events**

`src/shared/telemetry/events` has one module per surface, covering app launch, chat sends, agent edits, home pins, and project mutations. Each is anchored to its commit point:

- App launch fires once per app start, not per reload; detached session windows initialize telemetry but do not report a launch.
- Chat send events fire at the user-message commit, including steer sends at the steer commit and released deferred-workspace sends; `is_first_message` is gated on the session history having replayed.
- Agent `Edit Completed` is anchored to the persisted persona write, and agent creation is counted from the duplicate, import, and onboarding surfaces.
- Pin/unpin reports only changes that survive persistence, uses the id the pin will keep, emits per item on bulk unpin, and stays silent for starter-agent seeding and the legacy-Berdy pin migration.
- berdctl project mutations are deliberately excluded; the exclusion is documented at the command modules.

Ported from `squareup/berd@5e28e6e1`, adapted to this repository's drift (worktree startup modes, pre-send workspace setup) with the telemetry behavior and wire contract intact.

### Related issue

None found — searched open and closed issues/PRs for `telemetry`, no duplicates.

### Testing

Tests accompany each surface, pinning the wire body contract (`exporter.test.ts`, `client.test.ts`), the event attribute shapes (`events.test.ts`, `homePinTelemetry.test.ts`), and the commit-point behavior (`useChatSessionController.test.ts`, `AgentBuilderRail.test.tsx`, `AgentsView.entry.test.tsx`, `usePinToHomeWidget.test.tsx`, `HomeView.test.tsx`, `OnboardingFlow.test.tsx`, `main.test.tsx`).

Relevant gates for a reviewer to run: `just check`, `just test`, `just tauri-check`, `just clippy`. No user-visible UI change, so no screenshots.


kalvinnchau (dismissed):

src-tauri/src/commands/telemetry.rs
```diff
@@ -0,0 +1,1251 @@
...
+        )
+    })?;
+    let body = serde_json::to_string_pretty(&PersistedTelemetrySettings {
+        schema_version: TELEMETRY_SETTINGS_SCHEMA_VERSION,
+        enabled,
+    })
+    .map_err(|error| format!("Failed to serialize telemetry settings: {error}"))?;
+    let tmp = path.with_extension("tmp");
+    fs::write(&tmp, body).map_err(|error| format!("Failed to write {}: {error}", tmp.display()))?;
+    fs::rename(&tmp, &path)
```
morgmart, line 582: 🤖 **P1 · Consent cannot be changed twice on Windows** (blocking)

The consent writer always renames a temporary file over telemetry-settings.json. On Windows, std::fs::rename does not replace an existing destination, so the first write can create the file but a later write fails once that destination exists. The in-memory AtomicBool is updated only after persistence succeeds, leaving the previous consent effective.

**User effect:** A Windows user who opts in cannot later turn telemetry off from Settings (and a user who first persists off cannot later opt in); the switch reports a save error and the prior choice remains active.

**Recommended fix:** Use a cross-platform atomic replacement implementation that safely replaces an existing destination while preserving fail-closed crash behavior, then update the cached consent only after that replacement succeeds.

**Test:** Add a Windows-capable native test that calls set_consent(true) and then set_consent(false) on the same state/path, asserts both calls succeed, and verifies a fresh TelemetryAuthState reads false.

src-tauri/src/commands/telemetry.rs
```diff
@@ -0,0 +1,1251 @@
...
+/// telemetry off: they reach this command, and stop here.
+async fn export_otel_logs_for_state(
+    auth: &TelemetryAuthState,
+    endpoint: &str,
+    body: String,
+) -> Result<OtelLogsExportResponse, String> {
+    if !auth.consent_granted() {
+        return Err("Telemetry is disabled for this installation".to_string());
+    }
+    let logs_url = allowed_otel_logs_endpoint(endpoint)?;
```
morgmart, line 384: 🤖 **P1 · Revocation does not stop pending exports** (blocking)

Consent is checked only once before the async bootstrap/upload sequence. If consent is revoked while an export is waiting for bootstrap or between the first 401 response and re-authentication, the sequence continues to bootstrap and post the queued body without consulting the now-disabled AtomicBool again.

**User effect:** After the Settings switch has confirmed telemetry is off, an already-started batch can still send the installation identifier and telemetry records, violating the user's revocation expectation.

**Recommended fix:** Carry a consent generation or cancellation signal through the export flow and re-check it immediately before bootstrap, each logs POST, and the 401 retry; abort without network activity whenever revocation supersedes the operation.

**Test:** Add an async native test whose bootstrap is paused, revoke consent while it is pending, release it, and assert that no /v1/logs request occurs; add the equivalent race around the 401 re-auth retry.

src-tauri/src/commands/telemetry.rs
```diff
@@ -0,0 +1,1251 @@
...
+/// `MAX_LOG_EXPORT_BATCH_SIZE` in `src/shared/telemetry/client.ts`) so a full
+/// batch stays under the gateway's request-body limit, which it would otherwise
+/// answer with a 413 that costs the whole batch. The body is sent as plain
+/// uncompressed JSON: the gateway parses the raw bytes, so any
+/// `content-encoding` would come back a terminal 400.
+#[tauri::command]
+pub async fn export_otel_logs(
+    state: tauri::State<'_, TelemetryAuthState>,
+    endpoint: String,
+    body: String,
```
morgmart, line 352: 🤖 **P2 · Native export accepts unbounded bodies** (non-blocking)

The Tauri command accepts an arbitrary renderer-provided String and passes it into a path that clones the full body for a possible retry. Renderer batching limits are producer-side only and do not protect this native ingress boundary from a malformed or compromised caller.

**User effect:** A bad renderer call can force the desktop process to allocate multiple copies of a very large payload and attempt an oversized upload, causing avoidable memory pressure or instability.

**Recommended fix:** Define the gateway request-body limit in the native layer and reject body.as_bytes().len() above it before endpoint validation, token bootstrap, or cloning.

**Test:** Add a native command-boundary test that supplies a body just above the limit and proves it is rejected before any bootstrap or logs request, plus a boundary-sized accepted case.

src-tauri/src/commands/telemetry.rs
```diff
@@ -0,0 +1,1251 @@
...
+/// `/v1/logs` path convention the bootstrap derivation relies on.
+fn allowed_otel_logs_endpoint(raw_url: &str) -> Result<reqwest::Url, String> {
+    let url = reqwest::Url::parse(raw_url)
+        .map_err(|error| format!("Invalid OTel logs endpoint {raw_url}: {error}"))?;
+
+    if url.scheme() != "https" {
+        return Err(format!("OTel logs endpoint must use https: {raw_url}"));
+    }
+
+    match url.host_str() {
```
morgmart, line 622: 🤖 **P2 · Gateway allowlist permits alternate ports** (non-blocking)

Endpoint validation checks HTTPS and host_str() but does not constrain the effective port. A handcrafted renderer invocation can therefore target an allowlisted hostname on any explicit port, and the native layer will send bootstrap traffic and bearer-authenticated logs there.

**User effect:** Telemetry credentials and records can be directed to a different service listening on the allowlisted host instead of the intended HTTPS gateway endpoint.

**Recommended fix:** Validate the complete authority by rejecting explicit non-443 ports (or allowlisting exact scheme/host/port tuples) before deriving the bootstrap URL.

**Test:** Add endpoint-validation cases proving :443 is accepted and an alternate explicit port such as :8443 is rejected without any network request.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for the exact three-dot comparison bec8c6ed35cac7a14c5c6b1ea9a5a844ddb648a3...eac38e8d4c37308028d06b92c306a578de49a9b4. Two blocking consent/boundary defects and two non-blocking native-boundary hardening issues are publishable. The supplied GitHub evidence was inspected: DCO, frontend checks/tests, frontend build smoke, clean-room install, transcript virtualization, and Tauri Rust checks on macOS/Windows/Linux all report success; required checks still independently govern merge readiness. Final self-check covered every changed telemetry user flow and instrumentation surface, accessibility, English/Spanish localization, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, project laws and design-system rules, duplicate overlap, changed-line anchors, and concrete user effect. No supplied review threads exist, and overlapping boundary concerns were consolidated.

src-tauri/src/commands/telemetry.rs
matt2e: Skipping this

src/features/chat/hooks/useChatSessionController.ts
```diff
@@ -1882,6 +1890,90 @@ export function useChatSessionController({
...
+        // the transcript, so "first" means it is the only user message there —
+        // and only once the session's history has landed, since an unreplayed
+        // old session shows the same empty transcript (see chatFirstMessage).
+        const isFirstMessage = isFirstCommittedUserMessage(sessionId);
+        // An override with `id: null` is an explicit "send without a persona";
+        // no override falls back to the session's selected persona.
+        const hasPersona = overridePersona
+          ? overridePersona.id !== null
+          : Boolean(selectedPersonaId);
+        const provider = selectedProvider;
```
morgmart, line 1942: 🤖 **P2 · Chat telemetry ignores the dispatched target** (non-blocking)

Foreground send telemetry derives provider and model from live controller selection state at commit time. The queued attempt itself dispatches with the authoritative target captured in sendOptions.sessionSelection, and a later provider/model selection can update pendingExecutionTarget and selectedProvider while the admitted attempt is still preparing. The event can therefore report a different target than the message actually used.

**User effect:** Analytics can attribute a successfully sent message to the wrong provider or model, making provider/model adoption and reliability data misleading even though the chat itself uses the correct captured target.

**Recommended fix:** Pass the active attempt's captured sessionSelection into the commit telemetry callback and derive both provider and model from that single authoritative target, falling back to the committed session target only when no attempt snapshot exists.

**Test:** Add a controller test that captures a queued attempt on provider/model A, changes the pending selection to provider/model B before invoking onUserMessageCommitted, and asserts both chat events report A only.

src/features/settings/ui/SystemSettings.tsx
```diff
@@ -349,6 +350,10 @@ export function SystemSettings() {
           ) : null}
         </SettingsSection>
 
+        {/* Renders nothing in enforced builds, where telemetry consent is
+          build policy and not a user choice. */}
+        <TelemetryConsentRow />
```
morgmart, line 356: 🤖 **P2 · Unavailable telemetry still shows consent** (non-blocking)

System Settings always renders TelemetryConsentRow unless telemetry is build-enforced, but the actual send path is also gated by the reactive telemetry profile capability. In builds compiled with telemetry disabled, or after runtime configuration disables the telemetry toggle, the row remains visible and can persist enabled consent even though no event can emit. Search likewise hides this entry only for enforced builds rather than when the capability is unavailable.

**User effect:** Users can turn on a 'Share usage data' control that has no effect, leaving Settings to claim sharing is enabled while the product is incapable of sending telemetry.

**Recommended fix:** Gate both the Settings row and its search item on the reactive telemetry profile capability in addition to the enforced-build rule, while preserving the persisted preference for when the capability becomes available again.

**Test:** Add Settings and search tests with the telemetry capability false that assert neither the row nor search result is exposed, plus a capability-true opt-in case.

morgmart (commented):
🤖 Automated code review

COMMENT. Fresh static review completed for the exact three-dot comparison bec8c6ed35cac7a14c5c6b1ea9a5a844ddb648a3...8d88edf4cd376455f58eb1880937a6b6408d327b. Two new non-blocking findings are publishable. One still-present native-boundary concern is suppressed forever because its resolved authoritative thread contains a substantive human deferral. The prior Windows replacement, mid-flight revocation, and alternate-port issues are fixed at this SHA and are not findings. Supplied GitHub evidence was inspected: DCO, frontend checks/tests, frontend build smoke, clean-room install, transcript virtualization, and Tauri Rust checks on Windows and macOS passed; Tauri Rust checks on Linux failed. CI remains separate from the code verdict and required checks independently govern merge readiness. Final self-check covered every changed app-launch, consent, native export, chat send/steer/deferred-send, agent, project, Home pin, onboarding, search, feedback, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and architectural-law compliance; duplicate overlap; current changed-line anchors; and concrete evidence plus user effect for severity. Unsupported and overlapping claims were removed.

Deterministic publication result: 0 blocking and 2 non-blocking finding(s) publishable; 1 duplicate(s) suppressed.

morgmart (dismissed):

src-tauri/Cargo.toml
```diff
@@ -133,6 +147,11 @@ block-automations = []
 block-builderbot = []
 block-feedback = []
 block-managed-connections = []
+# Forces telemetry consent ON, skipping the user's telemetry-settings.json
+# (the renderer hides the toggle via the paired VITE_TELEMETRY_ENFORCED flag).
+# For managed internal distributions where consent is an employment-policy
+# fact, not a per-user choice.
+block-telemetry-enforced = []
```
morgmart, line 155: 🤖 **P1 · Windows omits enforced telemetry feature** (blocking)

This PR adds the native block-telemetry-enforced Cargo feature and pairs it with VITE_TELEMETRY_ENFORCED in the renderer and Unix/macOS build paths, but the Windows Get-BerdAppFeatures mapping still omits that environment gate. A Windows build with VITE_TELEMETRY_ENFORCED=1 therefore hides the consent control and treats consent as enforced in the renderer while compiling the native export gate without the matching feature, where consent remains default-off unless a settings file independently enabled it.

**User effect:** Managed Windows distributions that require telemetry can ship with no consent control yet reject all telemetry exports, silently losing the product analytics the distribution policy says are enabled.

**Recommended fix:** Add VITE_TELEMETRY_ENFORCED to Get-BerdAppFeatures with the block-telemetry-enforced Cargo feature, update its stale gate-count documentation, and keep the Windows bundle/check/dev feature resolution in lockstep with scripts/block-feature-gates.sh and build-macos.sh.

**Test:** Extend the Windows feature-resolution tests to set VITE_TELEMETRY_ENFORCED=1 and assert block-telemetry-enforced is included, then assert the default remains absent and invalid values are rejected.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for the exact three-dot comparison 64979a007ac78c0bf06c8c2664e819da585638db...0f7f0efd023e0a416c4f92d6c7168fd18af65b2d. One new blocking cross-platform build-integration finding is publishable. Two still-present concerns are suppressed by authoritative thread memory: one unresolved automated thread and one resolved thread with a substantive human deferral. The prior unavailable-consent UI issue is fixed at this SHA; the Windows settings replacement, mid-flight revocation, and alternate-port issues also remain fixed. Supplied GitHub evidence was inspected and all listed checks report success: DCO, frontend checks/tests, frontend build smoke, clean-room install, transcript virtualization, and Tauri Rust checks on Linux, Windows, and macOS. CI remains separate from the code verdict and required checks independently govern merge readiness. Final self-check covered every changed app-launch, consent, native export, chat send/steer/deferred-send, agent, project, Home pin, onboarding, search, feedback, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; release/build integration; design-system and architectural-law compliance; duplicate overlap; current changed-line anchors; and concrete evidence plus user effect for blocking severity. The documented startup limitation of the runtime telemetry toggle was treated as an explicit tradeoff rather than invented as a finding.

Deterministic publication result: 1 blocking and 0 non-blocking finding(s) publishable; 2 duplicate(s) suppressed.

torstenrudolf (approved):

torstenrudolf:
🤖 Follow-up review at `0f7f0ef`: I don’t think this is ready to merge yet. All current checks are green, but these concerns remain:

1. **P1 — enforced telemetry can silently send nothing.** `VITE_TELEMETRY_ENFORCED=1` is not mapped to the native `block-telemetry-enforced` feature in the [Windows feature mapper](https://github.com/block/berd/blob/0f7f0efd023e0a416c4f92d6c7168fd18af65b2d/scripts/windows/WindowsDev.psm1#L483-L514). The same drift exists in the hand-built Unix [`just bundle`](https://github.com/block/berd/blob/0f7f0efd023e0a416c4f92d6c7168fd18af65b2d/justfile#L344-L354) and [`bundle-debug`](https://github.com/block/berd/blob/0f7f0efd023e0a416c4f92d6c7168fd18af65b2d/justfile#L434-L444) feature lists. Those builds can hide the consent control while native consent remains default-off, so every export is rejected unless an enabled settings file already exists. Please map the flag on Windows and include the Cargo feature in both Unix recipes.

2. **P2 — the 256 KiB body guarantee is not UTF-8 safe.** The [sizing test](https://github.com/block/berd/blob/0f7f0efd023e0a416c4f92d6c7168fd18af65b2d/src/shared/telemetry/exporter.test.ts#L418-L478) uses ASCII only, while the cap is in JavaScript characters rather than bytes. With this PR’s pinned OpenTelemetry versions, an exact serializer probe for 128 `berd_chat_session_started` records with 256-character CJK provider/model values produced **264,667 bytes**, above the configured **262,144-byte** gateway limit. That batch gets a 413 and is permanently dropped. Lowering the batch limit to 64, or enforcing UTF-8 byte limits, plus a multibyte worst-case test would close the gap.

3. **P2 — foreground chat attribution still reads live selection state.** The telemetry callback uses the controller’s current provider/model instead of the queued attempt’s captured `sessionSelection`. A selection change during preparation can therefore misattribute the committed send. The existing thread has the concrete fix context: https://github.com/block/berd/pull/16#discussion_r3801155392

morgmart (commented):
🤖 Automated code review

NO NEW PUBLICATION. Fresh static review completed for the exact three-dot comparison 64979a007ac78c0bf06c8c2664e819da585638db...4324a48c82ed3df2aa78193767f66a014d2492c7 with zero new publishable findings. One still-present issue is suppressed because its authoritative automated thread remains unresolved, so approval is not recommended; one native-boundary concern is suppressed forever because its resolved thread contains a substantive human deferral. The prior Windows feature-gate mismatch is fixed at this SHA, as are the unavailable-consent UI, Windows settings replacement, mid-flight revocation, and alternate-port issues. Supplied GitHub evidence was inspected: clean-room install and DCO passed; frontend checks/tests and Tauri Rust checks on Linux, macOS, and Windows were in progress. Pending checks do not make validation incomplete, and required checks independently govern merge readiness. Final self-check covered every changed app-launch, consent, native export, chat send/steer/deferred-send, agent, project, Home pin, onboarding, search, feedback, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; release/build integration; design-system and architectural-law compliance; duplicate overlap; current changed-line anchors; and concrete evidence plus user effect for blocking severity. No unsupported finding was added merely to produce output.

src/app/AppShell.tsx
```diff
@@ -4945,10 +4948,7 @@ export function AppShell({
     );
   }
 
-  const shouldShowOnboarding =
-    onboardingExperiment?.enabled === true &&
-    onboardingState.lifecycle !== "completed";
-  if (shouldShowOnboarding) {
+  if (onboardingState.lifecycle !== "completed") {
```
morgmart: 🤖 **P1 · Welcome waits for agent startup** (blocking)

The experiment gate is removed so incomplete onboarding is now the default fresh-install route, but AppShell still evaluates the full startup loading and diagnostic gates before rendering OnboardingFlow. That startup includes ACP/provider runtime initialization and can remain pending or fail before the user has configured anything; the new landing experience therefore cannot serve as the first usable surface independently of the agent runtime it is meant to help configure.

**User effect:** A first-time user can be left indefinitely on a startup spinner, or shown a technical startup diagnostic, without ever reaching the welcome and setup flow that should help them get Berd working.

**Recommended fix:** Give incomplete onboarding an early startup route that can render the welcome and non-runtime-dependent steps before the full agent runtime is ready, while gating only the later steps that genuinely require runtime state and retaining an explicit retry/error path there.

**Test:** Add an AppShell integration test with incomplete onboarding and a never-resolving or failed agent startup, and assert the welcome screen remains reachable while runtime-dependent actions communicate their pending/error state.

src/main.tsx
```diff
@@ -161,30 +169,56 @@ if (bootError) {
...
     </React.StrictMode>,
   );
+  getInstallationCohort()
+    .then((cohort) => {
+      initializeOnboardingGraduation(cohort);
+    })
+    .catch((error) => {
+      console.error("Failed to resolve installation cohort:", error);
+      reportRendererError("installation_cohort_failed", error);
+      initializeOnboardingGraduation("unknown");
```
morgmart, line 197: 🤖 **P1 · Unknown cohort becomes first run** (blocking)

A cohort lookup timeout or command failure is converted to the unknown cohort, and initializeOnboardingGraduation graduates only the positively established cohort. Malformed, unreadable, or future-version native markers also resolve to unknown. With no valid local onboarding record, unknown therefore retains the initial incomplete state and the now-unconditionally enabled AppShell route presents first-run onboarding, even though the native layer explicitly could not determine whether this is an existing installation.

**User effect:** An existing user whose cohort marker or startup lookup is temporarily unavailable can be forced through the new-user onboarding flow on launch, interrupting access to their established workspace and making the product appear reset.

**Recommended fix:** Do not equate indeterminate evidence with a fresh installation. Show a bounded retry/recovery state or graduate conservatively unless the native cohort is positively fresh; preserve unsupported markers for forward compatibility without routing their owners into onboarding.

**Test:** Replace the test that expects unknown to remain incomplete with cases for timeout, malformed/future marker, and metadata failure that prove established users are never shown first-run onboarding without affirmative fresh-install evidence, while a positively fresh cohort still is.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +22,198 @@ const reveal = {
...
+            className="mt-9 w-[calc(100%-80px)] max-[760px]:mx-auto"
+            onClick={() => {
+              // Fire-and-forget: the landing page advances regardless of
+              // whether the native write lands. A failed write fails closed —
+              // the Rust-owned setting keeps its OFF default, so nothing is
+              // sent — and the Settings toggle remains the repair path. No
+              // explicit pipeline start either: telemetry initialized at boot,
+              // and the consent store update brings the pipeline up.
+              if (!consentHidden) {
+                updateTelemetryEnabled(shareUsageData).catch((error) => {
```
morgmart: 🤖 **P2 · Consent choice advances before save** (non-blocking)

The welcome action starts updateTelemetryEnabled without awaiting it, immediately advances onboarding, and handles rejection only with a console warning. The checkbox is presented as a user choice, but the flow gives no visible confirmation or repair cue when persistence fails; the Settings control may later disagree with what the user selected on welcome.

**User effect:** A user can explicitly choose whether to share usage data, continue believing that choice was saved, and later discover the opposite state—or silently remain opted out after choosing to share—with no explanation.

**Recommended fix:** Keep the action in a saving state until the native write confirms, advance only after success, and show a localized inline or toast error with retry on failure. If product intentionally permits skipping the write, clearly tell the user that the choice was not saved before advancing.

**Test:** Change the persistence-failure test to assert that onboarding does not advance, the action exits its saving state, and a localized retryable error is visible; also cover successful opt-in and opt-out writes.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for the exact three-dot comparison 64979a007ac78c0bf06c8c2664e819da585638db...78274b60bf23ecae0567714b725d55734910aaad. Two new blocking onboarding findings and one new non-blocking consent finding are publishable. Two still-present concerns are suppressed by authoritative thread memory: one unresolved automated thread and one resolved thread with a substantive human deferral. The prior Windows feature-gate mismatch, unavailable-consent UI, Windows settings replacement, mid-flight revocation, and alternate-port issues remain fixed. Supplied GitHub evidence was inspected and all listed checks report success: DCO, frontend checks/tests, frontend build smoke, clean-room install, transcript virtualization, and Tauri Rust checks on Linux, Windows, and macOS. CI remains separate from the code verdict and required checks independently govern merge readiness. Final self-check covered every changed app-launch, installation-cohort classification and migration, onboarding and consent flow, native export, chat send/steer/deferred-send, agent, project, Home pin, search, feedback, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; release/build integration; design-system and architectural-law compliance; duplicate overlap; current changed-line anchors; and concrete evidence plus user effect for blocking severity. The lower-confidence concurrent consent-writer concern was removed rather than overstated, and no finding was invented merely to produce output.

Deterministic publication result: 2 blocking and 1 non-blocking finding(s) publishable; 2 duplicate(s) suppressed.

src-tauri/src/commands/telemetry.rs
matt2e: 🤖 I don't think this one holds — `std::fs::rename` *does* replace an existing destination file on Windows. That's its documented contract, and it's what the implementation actually does.

**The docs.** [`std::fs::rename`](https://doc.rust-lang.org/std/fs/fn.rename.html) opens with: "Renames a file or directory to a new name, **replacing the original file if `to` already exists**." The only Windows-specific caveat is that `to` must not be an existing *directory* on filesystems/versions without `FileRenameInfoEx` support — not applicable to a JSON file.

**The implementation**, on this repo's toolchain (rustc 1.94.1, [`library/std/src/sys/fs/windows.rs`](https://github.com/rust-lang/rust/blob/1.94.1/library/std/src/sys/fs/windows.rs)):

```rust
if unsafe { c::MoveFileExW(old.as_ptr(), new.as_ptr(), c::MOVEFILE_REPLACE_EXISTING) } == 0
```

with a fallback on `ERROR_ACCESS_DENIED` (readonly destination) to `SetFileInformationByHandle` + `FileRenameInfoEx` passing `FILE_RENAME_FLAG_REPLACE_IF_EXISTS | FILE_RENAME_FLAG_POSIX_SEMANTICS`. Both paths replace. And per [Microsoft's `MoveFileExW` docs](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw), `MOVEFILE_REPLACE_EXISTING` means: "If a file named *lpNewFileName* exists, the function replaces its contents with the contents of the *lpExistingFileName* file."

I suspect the confusion is with **C's `rename()`** on the MSVC CRT (and `os.rename` in Python, and bare Win32 `MoveFileW`), which really do fail when the destination exists. Rust's std deliberately diverges. That's why the doc comment right above this function calls it out explicitly:

> The rename replaces an existing destination on Windows too — `std::fs::rename`'s documented contract, via `MoveFileExW`/`FileRenameInfoEx` with replace-if-exists, **unlike C's `rename`** — so re-toggling consent needs no remove-first step.

**On the requested test:** it already exists. `set_consent_persists_and_survives_restart` (telemetry.rs:1318-1339) sets consent true, restarts, then sets consent false — the second persist renames over the already-existing `telemetry-settings.json` — and asserts a fresh `TelemetryAuthState` reads false. There's also a false → true toggle at :1584-1585. Both are plain `#[test]`s with no `cfg` gating, so they run on Windows CI too.

**One real Windows caveat**, though not the described bug: if another process holds the file open without `FILE_SHARE_DELETE` (AV, backup tooling), the rename can fail transiently with `ERROR_SHARING_VIOLATION`. That's environmental and retryable rather than a deterministic second-write failure, and the current code handles it correctly — the error surfaces and the cached `AtomicBool` stays at the prior value, so memory and disk don't diverge.

Going to resolve this unless I've missed something.

src/features/onboarding/ui/WelcomeStep.tsx
matt2e: I think its fine for the rare case for this to fail it will use the default which is off

src/main.tsx
morgmart, line 197: 🤖 **P1 · Indeterminate installs enter onboarding** (blocking)

The new detection logic improves affirmative layout handling, but a cohort lookup timeout, IPC failure, unreadable metadata, corrupt marker, or unsupported marker still becomes the unknown cohort, and initializeOnboardingGraduation graduates only the explicit established cohort. With no valid local record, unknown retains the initial welcome state. A second path has the same contradiction: for an established cohort, a future-version onboarding record is preserved without setting the in-memory snapshot to completed, then hydration interprets that record as INITIAL_ONBOARDING_STATE. In both cases the system has not positively identified a fresh install but routes it into first-run onboarding.

**User effect:** An existing user can be forced through onboarding after a transient cohort failure, damaged/future marker, or downgrade from a newer app version, interrupting access to their established workspace and making Berd appear reset.

**Recommended fix:** Make first-run onboarding require affirmative fresh-install evidence. Treat unknown as a bounded recovery/retry or conservative established outcome, and when an established cohort has a future-version onboarding record, preserve its bytes while setting the current in-memory lifecycle to completed rather than parsing it as a fresh state.

**Test:** Replace the unknown-is-incomplete expectation with timeout, IPC error, malformed/future marker, and metadata-failure cases proving they cannot show first-run onboarding without affirmative freshness. Add a downgrade test that preserves a future-version record byte-for-byte while getOnboardingSnapshot reports completed for an established cohort; retain a positive fresh-cohort test.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for the exact three-dot comparison 64979a007ac78c0bf06c8c2664e819da585638db...0dad9bd081d84614117ac26e78e32d2425da6e55. One consolidated blocking onboarding finding is publishable: the previously resolved unknown-cohort issue still exists without a substantive human reply and can therefore be reopened, and the same fail-unsafe routing also occurs for a preserved future-version onboarding record. Three still-present concerns are suppressed by authoritative thread memory: one unresolved automated thread and two resolved threads with substantive human replies. The prior welcome-before-runtime issue is fixed at this SHA, as are the Windows feature-gate mismatch, unavailable-consent UI, mid-flight revocation, and alternate-port issues. Supplied GitHub evidence was inspected and all listed checks report success: DCO, frontend checks/tests, frontend build smoke, clean-room install, transcript virtualization, and Tauri Rust checks on Linux, Windows, and macOS. CI remains separate from the code verdict and required checks independently govern merge readiness. Final self-check covered every changed app-launch, installation-cohort classification and migration, onboarding/runtime/consent flow, native export, chat send/steer/deferred-send, agent, project, Home pin, search, feedback, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; release/build integration; design-system and architectural-law compliance; duplicate overlap; current changed-line anchors; and concrete evidence plus user effect for blocking severity. Overlapping established-user onboarding failures were consolidated into one finding, and no unsupported finding was added merely to produce output.

Deterministic publication result: 1 blocking and 0 non-blocking finding(s) publishable; 3 duplicate(s) suppressed.

src/main.tsx
matt2e: i think this is fine for launch

morgmart (commented):
🤖 Automated code review

NO NEW PUBLICATION. Fresh static review completed for the exact three-dot comparison 64979a007ac78c0bf06c8c2664e819da585638db...17c34a7bfa8e41a1126a686a7a381650fb3cb79e with zero new publishable findings. Four still-present concerns are suppressed by authoritative thread memory: two unresolved automated threads and two resolved threads with substantive human replies. Because unresolved prior automated findings remain, approval is not recommended and no new publication is required. The latest changes preserve the onboarding consent answer across back navigation/replay, prevent a delayed startup read from overwriting a completed consent write, and harden Home pin telemetry without exposing a new supported finding. The welcome-before-runtime fix remains present; the Windows feature-gate mismatch, unavailable-consent UI, mid-flight revocation, and alternate-port issues remain fixed. Supplied GitHub evidence was inspected: DCO, clean-room install, and Linux Rust checks passed; frontend checks/tests and Rust checks on macOS and Windows were still in progress. Pending checks do not make validation incomplete, and required checks independently govern merge readiness. Final self-check covered every changed app-launch, installation-cohort classification and migration, onboarding/runtime/consent flow, native export, chat send/steer/deferred-send, agent, project, Home pin, search, feedback, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; release/build integration; design-system and architectural-law compliance; duplicate overlap; current changed-line anchors; and concrete evidence plus user effect for blocking severity. No unsupported finding was added merely to produce output.

Deterministic publication result: 0 blocking and 0 non-blocking finding(s) publishable; 4 duplicate(s) suppressed.

kalvinnchau (approved):

PR #21, opened 2026-08-13
fix(release): forward unix provenance arguments

kalvinnchau:
Fixes Linux and macOS release provenance generation under Ubuntu `/bin/sh`.

Technical changes:
- forwards the complete positional argument list with POSIX-compatible `"$@"` before invoking the nested Bash process
- removes the `${@:5}` expansion that fails under `dash`
- updates the release recipe assertion to match the corrected command

Validation:
- focused Ubuntu `act` job generated and validated Linux and macOS provenance receipts without running application builds
- `just release-scripts-test` (78 tests passed)

### DIFF
--- a/justfile
+++ b/justfile
@@ -277,7 +277,7 @@ release-reconcile-assets repository tag version output_file:
 [unix]
 [positional-arguments]
 release-write-provenance source_sha version platform output_dir *ASSETS:
-    bash -euo pipefail -c 'scripts/release/write-provenance.sh "$@"' _ "$1" "$2" "$3" "$4" "${@:5}"
+    bash -euo pipefail -c 'scripts/release/write-provenance.sh "$@"' _ "$@"
 
 # On Windows, run the shared Bash script as one generated recipe so argv remains
 # positional and is not interpolated into shell source.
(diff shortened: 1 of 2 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of exact three-dot range 6e6e19ee58e1cbd981e88108f6e27b733d718140...8f5f41db69dc07cf0c91472e9e83ec9237e6fa8a found no publishable findings. The Unix release-provenance recipe now forwards its complete positional argv through the POSIX outer shell into Bash without the dash-incompatible ${@:5} expansion, and the existing behavioral test exercises the changed recipe with spaces, a quote, and a literal dollar sign before verifying the complete receipt, so the regression coverage is discriminating rather than assertion-only. Final self-check covered the changed macOS/Linux provenance flow and unchanged Windows counterpart; accessibility, i18n, navigation, consent, and design-system rules (not applicable because there is no UI/product-surface change); synchronous failure handling plus async, never-completes, lifecycle, and race behavior; test honesty; project laws; duplicate overlap; and the evidence/user-effect bar for blocking severity. No unsupported or overlapping issue survived. Supplied GitHub evidence was inspected: all 10 captured check runs completed successfully at the exact head SHA, while the separate legacy combined status is pending with zero status contexts; required checks still govern merge readiness. Supplied review threads are empty, so there is no unresolved prior automated finding or suppression.

PR #27, opened 2026-08-13
fix(release): generate cumulative stable notes

kalvinnchau:
## summary

- fetch origin/main and published release tags into an isolated, pruned snapshot
- keep prerelease notes incremental while stable notes span the previous stable release
- regenerate reviewed notes when either origin/main or the selected release baseline changes
- preserve public tag names in full changelog links and the existing release command interface

## validation

- just ci
- pnpm test:release-scripts

### DIFF
--- a/scripts/release/release.mjs
+++ b/scripts/release/release.mjs
@@ -423,6 +423,35 @@ function fetchReleaseState(root) {
   );
 }
 
+function remoteReleaseTags(root) {
+  const output = run(
+    "git",
+    ["ls-remote", "--tags", "--refs", "origin", "refs/tags/v*"],
+    { cwd: root },
+  );
+  return new Set(
+    output
+      .split("\n")
+      .filter(Boolean)
+      .map((line) => line.split("\t", 2)[1]?.replace("refs/tags/", ""))
+      .filter(Boolean),
+  );
+}
+
+function describeRemoteReleaseTag(root, stableOnly) {
+  const remoteTags = remoteReleaseTags(root);
+  const args = ["describe", "--tags", "--abbrev=0", "--match", "v*"];
+  if (stableOnly) args.push("--exclude", "v*-*");
+
+  while (true) {
+    const result = commandResult("git", args, { cwd: root });
+    if (result.status !== 0) return null;
+    const tag = result.stdout.trim();
+    if (remoteTags.has(tag)) return tag;
+    args.push("--exclude", tag);
+  }
+}
+
 function remoteTagTarget(root, tag) {
   const result = commandResult(
     "git",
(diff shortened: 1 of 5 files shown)


scripts/release/release.mjs
```diff
@@ -462,9 +463,42 @@ function validateLocalTag(root, tag, expectedTarget) {
   }
 }
 
-async function generateReviewedNotes(root) {
-  process.stderr.write("generating release notes...\n");
-  const notes = run("just", ["release-notes"], { cwd: root });
+export function releaseNotesFrom(root, version, changelog) {
+  const args = ["describe", "--tags", "--abbrev=0", "--match", "v*"];
```
morgmart: 🤖 **P1 · Local tags can truncate notes** (blocking)

The baseline is selected with `git describe --tags`, which searches every reachable local `v*` tag. `fetchReleaseState` adds the remote tags but neither distinguishes them from nor removes local-only tags, so an unpublished developer tag can become the nearest prerelease or stable baseline. The new test reinforces this behavior by selecting `v0.6.0-rc.1` without ever pushing that tag to the fixture's origin.

**User effect:** A release operator can approve notes that omit commits shipped since the real published release, or note generation can stop with an empty range, based solely on an unrelated local tag that users have never received.

**Recommended fix:** Resolve baselines only from tags confirmed on origin, preferably by fetching release tags into a dedicated remote namespace and selecting reachable refs there, or by verifying the selected tag and object against the remote before using it. Do not make release-note correctness depend on the developer's local tag namespace.

**Test:** Push the legitimate baseline tags to the fixture origin, then add a closer reachable local-only `v*` tag and assert both prerelease and stable baseline selection ignore it and return the published origin tag.

scripts/release/release.mjs
```diff
@@ -549,13 +583,13 @@ async function prepare(version, notesPath) {
   const branch = `release/v${version}`;
   const subject = `chore: release v${version}`;
   assertClean(root);
+  run("gh", ["auth", "status", "--hostname", "github.com"], { cwd: root });
+  fetchReleaseState(root);
```
morgmart, line 665: 🤖 **P1 · Approval uses stale main snapshot** (blocking)

The origin/main refresh now happens before generated notes are created and the operator is asked to review them. That review can take an unbounded amount of time. If main advances meanwhile, the later up-to-date check still compares local HEAD with the stale origin/main ref and passes, allowing preparation to continue from the old cutoff. The release PR can then merge after the newer main commits, making the eventual release tag include changes that were absent from the approved notes.

**User effect:** Changes that actually ship in the release can be missing from its changelog even though the operator reviewed and explicitly approved the generated notes.

**Recommended fix:** Refresh origin/main again after note approval and before creating or validating the release branch. If the reviewed cutoff moved, regenerate the notes and require fresh approval (or abort with a clear restart message) rather than proceeding with stale notes.

**Test:** Have the fixture advance origin/main during the generated-note/approval interval, then assert preparation does not create or push a release branch or PR from the stale snapshot and instead aborts or regenerates notes from the new cutoff.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the stable-note flow can select an unpushed local stable tag, and generated notes can be approved against a stale main snapshot, so a published release may omit shipped changes from its changelog. Static review covered all three changed files and the full exact three-dot comparison. The final self-check covered every changed release flow, consent, accessibility and i18n applicability, lifecycle/race/error behavior, test honesty, project rules, duplicates, and severity evidence; no other supported findings remained. Supplied GitHub evidence was structurally valid and inspected: seven checks were successful and Frontend checks and unit tests was still in progress at capture time; required checks still govern merge readiness.

scripts/release/release.mjs
kalvinnchau: 🤖 fixed in 1e731f1. baseline selection now filters `git describe` candidates against tags currently advertised by `origin`, ignoring local-only and remotely deleted tags. regression coverage pushes the legitimate baselines and verifies both cases.

scripts/release/release.mjs
kalvinnchau: 🤖 fixed in c0483a6. preparation snapshots `origin/main` around approval; if it advances, it fast-forwards local `main`, regenerates notes, and asks for fresh approval before creating the release branch. the regression advances `origin/main` from a separate clone and verifies both generations plus successful PR creation.

morgmart (approved):

PR #38, opened 2026-08-14
Simplify Compose control-plane authentication

nathan-thillairajah:
## Summary

`bb apps` previously exchanged the normal opaque bbidentity session for a Compose-specific purpose token, then cached, refreshed, and retried requests with that token. The new contract moves that exchange to public ingress: while the request is paused, ingress sends the bbidentity session to the authorizer and replaces it with a short-lived Compose JWT before forwarding.

The CLI now sends `Authorization: BBIdentity <credential>` only to the exact staging and production Compose control-plane hosts over HTTPS. Redirects stay disabled, cookies and legacy identity headers are omitted, and reflected credentials are redacted from errors. Apps commands require an already-valid stored session in every mode; otherwise they fail promptly with `auth_required` and direct the caller to explicit `bb auth login`. Only the explicit auth command owns the browser login flow. The old exchange, cache, refresh, and retry machinery is removed.

The staging infrastructure still needs to be activated before this can be smoke-tested end to end. Merging this PR also does not update an installed `bb` binary: Berd ships the CLI through its normal pinned release process.

### Related issue

N/A. This replaces auth work from the archived pre-OSS repository.

### Testing

- `cd bb-cli && source ./bin/activate-hermit && just check`
- `cd bb-cli && source ./bin/activate-hermit && just build-bb-release`


bb-cli/src/bb/apps.rs
```diff
@@ -363,263 +349,60 @@ fn validate_artifact_path(path: &Path) -> Result<()> {
...
+    authorization: HeaderValue,
+    secret: String,
 }
 
-impl KgoosePurposeTokenProvider {
+impl ComposeSessionCredential {
     fn from_config(config: &SkillsConfig) -> Result<Self> {
         let storage = default_session_storage(config)?;
         let session_storage_key = session_storage_key_from_config(config);
+        Self::after_login(storage.as_ref(), &session_storage_key, || {
```
morgmart: 🤖 **P1 · Embedded login can wait forever** (blocking)

Every Apps command now calls ensure_browser_login before sending its request. When the stored session is missing or invalid, that helper starts a loopback callback server and blocks on recv() with no deadline. If the browser opens but the user abandons login, the command never returns; in JSON or other noninteractive use it can wait without even printing the login URL to stdout. The new after_login unit test uses an immediately completing closure and therefore cannot exercise this path.

**User effect:** An agent, CI job, JSON caller, or terminal user running bb apps contract/create/deploy can hang indefinitely instead of receiving an actionable auth-required result.

**Recommended fix:** Give embedded authentication an explicit interaction policy: fail fast with auth_required for JSON/noninteractive callers, and bound any permitted interactive callback wait with timeout/cancellation. Keep the unbounded user-owned flow, if desired, behind the explicit bb auth login command.

**Test:** Add a discriminating command-level test with no valid session and no callback completion; assert noninteractive/JSON Apps execution exits promptly with auth_required. Add a bounded-timeout test for the interactive embedded policy if that path remains.

bb-cli/src/bb/apps.rs
```diff
@@ -363,263 +349,60 @@ fn validate_artifact_path(path: &Path) -> Result<()> {
...
-        let file = OpenOptions::new()
-            .read(true)
-            .write(true)
-            .create(true)
-            .truncate(false)
-            .open(&self.refresh_lock_path)
-            .with_context(|| format!("open {}", self.refresh_lock_path.display()))?;
-        #[cfg(unix)]
+    fn new(secret: String) -> Result<Self> {
+        if !(32..=512).contains(&secret.len())
```
morgmart: 🤖 **P1 · Apps redefines credential validity** (blocking)

After builderbot-auth has loaded the session and /v1/auth/me has accepted it, Apps imposes a second feature-local format contract: 32-512 ASCII alphanumeric, underscore, or hyphen characters. The shared auth boundary only requires a nonempty HTTP-header-safe credential, so a backend-issued credential containing another valid header character, or a future credential format change, can authenticate successfully and then be rejected only by bb apps. The added test pins this duplicated rule rather than checking compatibility with the authoritative auth contract.

**User effect:** Users with a valid BuilderBot session can be blocked from every Apps command and told to log in again, even though logging in can return the same valid credential.

**Recommended fix:** Let builderbot-auth own session validation and have Apps perform only the encoding needed to construct the BBIdentity Authorization header. If these exact length/character constraints are a protocol requirement, move and document them in the shared auth contract so login, storage, verification, and Apps cannot drift.

**Test:** Verify that a credential accepted by the shared auth boundary, including representative header-safe punctuation, can be used by Apps; or, if the protocol truly forbids it, add the same validation at shared auth ingress and test that an invalid exchange response is rejected there.

bb-cli/src/bb/apps.rs
```diff
@@ -927,179 +678,145 @@ fn network_failure(method: &str, path: &str, error: reqwest::Error) -> anyhow::E
...
-    let next_action = parsed
-        .as_ref()
-        .and_then(|value| {
-            value
-                .get("next_action")
-                .or_else(|| value.pointer("/error/next_action"))
-        })
-        .and_then(Value::as_str);
+    let code = credential.redact(&terminal_safe_text(code));
+    let next_action = if status == StatusCode::UNAUTHORIZED {
```
morgmart, line 822: 🤖 **P2 · 401 guidance reuses rejected session** (blocking)

On a Compose 401 the new error tells the user to run bb auth login. That command first verifies the stored credential with /v1/auth/me and returns it unchanged whenever kgoose still considers it valid. This is the same preflight used before the rejected Apps request, so the documented recovery can immediately reuse the credential Compose just rejected and reproduce the failure indefinitely.

**User effect:** A user whose otherwise-valid session is rejected by Compose can follow the displayed recovery instructions repeatedly without regaining access or learning how to force a fresh credential.

**Recommended fix:** Provide a forced-refresh path for this rejection (for example, invalidate the stored session after an Apps 401 and then authenticate), or change the guidance to a sequence that actually replaces the credential, such as logout followed by login. Keep the action suitable for noninteractive callers.

**Test:** Simulate /v1/auth/me accepting the stored session while Compose rejects that same session with 401, then exercise the recommended recovery and assert that the next Apps request uses a newly issued credential rather than the rejected one.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review of the exact three-dot range found 3 blocking findings and 0 non-blocking findings. The shared authentication path can wait forever when embedded in an Apps command, Apps independently rejects credentials already accepted by the authoritative auth service, and the prescribed recovery from a Compose 401 reuses the same rejected credential. Final self-check covered contract/create/deploy flows; CLI accessibility and localization applicability; allowlist, redirect, credential-consent, and navigation guards; async, failure, never-completes, lifecycle, and race behavior; test honesty; project rules; duplicate overlap; and blocking evidence/user effect. No frontend UI changed, so visual accessibility, responsive design, and design-system token checks are not applicable. Supplied GitHub evidence was inspected: it is structurally available, reports no check runs and a pending combined status; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 83874233. Apps now treats JSON and non-TTY callers as noninteractive: it verifies any stored session with /v1/auth/me and returns structured auth_required without opening a browser when no valid session exists. Interactive embedded login has a five-minute callback deadline, while the explicit bb auth login flow remains user-owned and unbounded. The command-level prompt-exit test and callback-timeout unit test cover both policies.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 83874233. Apps no longer defines a separate credential length or character alphabet. It accepts the credential already validated by shared BuilderBot auth, including representative header-safe punctuation, and relies on the HTTP HeaderValue parser when constructing BBIdentity authorization so CR/LF injection and other unsafe header values are still rejected. The updated unit test covers both cases.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 83874233. A Compose 401 now tells the user to run bb auth logout and then bb auth login, so recovery deletes the rejected stored session before starting a fresh exchange instead of allowing login to reuse it after /v1/auth/me succeeds. The 401 unit test locks that sequence and redaction, while the auth logout end-to-end test verifies the stored credential is removed.

bb-cli/src/bb/apps.rs
```diff
@@ -363,263 +349,60 @@ fn validate_artifact_path(path: &Path) -> Result<()> {
...
+    secret: String,
 }
 
-impl KgoosePurposeTokenProvider {
+impl ComposeSessionCredential {
     fn from_config(config: &SkillsConfig) -> Result<Self> {
         let storage = default_session_storage(config)?;
         let session_storage_key = session_storage_key_from_config(config);
+        if config.json || !stdin_is_tty() {
+            if verify_stored_session(config, storage.as_ref())?.is_none() {
```
morgmart: 🤖 **P2 · Verification races credential selection** (blocking)

The noninteractive path calls verify_stored_session, which loads and verifies the credential currently in storage, but discards that credential. It then calls after_login, which performs a second storage read and constructs the Compose Authorization header from whatever value is present then. A concurrent login, logout, second CLI process, or file replacement between those reads can therefore cause Apps to send a credential that was never verified. The interactive path has the same re-read after ensure_browser_login returns.

**User effect:** An Apps command can fail unpredictably after a successful auth check or send a newly substituted session to Compose without first validating that session against the authoritative auth service.

**Recommended fix:** Have the shared authentication operation return the exact verified or newly issued StoredSessionCredential and construct ComposeSessionCredential directly from that value. Avoid using a verification result as a predicate followed by another storage lookup.

**Test:** Use a storage double that returns credential A for verification and credential B on the next read. Assert the command either sends verified credential A or rejects/reverifies B, and never sends B merely because storage changed between calls. Cover both noninteractive verification and interactive login completion.

bb-cli/tests/bb_e2e.rs
```diff
@@ -3588,424 +3641,18 @@ fn bb_apps_help_distinguishes_external_and_internal_paths() {
 }
 
 #[test]
-fn bb_apps_contract_exchanges_session_and_calls_control_plane() {
```
morgmart, line 3591: 🤖 **P2 · Successful CLI coverage was removed** (blocking)

The PR deletes the successful process-level tests for contract, create with initialization, create without initialization, and multipart deploy. Those tests exercised real command parsing, config and stored-session wiring, request bodies, response shaping, and command output. The replacements test rejected origins and missing auth at process level, while successful requests use ControlPlaneClient::new_for_test, which bypasses production URL validation and does not exercise run_create or the full CLI dispatch path. The suite can now pass while the production command cannot complete a successful request or while create orchestration/output regresses.

**User effect:** A release can ship with contract, app creation, initialization decisions, or deploy broken end to end even though the changed test suite remains green.

**Recommended fix:** Restore successful command-level coverage for contract, both create branches, and deploy using a test seam that preserves production destination validation while routing approved hosts to a local mock. Keep assertions for BBIdentity authorization, omitted cookie/legacy identity headers, request payloads and multipart fields, and final output.

**Test:** The restored tests should invoke the built bb command with a stored session, have /v1/auth/me validate it, complete each approved control-plane flow, and assert the exact requests and outputs. They must fail if production allowlist validation, auth-to-dispatch wiring, create branching, or multipart construction breaks.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found 2 blocking findings and 0 non-blocking findings. The noninteractive authentication path verifies one stored credential and then re-reads storage before sending, allowing a concurrent change to substitute an unverified credential. The PR also removes all successful process-level contract/create/deploy coverage while changing the authentication and dispatch path for each command; replacement unit tests bypass production destination validation and do not preserve create orchestration or full CLI wiring. The three previously automated issues were re-evaluated against current code and are fixed: embedded waits are bounded with prompt suppression for JSON/non-stdin-TTY execution, feature-local credential format restrictions were removed, and 401 guidance now prescribes logout before login. Final self-check covered contract, create/initialize, create-without-initialize, and deploy flows; CLI accessibility and localization applicability; destination, redirect, consent, and interaction guards; async, timeout, error, never-completes, lifecycle, and race behavior; test honesty; project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No frontend UI changed, so responsive layout, visual accessibility, screenshots, and design-system tokens are not applicable. Supplied GitHub evidence and all three supplied resolved threads were inspected. Evidence is structurally available, reports no completed check runs and a pending combined status; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in d257c3ec. Shared auth now returns the exact credential it verified or issued, and Apps constructs BBIdentity directly from that value without rereading storage. Focused storage-double tests cover both branches: verification returns credential A even when the next stored value is B, and embedded login completion returns the issued credential without a follow-up read.

bb-cli/tests/bb_e2e.rs
nathan-thillairajah: 🤖 Addressed in d257c3ec. Successful built-binary coverage is restored for contract, create with initialization, create without initialization, and multipart deploy. Each flow uses an approved production hostname, validates the stored session through /v1/auth/me, and then routes only the debug test transport to loopback after allowlist validation. The tests assert the same BBIdentity credential, omitted cookie and legacy identity headers, request payloads and multipart fields, create branching, and final JSON output; release builds contain no transport override.

bb-cli/src/bb/apps.rs
```diff
@@ -649,19 +420,21 @@ impl ControlPlaneClient {
...
-                "Apps Platform control-plane URL must use https unless it targets loopback local development"
-            );
-        }
         if !is_trusted_control_plane_url(&contract_url) {
             anyhow::bail!(
-                "Apps Platform control-plane URL must target an approved Builderlab ingress host or loopback local development"
+                "Apps Platform control-plane URL must use HTTPS and target an approved Builderlab ingress host"
             );
         }
+        let transport_base_url = control_plane_transport_base_url(base_url)?;
```
morgmart: 🤖 **P1 · Debug builds bypass recipient allowlist** (blocking)

ControlPlaneClient first validates the user-facing base URL as one of the two approved HTTPS ingress hosts, but then control_plane_transport_base_url can replace the actual request destination from BB_APPS_E2E_CONTROL_PLANE_URL whenever debug_assertions is enabled. That condition includes ordinary locally built debug binaries, not only the test harness. Any local process can bind a loopback port, so an environment value can cause the long-lived BBIdentity session to be sent to a recipient that was never approved even though the preceding allowlist check succeeded. The restored process tests rely on this branch, so they also do not exercise the release destination wiring they appear to validate.

**User effect:** A user running a development build can unknowingly disclose their reusable BuilderBot session to an untrusted local process, and successful tests can pass while the production recipient path is broken.

**Recommended fix:** Keep the validated URL authoritative through request dispatch. Move interception below URL ownership into a test-only connector/client seam, or compile the override only under a dedicated test-only feature unavailable in ordinary debug binaries. Ensure successful command tests exercise the same destination-selection logic as release builds.

**Test:** Add a test against an ordinary debug binary showing that setting the environment variable cannot change the credential recipient. Keep successful CLI coverage through an injected test transport that preserves the validated production URL and fails if release destination construction diverges.

bb-cli/src/bb/auth_login.rs
```diff
@@ -111,60 +196,100 @@ pub fn run_browser_login(
...
     }
 
-    let code = rx
-        .recv()
-        .context("loopback auth server stopped before login completed")??;
+    let code = wait_for_login_callback(rx, callback_timeout)?;
     let verified =
         exchange_login_code_and_verify(&client, config.playpen.as_deref(), &service_url, &code)?;
-    let stored = verified.credential;
+    let stored = store_login_credential(storage, &storage_key, verified.credential)?;
```
morgmart: 🤖 **P2 · Failed login persists unusable session** (blocking)

After the browser exchange succeeds, the revised flow calls store_login_credential before active_workspace_name validates the verified response. If /v1/auth/me contains no active workspace, the command returns an error but leaves the new credential installed. On the next login attempt, that stored credential can verify again and hit the same workspace error in the early stored-session branch, so the browser flow is never offered. The base implementation validated the workspace before storage; this regression was introduced while changing the function to return the exact issued credential.

**User effect:** A user whose login response lacks an active workspace can become stuck in a repeatable failed-login state and must discover that they need to log out or manually clear local credentials before trying again.

**Recommended fix:** Validate all required response state, including the active workspace, before persisting the issued credential. Then store and return that same credential so the race fix remains intact without committing an unusable session.

**Test:** Simulate an otherwise verified browser exchange whose /v1/auth/me response has no active workspace. Assert login fails without writing the credential, then assert a subsequent attempt can start a fresh browser login rather than reusing the failed session.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found 2 blocking findings and 0 non-blocking findings. Ordinary debug builds now contain an environment-controlled transport override that can send the long-lived BBIdentity credential to an unapproved loopback process after validating a different approved URL. The revised login flow also persists a newly issued credential before validating that the response has an active workspace, so this failure leaves a session installed that causes subsequent login attempts to fail before reopening the browser. All five supplied resolved automated threads were re-evaluated against current code; their underlying issues are fixed and no current finding duplicates them. Final self-check covered contract, create/initialize, create-without-initialize, deploy, login, status, logout, and legacy-cache flows; CLI/browser accessibility and localization applicability; destination, redirect, navigation, consent, and interaction guards; async, timeout, error, never-completes, lifecycle, and race behavior; test honesty; project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No frontend product UI changed, so responsive layout, application design-system tokens, and PR screenshots are not applicable. Supplied GitHub evidence and all five supplied threads were inspected. Evidence is structurally available, reports no completed check runs and a pending combined status; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 46ad0138. The environment-controlled destination override is removed entirely: ordinary debug and release binaries now execute requests directly through the client against the validated approved HTTPS base URL. Successful contract/create/deploy coverage uses a `cfg(test)`-only executor after request construction, and that executor asserts the request still targets an approved production URL before forwarding it to a local test server.

bb-cli/src/bb/auth_login.rs
nathan-thillairajah: 🤖 Addressed in 46ad0138. A newly exchanged session now has its active workspace validated before the credential is persisted. If that validation fails, login returns the error with storage untouched, so the next attempt can start a fresh browser flow; the exact issued credential is still stored and returned on success, preserving the earlier race fix. A focused regression test covers the no-workspace/no-write case.

bb-cli/src/bb/apps.rs
```diff
@@ -363,274 +369,63 @@ fn validate_artifact_path(path: &Path) -> Result<()> {
...
-        if status == StatusCode::TOO_MANY_REQUESTS {
-            self.style
-                .verbose(&format!("POST {COMPOSE_TOKEN_EXCHANGE_PATH} -> {status}"));
-            return Ok(PurposeTokenExchangeOutcome::RateLimited);
-        }
-        if !status.is_success() {
-            self.style
-                .verbose(&format!("POST {COMPOSE_TOKEN_EXCHANGE_PATH} -> {status}"));
-            return Err(exchange_http_failure(status));
+        if config.json || !stdin_is_tty() {
```
morgmart: 🤖 **P2 · Pipelines trigger embedded browser login** (blocking)

Apps decides whether implicit browser authentication is allowed using config.json or stdin_is_tty only. A normal pipeline such as bb apps contract ... | jq keeps terminal stdin while stdout is piped, and Apps commands emit JSON even without --json. With no valid session, that machine-consumed command therefore enters the interactive branch, opens a browser, and can wait for the five-minute callback timeout instead of failing promptly. The existing prompt-suppression test passes --json, so it does not exercise this common pipeline shape.

**User effect:** Scripts and terminal pipelines can unexpectedly launch a browser and stall for up to five minutes, making automation appear hung and requiring user intervention where a deterministic auth_required error is expected.

**Recommended fix:** Treat Apps as noninteractive when its output is not attached to a terminal as well as when stdin is non-TTY or --json is set, or remove implicit login from these JSON-producing commands entirely and reserve it for explicit bb auth login.

**Test:** Spawn bb apps contract without --json using terminal-like stdin but piped stdout and no stored session. Assert it exits promptly with auth_required, produces no browser-login guidance, and never starts a callback wait.

bb-cli/tests/bb_e2e.rs
morgmart, line 3591: 🤖 **P2 · Successful CLI coverage was removed** (blocking)

This head again deletes the successful built-binary tests for contract, create with initialization, create without initialization, and multipart deploy. The replacements call create_response, deploy_response, and ControlPlaneClient directly inside the module with a cfg(test) transport. They cover parsing and request construction, but they do not exercise production command dispatch, config and stored-session verification, ComposeSessionCredential::from_config, output behavior, or the final self.client.execute path. A broken successful command can therefore coexist with a green suite; the remaining built-binary Apps tests cover only rejected destinations and missing authentication.

**User effect:** Contract retrieval, app creation, initialization branching, or deploy can break in the shipped CLI wiring while all changed tests continue to pass.

**Recommended fix:** Restore successful built-binary coverage for contract, both create branches, and deploy without adding a runtime credential-routing override. Use a test-only network seam below destination ownership, or another harness that exercises the production dispatch/auth/output path while preserving the approved request URL through the final executor.

**Test:** Each restored test should invoke the bb binary with an approved base URL and stored session, verify that /v1/auth/me checks the same credential, complete the control-plane operation, and assert BBIdentity authorization, omitted cookie and legacy identity headers, request payload or multipart fields, branching, and final JSON output. It must fail if production dispatch, auth wiring, or the release executor path diverges.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found 2 blocking findings and 0 non-blocking findings. Apps still classifies ordinary stdout pipelines as interactive because it checks only stdin, so a JSON-producing command piped to another program can unexpectedly open a browser and wait five minutes. In addition, the successful built-binary contract/create/deploy tests were removed again; replacement module tests bypass production configuration, session verification, dispatch, and the final production HTTP executor. The seven supplied resolved automated threads were re-evaluated against current code. The credential recipient override, credential-selection race, local format restriction, ineffective 401 guidance, failed-login persistence ordering, and unbounded callback wait are fixed. The successful CLI coverage issue and the broader embedded-interaction issue remain present in their current forms; because their resolved threads contain no substantive human replies, they are eligible to be raised again under the supplied duplicate policy. Final self-check covered contract, create/initialize, create-without-initialize, deploy, login, status, logout, and legacy-cache flows; CLI/browser accessibility and localization applicability; destination, redirect, navigation, consent, and interaction guards; async, timeout, error, never-completes, lifecycle, and race behavior; test honesty; project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No frontend product UI changed, so responsive layout, application design-system tokens, and screenshots are not applicable. Supplied GitHub evidence and all seven supplied threads were inspected. Evidence is structurally available, reports no completed check runs and a pending combined status; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 8bc2ef96. Apps commands no longer perform implicit browser login in any mode: they only verify an existing stored session and otherwise return auth_required, leaving browser ownership to explicit bb auth login. A built-binary regression test runs without --json with piped output and asserts prompt exit with no browser guidance.

bb-cli/tests/bb_e2e.rs
nathan-thillairajah: 🤖 Addressed in 8bc2ef96. Restored built-binary success coverage for contract, create with and without initialize, and multipart deploy. A dedicated test feature resolves the approved HTTPS host to a loopback TLS server with a test CA; requests keep the approved URL and execute through the production client. The tests verify the exact /v1/auth/me credential, BBIdentity and omitted legacy headers/cookies, payloads, branching, multipart data, and JSON output; ordinary debug and release binaries contain no test routing.

bb-cli/src/bb/apps.rs
```diff
@@ -827,6 +605,44 @@ impl ControlPlaneClient {
...
+        if let Some(transport) = self.test_transport.as_ref() {
+            return transport.execute(request);
+        }
+        self.client.execute(request)
+    }
+}
+
+fn build_control_plane_http_client(timeout: Duration) -> Result<Client> {
+    let builder = Client::builder().redirect(Policy::none()).timeout(timeout);
+    #[cfg(feature = "apps-e2e-test")]
```
morgmart: 🤖 **P1 · Test feature ships credential bypass** (blocking)

apps-e2e-test is an ordinary public Cargo feature, not a cfg(test)-only boundary. Any bb release compiled with this feature includes BB_APPS_E2E_RESOLVE_ADDR handling, trusts the checked-in test CA, and resolves both approved production hostnames to a caller-selected loopback socket. The matching server private key is also committed, so a local process can present a valid certificate for those names. The URL allowlist still passes while the reusable BBIdentity session is delivered to that process. All-features and other feature-enabled release workflows can therefore produce a distributable binary containing this bypass.

**User effect:** A feature-enabled bb build can disclose the user's full BuilderBot session to an untrusted local process while appearing to connect securely to an approved Compose hostname.

**Recommended fix:** Make the shipped bb target incapable of compiling the DNS override, embedded CA, or known matching private-key trust path under every feature combination. Put the TLS/DNS substitution in a separate test harness or test-only binary boundary rather than a Cargo feature on the distributable target.

**Test:** Add a release/all-features artifact check proving that the shipped bb binary does not contain or honor BB_APPS_E2E_RESOLVE_ADDR and does not embed the fixture CA. Keep E2E coverage in a harness that cannot become the release executable.

bb-cli/src/bb/apps.rs
```diff
@@ -778,26 +559,20 @@ impl ControlPlaneClient {
...
-        let (mut status, mut body) =
-            self.request_response(method, path, &send, authorization.clone())?;
-        if status == StatusCode::UNAUTHORIZED {
-            if let Some(replacement) =
-                credential_provider.authorization_header_after_rejection(&authorization)?
-            {
-                (status, body) = self.request_response(method, path, &send, replacement)?;
-            }
-        }
+        let authorization = credential.authorization_header();
```
morgmart, line 620: 🤖 **P2 · Success responses can leak session** (blocking)

The control plane now receives the long-lived BuilderBot session in Authorization, but authorized_json_request only invokes credential-aware redaction for non-success statuses. On success it parses the body and returns the Value unchanged; contract, create, and deploy later print that value. A buggy or compromised approved ingress can therefore reflect the session in a 200 JSON field and cause the CLI to write it to stdout. Existing successful-response tests merely use fixtures that do not contain the credential, so they cannot detect this regression.

**User effect:** A reflected session can be exposed in terminal output, shell pipelines, CI logs, or captured command artifacts even though error responses are explicitly protected from the same leak.

**Recommended fix:** Apply credential redaction to every control-plane response before it reaches command output, including recursively sanitizing string values in successful JSON while preserving the rest of the response shape. Alternatively, reject any successful response containing the exact secret.

**Test:** Return a successful JSON response with the exact BBIdentity session nested in multiple string fields for contract, create, or deploy. Assert stdout and the returned Value never contain the secret and that ordinary non-secret response data remains intact.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found 2 blocking findings and 0 non-blocking findings. The new apps-e2e-test Cargo feature can be compiled into a distributable bb binary; in that configuration an environment variable, embedded test CA, and checked-in matching private key let a loopback process impersonate either approved host and receive the reusable BBIdentity session. Also, successful control-plane responses are returned and printed without credential redaction, so an approved ingress that reflects the Authorization secret in a 200 JSON response can leak it to stdout. All nine supplied resolved automated threads were re-evaluated against current code. Their underlying issues are fixed: Apps is now always noninteractive, exact credential selection avoids a second storage read, failed workspace validation precedes persistence, URL validation remains exact, and built-binary success coverage has been restored. Neither current finding duplicates those threads. Final self-check covered contract, create/initialize, create-without-initialize, deploy, login, status, logout, legacy-cache, and test-feature flows; CLI/browser accessibility and localization applicability; destination, TLS, redirect, navigation, consent, and interaction guards; async, timeout, error, never-completes, lifecycle, and race behavior; successful and failing response redaction; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No frontend product UI changed, so responsive layout, application design-system tokens, and screenshots are not applicable. Supplied GitHub evidence and all nine supplied threads were inspected. Evidence is structurally available, reports no completed check runs and a pending combined status; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 7bf10a3d. Removed the Cargo feature, DNS override, embedded test CA, matching private key, and TLS fixture dependency. Successful CLI flows now run in a self-spawned Rust test executable whose loopback transport exists only under cfg(test); the distributable bb target has no runtime or feature path to it. The all-features test gate scans the built bb artifact for both override names and the test CA marker, and a release/all-features build was also verified clean.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 7bf10a3d. Every successful control-plane JSON response is now recursively sanitized before it can reach contract/create/deploy output, including string values and object keys while preserving non-secret data and response shape. The process-level contract test reflects the exact session in nested successful fields and asserts stdout contains only [REDACTED] while ordinary contract data remains intact.

bb-cli/src/bb/apps.rs
```diff
@@ -363,272 +355,90 @@ fn validate_artifact_path(path: &Path) -> Result<()> {
...
+}
 
-    fn authorization_header_after_rejection(
-        &self,
-        rejected: &HeaderValue,
-    ) -> Result<Option<HeaderValue>> {
-        self.refresh_or_adopt(Some(rejected))
+#[cfg(test)]
+impl LoopbackTestTransport {
+    fn new(base_url: &str, timeout: Duration) -> Result<Self> {
```
morgmart, line 423: 🤖 **P1 · Test helper accepts arbitrary recipient** (blocking)

The test-only LoopbackTestTransport parses BB_APPS_E2E_CONTROL_PLANE_URL but does not validate that it is an HTTP loopback origin, has an explicit port, lacks userinfo, or otherwise belongs to the parent test server. It then replaces the approved production URL with that arbitrary URL while retaining Authorization: BBIdentity. The self-spawned test helper is activated by environment variables and does not itself require isolated BB_HOME/auth storage. Although cfg(test) keeps this path out of the distributable bb binary, a crafted developer or CI test-binary invocation can make the helper read normal stored auth and send the reusable session to an attacker-controlled recipient.

**User effect:** A developer or CI identity can be disclosed when an untrusted environment or command invokes the test executable with a malicious control-plane URL.

**Recommended fix:** Validate the test transport destination at first ingress as an HTTP loopback IP origin with an explicit port and no userinfo, path, query, or fragment. Require the helper to receive and verify an isolated temporary auth-storage path and preferably a parent-generated nonce before running any command.

**Test:** Attempt to start the process helper with remote HTTP/HTTPS hosts, localhost names, userinfo, paths, queries, and missing ports and assert rejection occurs before auth storage is read or a request is sent. Add a positive test for the parent-created loopback server and isolated storage.

bb-cli/src/bb/apps.rs
```diff
@@ -827,6 +653,42 @@ impl ControlPlaneClient {
...
+        Value::Array(items) => {
+            for item in items {
+                redact_json_value(item, credential);
+            }
+        }
+        Value::Object(object) => {
+            let mut sanitized = Map::new();
+            for (key, mut value) in std::mem::take(object) {
+                redact_json_value(&mut value, credential);
+                sanitized.insert(credential.redact(&key), value);
```
morgmart: 🤖 **P2 · Redacted keys can drop fields** (blocking)

redact_json_value rebuilds every object by replacing the credential inside each key and inserting the result into a new map. Two distinct source keys can therefore sanitize to the same key—for example, a key equal to the credential and an existing [REDACTED] key. The later insertion silently overwrites the earlier value, changing field count and potentially associating the wrong value with the surviving key. Current coverage checks credential reflection in values but not keys or collisions.

**User effect:** Contract, plan, initialize, or deploy output can silently lose or misattribute fields when an approved service reflects the session in an object key, making successful command output structurally unreliable.

**Recommended fix:** Preserve object cardinality during sanitization. Reject successful responses containing the credential in a key, or replace secret-bearing keys with deterministic unique placeholders and fail if any collision remains; never silently overwrite a field.

**Test:** Return nested objects and arrays containing both a credential-bearing key and its would-be redacted key. Assert the command either rejects the unsafe response or preserves every field under unique sanitized keys, with no secret in output and no overwritten values.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found 2 blocking findings and 0 non-blocking findings. The cfg(test) process helper accepts an arbitrary BB_APPS_E2E_CONTROL_PLANE_URL and rewrites a credential-bearing approved request to that destination without enforcing loopback or isolated auth storage, so a crafted test invocation can exfiltrate a developer or CI session. Separately, recursive successful-response redaction rebuilds JSON objects using redacted keys; distinct original keys can collapse to the same sanitized key and silently overwrite a field. All eleven supplied resolved automated threads were re-evaluated against current code. Their underlying issues are fixed: the distributable bb target no longer includes the feature-gated DNS/CA bypass, Apps is noninteractive, credential selection avoids a second read, failed workspace validation precedes persistence, successful built-binary flow coverage exists in the test executable, and successful response strings are now redacted. Neither current finding duplicates those threads. Final self-check covered contract, create/initialize, create-without-initialize, deploy, login, status, logout, legacy-cache, and cfg(test) process-helper flows; CLI/browser accessibility and localization applicability; destination, redirect, navigation, consent, and interaction guards; async, timeout, error, never-completes, lifecycle, and race behavior; success/error redaction and response-shape integrity; test honesty; project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No frontend product UI changed, so responsive layout, application design-system tokens, and screenshots are not applicable. Supplied GitHub evidence and all eleven supplied threads were inspected. Evidence is structurally available, reports no completed check runs and a pending combined status; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 996fc255. Both test endpoints are now validated at first ingress as explicit-port HTTP loopback IP origins with no userinfo, path, query, or fragment; remote IPs, HTTPS, localhost names, and malformed origins are rejected. The helper no longer accepts or inherits BB_HOME/auth storage: it creates its own temporary home and file store and permits only an explicitly supplied synthetic apps-e2e-only credential. Focused rejection tests cover every unsafe origin class, while the process flows cover the positive isolated path.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 996fc255. Successful responses now preserve object keys exactly and reject the response if any key contains the session, so sanitization cannot collapse two fields or silently overwrite data. String values remain recursively redacted. A focused regression response includes both a credential-bearing key and an existing [REDACTED] key and asserts a secret-free failure instead of a lossy value.

morgmart (approved):
🤖 Automated code review

APPROVE. A fresh static review of the exact three-dot comparison found no publishable blocking or non-blocking findings. All thirteen supplied resolved automated threads were re-evaluated against current code and their underlying issues are fixed: Apps never starts implicit browser authentication; verification and transmission use the same credential instance; failed workspace validation occurs before persistence; 401 recovery replaces the rejected session; the destination is exact-host HTTPS with redirects disabled; no test transport is present in the distributable bb target; the cfg(test) process helper validates explicit-port loopback IP origins and creates isolated synthetic auth storage; successful built-binary-equivalent process flows cover contract, both create branches, and deploy; successful response string values are recursively redacted; and secret-bearing object keys are rejected without mutating response shape. Final self-check covered contract, create/initialize, create-without-initialize, deploy, login, status, logout, legacy-cache, and cfg(test) process-helper flows; accessibility and localization applicability; navigation, recipient, redirect, consent, and interaction guards; async, timeout, error, never-completes, lifecycle, and race behavior; success/error redaction; test honesty; design-system/project rules; duplicate and overlap handling; and evidence plus user effect for blocking severity. No frontend product UI changed, so responsive layout, application design-system tokens, and screenshots are not applicable. Supplied GitHub evidence and all thirteen threads were inspected. Evidence is structurally available, reports no completed check runs and a pending combined status; required checks still independently govern merge readiness.

PR #41, opened 2026-08-14
send queued messages in the background when the chat is not open

tulsi-builder:
**Category:** fix
**User Impact:** Queued messages now send automatically once the agent is free, even if you've navigated away from the chat.

**Problem:** Queuing a follow-up message and then leaving the chat silently stranded it — the queue only drained while the chat was open, so the message sat unsent until you happened to reopen that session. **Solution:** The always-mounted released-queue drain is generalized into a background queue drain: the open chat registers as the queue's foreground owner, and when it unmounts the background drain takes over and sends the head once the session is ready. The existing dispatch leases and ownership assertions still guarantee at most one send attempt per message (`LAWS/CHAT.md`).

<details>

**src/features/chat/lib/foregroundQueueOwnership.ts** (new)
Small registry where a mounted, interactive chat claims ownership of its session's queue. This is the coordination point that decides whether the foreground hook or the background drain sends a queued head — exactly one drain claims each head by design.

**src/features/chat/hooks/useBackgroundQueuedMessageDrain.ts** (renamed from useReleasedQueuedMessageDrain.ts)
The global drain now also claims ordinary queued heads when no foreground chat owns the session, and reacts when ownership is released (user leaves the chat). Hardened from review: reclaimed session-window queues are refreshed from persistence before draining (a stale in-memory copy can't re-send a head the window already sent or dismissed); restored-from-relaunch heads are excluded per record id until the user actually reopens the chat; and background failures raise an error toast since the user isn't looking at the failed session.

**src/features/chat/hooks/useMessageQueue.ts**
Registers as the foreground queue owner while mounted and not read-only. Its berdctl-origin check moved to the shared helper.

**src/features/chat/lib/queuedMessageOrigin.ts** (new)
Single shared definition of the berdctl cross-session origin check, previously duplicated in two modules. berdctl sends keep their dedicated drain and are excluded from both chat drains.

**src/features/berdctl/commands/runtime/sessionSend.ts**
Re-exports the shared origin check instead of defining its own copy.

**src/features/chat/ui/BackgroundQueuedMessageDrain.tsx** / **src/features/chat/ui/ReleasedQueuedMessageDrain.tsx**
Component renamed to match the hook's broader role; mount sites unchanged in behavior.

**src/main.tsx** / **src/app/SessionWindowApp.tsx** / **src/app/main.berdctl.test.ts**
Updated imports/assertions for the rename.

**src/features/chat/hooks/useBackgroundQueuedMessageDrain.test.tsx**
Existing released-head coverage kept; adds discriminating tests for foreground-owner deferral and handoff, window-reclaim persistence refresh, restored-head exclusion surviving incidental flag clears, and the failure toast. The two behavioral fixes were verified to fail against the pre-fix implementation.

Renames `queue.releasedSendFailed` to `queue.backgroundSendFailed` (single usage) and adds a title for the background failure toast.

</details>

### Reproduction Steps

1. Open a chat and send a prompt so the agent is busy responding.
2. While it responds, type a follow-up — it lands in the queue above the composer.
3. Navigate away: go Home, open another chat, or any other view.
4. When the agent finishes, the queued message sends on its own — return to the chat to see it as the next user turn. Before this change it stayed stuck in the queue until the chat was reopened.
5. Failure path: if the background send fails, an error toast appears wherever you are, and the message parks in the queue in a retryable failed state.


src/features/chat/hooks/useBackgroundQueuedMessageDrain.ts
```diff
@@ -33,10 +41,115 @@ type ContentionWaiter = {
...
+  const reclaimedSessionIds = Object.keys(previousOpenSessions).filter(
+    (sessionId) => !(sessionId in openSessions),
+  );
+  if (reclaimedSessionIds.length === 0) {
+    drainReadyQueuedMessages();
+    return;
+  }
+  const sequence = ++ownershipRefreshSequence;
+  const persistedQueues = await loadPersistedMessageQueues();
+  if (sequence !== ownershipRefreshSequence) return;
```
morgmart: 🤖 **P1 · Overlapping closes can replay stale queues** (blocking)

The global sequence drops every reclaim refresh except the latest, but each refresh reconciles only the sessions reclaimed by its own window-snapshot transition. If window A closes and starts a persistence load, then window B closes before A's load resolves, A's result is discarded. B's refresh reconciles only B and then drains all currently unwindowed sessions, including A from the main renderer's explicitly stale in-memory copy. A head already sent, edited, or dismissed in A's window can therefore be dispatched from that stale copy. This violates LAWS/CHAT.md's at-most-one-user-turn requirement.

**User effect:** Closing two detached chat windows close together can resend an old queued prompt or send text the user already edited or dismissed, producing an unintended duplicate user turn.

**Recommended fix:** Serialize reclaim reconciliation or accumulate reclaimed session IDs across overlapping refreshes, and do not drain any reclaimed session until a fresh persistence read has reconciled every still-reclaimed session. Re-read ownership before applying the result, and preserve newer transitions instead of discarding an earlier session's required reconciliation.

**Test:** Use deferred persistence promises: close window A, close window B before A's read resolves, resolve the reads out of order with persistence showing A's old head removed or changed, and assert the stale A record is never sent while both reclaimed queues are reconciled correctly.

src/shared/i18n/locales/en/chat.json
```diff
@@ -411,7 +411,8 @@
     "dismiss": "Dismiss queued message",
     "edit": "Edit queued message",
     "sendAnyway": "Send anyway",
-    "releasedSendFailed": "The queued message could not be sent. Retry it or edit the message.",
+    "backgroundSendFailed": "The queued message could not be sent. Retry it or edit the message.",
+    "backgroundSendFailedTitle": "Queued message failed",
```
morgmart, line 415: 🤖 **P2 · Localize the new failure toast** (non-blocking)

The new background-send failure title and message are added only to the English chat catalog. Spanish is a supported locale, but its chat catalog has neither key, so i18next falls back to English for this newly visible error flow. The repository's general i18n check does not enforce locale-key parity, so the passing CI evidence does not cover this gap.

**User effect:** A user running Berd in Spanish sees the background queued-message failure toast in English at the moment they need recovery guidance.

**Recommended fix:** Add Spanish translations for queue.backgroundSendFailed and queue.backgroundSendFailedTitle, and add locale-parity coverage for these queue error keys.

**Test:** Switch i18n to Spanish, trigger the background-send failure path, and assert both the toast title and description use the Spanish catalog rather than English fallback text.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Static review of the exact three-dot range found one blocking queue-replay race and one non-blocking localization gap. The supplied GitHub evidence was inspected: all nine check runs for the exact head SHA completed successfully, while the combined commit status was pending with no legacy statuses. Final self-check covered the ordinary background-send and foreground handoff flows, detached-window ownership and reclaim, restored queues, berdctl-origin exclusion, contention and failure handling, navigation/consent guards, accessibility, i18n, async error/never-completes/lifecycle/race behavior, test honesty, and project laws/design-system rules. Findings were consolidated by underlying issue and have concrete changed-line evidence and user effects.

src/features/chat/hooks/useBackgroundQueuedMessageDrain.ts
morgmart: 🤖 Fixed in ce14a6e9. Reclaim reconciliation is now shared by the background and Berdctl drains, serialized across overlapping window closes, and blocks each reclaimed session until its persisted queue has been reconciled after re-checking current window ownership. Added a deferred-persistence two-window regression test proving neither stale head sends.

src/shared/i18n/locales/en/chat.json
morgmart: 🤖 Fixed in ce14a6e9. Added Spanish copy for both background queue failure strings and focused locale coverage that requires both keys to exist and differ from English.

morgmart (approved):
🤖 Re-reviewed at ce14a6e9. The blocking overlapping-reclaim race is fixed through shared serialized reconciliation with regression coverage, and the Spanish failure-toast gap is closed with focused locale coverage. Local `just check` and the full 6,219-test suite pass.

PR #47, opened 2026-08-15
fix: preserve transcripts after invalid replay refreshes

morgmart:
**Category:** fix
**User Impact:** Chats keep their visible history and show a recoverable error when a reload or compaction refresh does not return usable conversation messages.

**Problem:** A replay buffer could exist but sanitize to an empty transcript, causing reload and compaction flows to replace visible history with nothing. Compaction could also claim success even though replacement history was never received.

**Solution:** Centralize replay replacement validation at the session replay boundary. Empty replay data can represent a genuinely empty session only when no history is expected; otherwise the prior transcript and message metadata are preserved, the operation reports failure, and the existing inline error-notification pattern explains that refresh history was not received.

<details>

**src/features/chat/lib/sessionReplayReplacement.ts**
Adds the shared replay consumption and replacement policy used by session loading and compaction.

**src/features/chat/lib/sessionActivation.ts**
Treats unusable replay data as a failed refresh, preserves prior message metadata, and surfaces a recoverable inline error.

**src/features/chat/lib/__tests__/sessionActivation.test.ts**
Covers populated forced reloads and cold loads whose metadata indicates expected history.

**src/features/chat/hooks/useChat.ts**
Requires usable replacement history before reporting compaction success or adding its success notice.

**src/features/chat/hooks/__tests__/useChat.compaction.test.ts**
Verifies invalid post-compaction replay keeps prior messages, returns failure, and does not add a false success notice.

**src/features/chat/hooks/__tests__/useChatSessionController.targetLeaseCompaction.test.ts**
Makes successful auto-compaction fixtures emit truthful replacement history.

**src/app/AppShell.navigation.test.tsx**
Makes the missing-folder recovery fixture provide the history its session metadata declares.

Adds user-facing recovery copy for incomplete load and compaction replay results.

**src/shared/i18n/locales/es/chat.json**
Adds Spanish recovery copy for incomplete load and compaction replay results.

</details>

### Validation

- `pnpm vitest run src/features/chat/lib/__tests__/sessionActivation.test.ts src/features/chat/hooks/__tests__/useChat.compaction.test.ts src/features/chat/hooks/__tests__/useChatSessionController.targetLeaseCompaction.test.ts`
- `just check`
- `just test` (532 files, 6,226 passed, 1 skipped)

### Laws

Reviewed `LAWS/README.md`, `LAWS/CHAT.md`, and `LAWS/AGENTS.md`. This change does not alter composer queue or agent invocation laws.

### DIFF
--- a/src/app/AppShell.navigation.test.tsx
+++ b/src/app/AppShell.navigation.test.tsx
@@ -18,6 +18,8 @@ import { getAppNavigationController } from "@/features/berdctl/navigation";
 import { resetAgentBuilderSourceLifecycleForTests } from "@/features/agents/lib/agentBuilderSourceLifecycle";
 import { useChatStore } from "@/features/chat/stores/chatStore";
 import { useChatSessionStore } from "@/features/chat/stores/chatSessionStore";
+import { ensureReplayBuffer } from "@/features/chat/hooks/replayBuffer";
+import { createUserMessage } from "@/shared/types/messages";
 import { useSessionWindowStore } from "@/features/chat/stores/sessionWindowStore";
 import type { ChatSession } from "@/features/chat/stores/chatSessionStore";
 import type { Message } from "@/shared/types/messages";
@@ -1770,6 +1772,11 @@ describe("AppShell global navigation", () => {
       activeSessionId: null,
     });
     mockCheckDirectoriesExist.mockResolvedValue(["/missing/session"]);
+    mockAcpLoadSession.mockImplementationOnce(async () => {
+      ensureReplayBuffer("missing-session").push(
+        createUserMessage("Existing history"),
+      );
+    });
 
     renderAppShell();
 
@@ -1787,8 +1794,8 @@ describe("AppShell global navigation", () => {
 
     const messages =
       useChatStore.getState().messagesBySession["missing-session"] ?? [];
-    expect(messages).toHaveLength(1);
-    expect(messages[0]?.content[0]).toMatchObject({
+    expect(messages).toHaveLength(2);
+    expect(messages[1]?.content[0]).toMatchObject({
       type: "systemNotification",
       notificationType: "warning",
       action: { type: "openContextPanel" },
--- a/src/app/AppShell.tsx
+++ b/src/app/AppShell.tsx
@@ -142,9 +142,9 @@ import {
 } from "@/features/chat/stores/chatSessionOperations";
 import {
   activateSession as activateChatSession,
-  hasConversationMessages,
   loadSessionMessagesAndPrepare,
 } from "@/features/chat/lib/sessionActivation";
+import { hasConversationMessages } from "@/features/chat/lib/sessionReplayReplacement";
 import {
   focusSessionWindow,
   releaseSession,
--- a/src/features/chat/hooks/__tests__/useChat.compaction.test.ts
+++ b/src/features/chat/hooks/__tests__/useChat.compaction.test.ts
@@ -171,6 +171,53 @@ describe("useChat compaction", () => {
     ).toBe(false);
   });
 
+  it("keeps the transcript and reports failure when compacted replay is invalid", async () => {
+    mockAcpLoadSession.mockImplementation(async (sessionId: string) => {
+      ensureReplayBuffer(sessionId).push(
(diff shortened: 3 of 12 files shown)


cynfria:
🤖 AI review found one retry-path issue:

**Pinned metadata is discarded after an invalid replay** (`src/features/chat/lib/sessionActivation.ts:421-425`). For a pinned placeholder, `messageCountAtRequest` may be stale (often `0`), while refreshed session info correctly reports nonzero history and causes the empty replay to fail. This branch then restores the stale count and clears `pinnedLoadState`. On the next load, the same empty replay can be accepted as a genuinely empty session because the count is now `0` and the visible error is system-only.

Could we preserve the refreshed nonzero `messageCount`, or retain `pinnedLoadState` so the retry refreshes authoritative metadata again? A regression test with two consecutive invalid replays—initial pinned count `0`, refreshed count nonzero—would protect this path.

Everything else in the replay replacement and compaction failure handling looked sound; targeted tests passed locally (49/49).

morgmart:
🤖 Addressed the review feedback in 6d0213e2. Refreshed backend `messageCount` now remains authoritative across invalid replay retries; replay validity uses an explicit empty/nonempty/unknown expectation, including unknown when pinned metadata refresh fails. I added consecutive invalid-replay coverage for a pinned placeholder and removed sanitized renderer replay length as a writer of backend session metadata.

I also separated committed compaction from transcript refresh: an incomplete/failed refresh preserves the prior transcript and warns, while the successfully compacted backend remains ready so a queued message continues in accordance with the chat queue laws.

loganj:
🤖 Independent review at head `6d0213e2` (loganj's agent team; read-only, findings independently verified twice).

## Recommendation: do not approve yet — one blocking retry-path gap. Everything else checks out.

**Blocking finding: a failed pinned-metadata refresh is safe only once; the next retry can accept the same invalid empty replay.**

- `sessionActivation.ts:394-404` handles the first failure correctly: when `acpGetSessionInfo` fails for a pinned placeholder, `historyExpectation` becomes `"unknown"` and the empty replay is rejected.
- But the invalid branch unconditionally clears `pinnedLoadState` (`sessionActivation.ts:~422`). If the placeholder carried a stale `messageCount: 0`, the retry no longer performs the metadata lookup (`:336-342`) and derives `historyExpectation = "empty"` from the stale count — `sessionReplayReplacement.ts:34-51` then treats the empty buffer as authoritative and accepts it.
- **Deterministic proof:** a temporary test (pinned placeholder with `messageCount: 0`; `acpGetSessionInfo` rejecting on every call; empty replay buffer; `loadSessionMessages` called twice) returns `false` on the first call and `true` on the second at this head. The same test passes with the minimal fix below.
- User effect: after one visible recoverable error, reopening/retrying can silently mark an existing conversation as loaded with no history — the exact failure class this PR exists to prevent, on the branch the second commit didn't cover.

**Suggested fix:** on invalid replay, retain the pinned/unknown state when the authoritative metadata lookup failed (or persist an explicit unknown-history expectation), so retries re-fetch metadata and can never downgrade unknown → empty. Add a two-consecutive-failed-lookups regression test. Happy to share our repro test.

**What we verified works at this head:**

- Behavior preservation: the empty-replay protections this PR reimplements hold — replay/compaction suites 48/48, focused empty-replay cases 2/2; an empty replay preserves the transcript and warns instead of wiping it. Also validated in a running renderer against the original fix it reimplements (squareup/berd#1182): identical protected behavior.
- The second commit's retry fix is correct for the refreshed-nonzero branch; compaction correctly distinguishes committed backend compaction from a failed transcript refresh; error notices are deduplicated and localized (en/es).
- `just check` passes; the four affected suites pass 65/65.

One fix + one regression test and this is an approve from us.

morgmart:
🤖 Addressed the remaining retry-path finding in 2c17db06. When pinned metadata refresh fails and replay is invalid, the session now retains `pinnedLoadState: "failed"`; the next retry therefore re-fetches metadata and keeps history expectation unknown instead of trusting the placeholder zero. Added the deterministic two-consecutive-failed-lookups regression test. Targeted tests, `just check`, and the full 6,226-test suite pass locally.

cynfria (approved):

PR #61, opened 2026-08-17
fix(chat): hold queued messages until their session finishes creating

matt2e:
## Problem

Starting a chat from the global composer reliably surfaced "The queued message could not be sent."

The background queued-message drain reacts to the store write immediately, while `ChatView`'s mount — and the foreground gate that *does* check creation state — is delayed for the composer animation. So the drain claimed the brand-new chat's first message while backend session creation was still in flight and dispatched it against the renderer-local draft UUID. The backend answered "Resource not found", the drain parked the head as failed, and the toast fired.

## Fix

- **Gate the shared attemptability predicate on `creationState == null`** (`queuedMessageAttemptOwnership.ts`). Every drain and store-driven wake-up funnels through it, so the head is held instead of sent. Promotion clears `creationState` and swaps in the backend id, which notifies the session store and drains the message against the real id.
- **Back that up in the send path** so no caller can reach the wire with a not-yet-created session. `acquireExistingSessionForBackgroundSend` returns a `creation-incomplete` status instead of trusting `loadSessionMessages`, which reports success for a pending draft ("nothing to load" is not a failure there) and lets preparation walk into `acpApi.loadSession`. Both background senders throw `SessionDispatchCreationIncompleteError`, a `PreCommitSendRejectedError` the drains already swallow silently.
- **berdctl `session send` handles it explicitly**, so `--if-running queue` holds the prompt and the refusal says the chat is still being created rather than blaming a run that isn't happening.
- **Creation failures now toast at the creation site** when the failed chat isn't the active session, reporting the actual problem ("Session failed to start") where the user is, instead of leaving the in-transcript error invisible until they open the chat.

## Laws

Conforms the background drain to `LAWS/CHAT.md`: a message MUST NOT be sent until its session is ready, MUST remain queued until its session begins processing it, and the queue MUST resume when the session becomes ready.

## Tests

New coverage for the acquisition hold, the queued-send rejection, both drains holding then sending after promotion, and the berdctl runtime refusing to reach the wire.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app/AppShell.tsx
+++ b/src/app/AppShell.tsx
@@ -44,7 +44,10 @@ import { planProjectChatWorkspacesAsIs } from "@/features/projects/lib/projectCh
 import { ProjectWorkspaceStartupNameDialog } from "@/features/projects/ui/ProjectWorkspaceStartupNameDialog";
 import { useWorkspaceRepository } from "@/features/workspaces/workspaceRepository";
 import type { TopBarChromeInsets } from "./ui/TopBar";
-import { useChatStore } from "@/features/chat/stores/chatStore";
+import {
+  isSessionActivelyViewed,
+  useChatStore,
+} from "@/features/chat/stores/chatStore";
 import { useActiveProjectTint } from "@/features/chat/hooks/useActiveProjectTint";
 import { useWorkspaceNameRequestQueue } from "@/features/chat/hooks/useWorkspaceNameRequestQueue";
 import {
@@ -2008,10 +2011,7 @@ export function AppShell({
           // `activeSessionId` pointing here, so the store's viewing flag is
           // what decides whether the in-transcript error is on screen.
           const reportCreationFailureIfHidden = (message: string) => {
-            const isViewingFailedSession =
-              useChatSessionStore.getState().activeSessionId === session.id &&
-              useChatStore.getState().isViewingActiveSession;
-            if (isViewingFailedSession) {
+            if (isSessionActivelyViewed(useChatStore.getState(), session.id)) {
               return;
             }
             toast.error(t("chat:toolbar.sessionStartFailed"), {
--- a/src/app/AppShell.tsx
+++ b/src/app/AppShell.tsx
@@ -1998,6 +1998,19 @@ export function AppShell({
         )
         .catch(async (error) => {
           const chatStore = useChatStore.getState();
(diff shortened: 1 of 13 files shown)


loganj:
🤖 This review comes from loganj's automated review team (multiple independent agent reviewers), not from Logan personally.

**Verdict: PASS across six independent reviews (8.5–9/10) at head `19a40e81d1d0c8191814452691b3dd5c7ca351d4`.** No merge-blocking defect found. Nice work — the race is closed on every drain path.

What we verified:
- **Centralized gate:** `isQueuedMessageTargetAttemptable` now requires `creationState == null` (`src/features/chat/lib/queuedMessageAttemptOwnership.ts`), and all three queue owners (foreground `useMessageQueue`, background drain, berdctl drain) consume it before dispatch. A repo-wide caller inventory found no production path bypassing the gate.
- **Send-boundary backstop:** `acquireExistingSessionForBackgroundSend` returns `creation-incomplete` before hydration or wire access, and `SessionDispatchCreationIncompleteError` is scoped to the pre-existing pre-commit rejection family — so a rejected attempt retains the queued head, and real transport errors still park + toast rather than being swallowed.
- **Behavioral proof:** at the exact head, the four queue/send suites pass 98/98 — hold-while-pending, resume-on-promotion via `becameQueuedMessageTargetAttemptable`, failed-creation parks visibly (no forever-hang: retry resets failed → pending), and no wire call on rejection.
- **Promotion handoff:** checking ownership under both the promoted backend ID and the retained `clientSessionId` correctly closes the ID-swap window; the lease-across-hydration extension closes the adjacent replay-vs-live race and reads as necessary, not scope creep.

Two non-blocking notes:
- The PR body's claim about berdctl `session send` reporting "still being created" and queueing with `--if-running queue` is slightly stale — the command resolves via backend `acpGetSessionInfo` first, so a renderer-local draft ID surfaces as session-not-found rather than the described message. Worth a one-line body tweak.
- Consider a focused test asserting the drain re-entry guard on `shouldResumeDrain`, though existing coverage is strong.

loganj (approved):
🤖 Approving on behalf of @loganj per his explicit instruction (GitHub UI unavailable to him during the outage). Verdict basis: unanimous PASS from his six-agent automated review team at head 19a40e81 — see the consolidated review comment above.

PR #62, opened 2026-08-17
refactor(acp): migrate session deletion to the standard session/delete method

matt2e:
## Summary
The pinned Goose backend now exposes session deletion as the standard ACP `session/delete` method instead of declaring it in its goose extension schema. This PR migrates Berd accordingly:

- Bumps `goose-backend.lock.json` (`11deb564` → `063694cf`) and regenerates the vendored SDK (`sdk/schema/*`, `sdk/src/generated/*`) from the new schema. The goose-extension `session/delete` declaration, its `DeleteSessionRequest` type, and the generated `client.goose.sessionDelete()` wrapper are removed by the regen.
- Switches `deleteSession()` in `src/shared/api/acpApi.ts` to call the standard method directly via the generic passthrough: `client.extMethod("session/delete", { sessionId })`.
- The regen also picks up new unstable `ProviderReadinessCheck` request/response types from the backend bump; the app does not use them yet.

### Related issue
N/A — none found.

### Testing
- `just check` passes (formatting, lint, i18n, and TypeScript checks, including the vendored SDK build).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


loganj:
🤖 This review comes from loganj's automated review team (multiple independent agent reviewers), not from Logan personally.

**Verdict: PASS across six independent reviews (8–9/10) at head `d380507a988d5a4d8278def9110509a11d680f14`.** No blocker or required change — a clean, faithful migration. One conscious acceptance for the merger, noted below.

What we verified:
- **Wire equivalence:** the removed generated `sessionDelete()` wrapper was itself just `await this.conn.extMethod("session/delete", params)`, so the new direct call at `src/shared/api/acpApi.ts:382-385` sends the identical request with identical error propagation; the `Promise<void>` caller contract is preserved. At goose pin `063694cf` the standard handler performs the same storage/session/agent-manager deletion; only the empty success-response type changed, which Berd discards.
- **Regen is genuinely mechanical:** two reviewers independently reproduced the generated SDK byte-for-byte from the vendored schema (`pnpm --dir sdk build` + clean `git diff` on `sdk/src/generated`), and the checked-in schema JSON is semantically equal to `063694cf:crates/goose/acp-{schema,meta}.json`. No hand edits.
- **`extMethod` is the right seam today:** Berd pins TS ACP SDK 0.19.2, which has no typed `deleteSession` (it first appears at 0.22). Suggested follow-up (not this PR): when the SDK is upgraded, expose a typed `deleteSession` to restore compile-time param checking on the `{ sessionId }` payload.
- **No dangling references:** three independent repo-wide sweeps found zero remaining `sessionDelete` / `DeleteSessionRequest` uses; the only wire-method occurrence is the intended one.

The acceptance: `goose-backend.lock.json` advances `11deb564 → 063694cf` — 20 commits, 229 files, ~+7,107/−5,532, including agent-loop, provider/OAuth, and computer-controller changes. Not a defect of this PR, but merging it is a broad backend roll-forward rather than a two-line protocol refactor; we'd suggest smoking the packaged sidecar (initialize → new/prompt → delete) before the next release. A focused unit test guarding the `extMethod("session/delete", { sessionId })` literal would also be cheap insurance.

loganj (approved):
🤖 Approving on behalf of @loganj per his explicit instruction (GitHub UI unavailable to him during the outage). Verdict basis: unanimous PASS from his six-agent automated review team at head d380507a — see the consolidated review comment above.

PR #63, opened 2026-08-17
feat(berdctl): expose automations as a gated `automation` noun

NickTitle:
## What

Adds a berdctl `automation` noun (registry group `automations`) with four verbs:

| verb | effect / visibility | what it does |
|---|---|---|
| `automation list` | read / none | Summaries (id, title, schedule, status, pause state, last success), `--query` title filter |
| `automation get` | read / none | One automation's full detail, including instructions and latest-run wiring |
| `automation create` | create / immediate | Create a scheduled automation (generic summary tile type) |
| `automation run` | update / discoverable | Trigger one immediate run, off schedule |

```
$ berdctl automation --help
Manage scheduled automations: list, get, create, run
```

## Why

Agents running inside Berd can already manage sessions/projects/skills via berdctl, but automations were UI-only. This lets an agent set up and operate recurring jobs (for example a scheduled PR-review sweep) through the app instead of ad-hoc local cron/schedulers.

## How it is gated

Follows the `feedback` precedent for gated distribution surfaces, since Automations depend on a backing service:

- **Renderer registry** (the trust boundary) includes the group only when `VITE_AUTOMATIONS === "1"`.
- **Every action** also refuses at `execute()` via the `automations` profile capability (`automations_disabled`), mirroring how the Automations view is gated at runtime.
- **Contract artifacts**: one variant per gate combination (public / feedback / automations / block), selected by matching cargo features (`block-feedback`, `block-automations`). The embedded CLI can never advertise a noun its renderer build filters out of `TOOL_GROUPS`, in any flag combination. Public artifacts are byte-identical to before.
- **Every berdctl build path** maps `VITE_AUTOMATIONS` → `block-automations`: dev justfile, `dev-e2e.sh`, `prepare-berdctl-sidecar.sh`, `Stage-Sidecar-Windows.ps1`, and the macOS release build. Release-script tests pin each seam.

## Scope decisions

- **No `delete` / `pause` verbs.** Delete is destructive and outside the v1 no-auth model; pause/resume is not currently a first-class single mutation in the UI API surface (`schedulePaused` is read-only on the tile). Both stay in the UI for now.
- `create` pins the generic summary tile type (`4`, same as the Automations builder); dashboard/Builderbot tile types are deliberately out of scope, matching the view's own filtering.
- `run` resolves the automation first so a bad id reads as `automation_not_found` rather than a backend refresh failure.

## Tests

- Behavior tests for all four commands: capability refusal before any backend call, not-found with the fixing command named, backend rejections surfaced as stable command errors, run session id passthrough.
- Schema-bounds coverage plus valid-args fixtures (strict-parse totality).
- Rust: wire-mapping minimal invocations for the new verbs; top-level help asserts `automation` appears iff `block-automations` (independent of feedback).
- `just check` green; `cargo test -p berdctl` green on **all four** feature combinations; clippy/fmt green; contracts regenerated byte-stable; release-script seam tests green (85).

## Review

Builderbot local review loop: round 1 flagged variant coupling + a missed release-build gate (both fixed: per-combination artifacts, full build-path propagation); round 2 asked for seam test coverage (added); final round: **0 findings, "patch is correct."**

🤖 Authored by Nick's AI agent; built against the berdctl-new-command SKILL and docs/berdctl-architecture.md.

### DIFF
--- a/justfile
+++ b/justfile
@@ -511,7 +511,8 @@ dev:
     # member needs an explicit build, resolved at runtime via BERDCTL_BIN
     # because tauri.dev.conf.json blanks externalBin.
     BERDCTL_FEATURES=()
-    [[ "${VITE_FEEDBACK:-0}" == "1" || "${VITE_AUTOMATIONS:-0}" == "1" ]] && BERDCTL_FEATURES+=(--features block-feedback)
+    [[ "${VITE_FEEDBACK:-0}" == "1" ]] && BERDCTL_FEATURES+=(--features block-feedback)
+    [[ "${VITE_AUTOMATIONS:-0}" == "1" ]] && BERDCTL_FEATURES+=(--features block-automations)
     # ${arr[@]+...} guards the empty-array expansion, which bash 3.2 (stock
     # macOS) treats as an unbound variable under `set -u`.
     (cd src-tauri && cargo build -p berdctl ${BERDCTL_FEATURES[@]+"${BERDCTL_FEATURES[@]}"})
--- a/justfile
+++ b/justfile
@@ -331,7 +331,7 @@ _bundle-unix:
       GOOSE_DEV_MODE=required GOOSE_BUILD_PROFILE=release ./scripts/ensure-local-goose.sh
     fi
     GOOSE_BUILD_PROFILE=release ./scripts/prepare-goose-sidecar.sh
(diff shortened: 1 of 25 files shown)


NickTitle:
Closing — the workflow this was built for ended up not needing a CLI surface for automations. The branch is complete and green (all four feature-gate variants tested) if anyone wants to pick it up later.

PR #64, opened 2026-08-17
fix(chat): admit messages while sessions start

loganj:
🤖

## Summary

- Lets the composer accept a message into a draft chat's queue while its backend session is still being prepared.
- Keeps dispatch blocked until session preparation completes, including Agent Builder draft identity migration.
- Preserves failed-creation, read-only, Agent Builder preparation, and workspace preparation blocks.

This is PR 1 of 2. [PR #65](https://github.com/block/berd/pull/65) updates the chat laws on top of this behavior.

Adapted from chailandau's [squareup/berd#1076](https://github.com/squareup/berd/pull/1076) for `block/berd`, while retaining the dispatch ownership introduced by [block/berd#61](https://github.com/block/berd/pull/61).

### Related issue

None found. Related pull requests: squareup/berd#1076 and block/berd#61.

### Testing

- `just check`
- `just test` — 6,308 passed, 1 skipped
- Pre-push fmt, frontend check, Tauri check, and Clippy gate
- Focused queue-admission, promotion/migration interleaving, failure, and draft-preservation regressions

### DIFF
--- a/src/features/agents/lib/agentBuilderSession.ts
+++ b/src/features/agents/lib/agentBuilderSession.ts
@@ -22,13 +22,17 @@ import {
   listAgentBuilderSources,
   promoteAgentBuilderDraftSource,
   readFreshAgentSource,
+  updateAgentBuilderSource,
   type DraftAgentDefaults,
+  type PersonaSourcePatch,
 } from "./agentBuilderSourceLifecycle";
 import type { AgentSourceEntry } from "@/shared/api/agents";
 import {
   deriveSlug,
   fileStem,
   isEmptyPlaceholderDraft,
+  isPlaceholderDraftForSession,
+  placeholderAgentName,
 } from "./agentBuilderIdentity";
 export {
   deriveSlug,
@@ -367,6 +371,50 @@ export async function preSeedDraftAgent(
     provider,
     modelSelection,
   });
+}
+
+export async function migratePendingDraftAgent(
+  draftSessionId: string,
+  backendSessionId: string,
+  targetPath?: string | null,
+): Promise<void> {
+  const resolvedTargetPath =
+    targetPath ??
+    useChatSessionStore.getState().getSession(backendSessionId)
+      ?.targetAgentPath;
+  if (!resolvedTargetPath) {
+    return;
+  }
+
+  const source = await findAgentBuilderSource(
+    draftSessionId,
+    resolvedTargetPath,
+  );
+  if (!source) {
+    throw new Error(
+      "Pending Agent Builder target could not be verified before session promotion.",
+    );
+  }
+  if (source.properties?.draft !== true) {
+    return;
+  }
+  if (source.properties.builderSessionId !== draftSessionId) {
+    throw new Error(
+      "Pending Agent Builder draft belongs to a different session.",
+    );
+  }
+
+  const patch: PersonaSourcePatch = {
+    properties: {
+      ...source.properties,
+      builderSessionId: backendSessionId,
+    },
+  };
+  if (isPlaceholderDraftForSession(source, draftSessionId)) {
+    patch.name = placeholderAgentName(backendSessionId);
+  }
+
+  await updateAgentBuilderSource(source.path, patch);
 }
 
 export async function recoverDraftAgent(
--- a/src/app/AppShell.navigation.test.tsx
+++ b/src/app/AppShell.navigation.test.tsx
@@ -15,12 +15,14 @@ import { i18n } from "@/shared/i18n";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { useAgentStore } from "@/features/agents/stores/agentStore";
 import { getAppNavigationController } from "@/features/berdctl/navigation";
+import { placeholderAgentName } from "@/features/agents/lib/agentBuilderIdentity";
 import { resetAgentBuilderSourceLifecycleForTests } from "@/features/agents/lib/agentBuilderSourceLifecycle";
 import { useChatStore } from "@/features/chat/stores/chatStore";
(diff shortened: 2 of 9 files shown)


src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -369,6 +373,50 @@ export async function preSeedDraftAgent(
...
+    properties: {
+      ...source.properties,
+      builderSessionId: backendSessionId,
+    },
+  };
+  if (isPlaceholderDraftForSession(source, draftSessionId)) {
+    patch.name = placeholderAgentName(backendSessionId);
+  }
+
+  await updateAgentBuilderSource(source.path, patch);
```
morgmart, line 417: 🤖 **P1 · Migration races active draft saves** (blocking)

The new migration reads a draft snapshot and then performs an independent full-source update through updateAgentBuilderSource. Agent Builder autosaves use the same read/merge/full-update boundary but have their own write queue, so migration is not serialized with edits that remain possible while session creation is pending. If an autosave and migration overlap, either writer can apply a stale full source: migration can restore older name/content/properties, or autosave can restore the obsolete draft session ID after migration.

**User effect:** A user who edits an Agent Builder draft while the chat session is starting can silently lose recent edits or end up with a draft still owned by the obsolete session, preventing reliable continuation of the queued request.

**Recommended fix:** Route the identity change through the draft editor's authoritative per-source write owner, or add an atomic/serialized source mutation that re-reads and conditionally changes only builderSessionId (and the placeholder name only if it is still the matching placeholder) immediately before writing.

**Test:** Add a discriminating race test in which an editor save changes the name/content or model properties between migration's read and write, then assert both the edit and the backend builderSessionId survive regardless of completion order.

morgmart, line 417: 🤖 **P1 · Promotion keeps stale renamed path** (blocking)

updateAgentBuilderSource explicitly supports returning an AgentSourceEntry whose path differs after an update, but migratePendingDraftAgent discards that return value. This migration can rename a placeholder, while AppShell and the already-captured Agent Builder send continue promoting and referencing the pre-update targetAgentPath. If the source update moves the file, the successful migration leaves the session and queued assistant prompt pointing to the old path.

**User effect:** The queued Agent Builder request can start against a file that no longer exists, so the user's first message fails to continue the draft that was just created.

**Recommended fix:** Return the authoritative updated source from migration and atomically adopt its path and derived slug in session promotion and queued send options, or avoid a path-changing rename during identity migration.

**Test:** Make updatePersonaSource return a different path after the placeholder rename, then assert the promoted session and dispatched Agent Builder prompt both use the returned path and that no stale path remains.

morgmart (dismissed):
🤖 Automated code review

REQUEST_CHANGES. Static review of the exact three-dot comparison found two blocking Agent Builder migration defects. Final self-check covered pending-session composer admission, ordinary and Agent Builder queue dispatch, draft-ID promotion, creation and migration failure, newer-draft preservation, navigation/consent guards, accessibility, i18n, async never-completes and race behavior, lifecycle, test honesty, design-system/project laws, duplicate overlap, and blocking user-effect evidence. The supplied GitHub evidence was inspected: all 8 captured check runs completed successfully; no local execution was performed. The findings are distinct: one concerns concurrent write/data-loss behavior, while the other concerns adopting a path returned by a successful rename. No review threads were supplied, so no findings were suppressed as duplicates.

src/features/agents/lib/agentBuilderSession.ts
loganj: Confirmed — this race is real at head 21d76631. `migratePendingDraftAgent` reads a full source snapshot and writes back through `updateAgentBuilderSource` directly, outside the editor autosave's per-source write queue, so an overlapping autosave and migration can each apply a stale full source (silent edit loss, or the obsolete draft session ID restored after migration).

**Disposition — resolved structurally rather than patched.** Per review of this PR against morgmart's #68, we're adopting the deferred-preparation architecture: #68 is being rebased onto this stack (#64 → #65 → rebased-#68 → fixes PR), and in that shape the Agent Builder draft is created only *after* session promotion, keyed to the final backend session ID from birth. `migratePendingDraftAgent` — the machinery this race lives in — is deleted outright; there is no migration write left to race the autosave queue. The fixes PR on top also closes a separate mid-preparation queue-record-removal race (revalidate the triggering record across async boundaries, fail closed, with regression).

The whole stack will be flattened before merge, so this defect never lands on main in the interim. Resolving on that basis; the deletion will be verifiable in the stacked PRs shortly.

loganj: Confirmed — real at head 21d76631. `migratePendingDraftAgent` returns void and discards the `AgentSourceEntry` returned by `updateAgentBuilderSource`, so a path-changing placeholder rename leaves session promotion and the queued Agent Builder send referencing the pre-migration `targetAgentPath`.

**Disposition — resolved structurally rather than patched.** Same resolution as the sibling finding: the stack adopts #68's deferred-preparation architecture (#64 → #65 → rebased-#68 → fixes PR), which deletes `migratePendingDraftAgent` entirely. Preparation happens post-promotion against the final session ID and immediately adopts the updater's returned `{path, slug}` — there is no rename-then-discard step left, so no stale path can survive into promotion or the queued send.

The stack flattens before merge, so no interim window ships with this defect. Resolving on that basis; the deleted call sites and the returned-path adoption will be verifiable in the stacked PRs.

loganj:
Changes requested here https://github.com/block/berd/pull/64#pullrequestreview-4954970416 are addressed down-stack in PRs #68 and #78

kalvinnchau (approved):

PR #43, opened 2026-08-14
add bundled starter agents and reliable first-run Home

cynfria:
**Category:** new-feature
**User Impact:** New users start with seven useful agents, with Berdy, Tinker, and Wildcard featured on a reliable first-run Home canvas.

**Problem:** New users previously received only Berdy by default, while optional recommendations and Home starter pins followed separate, fragile setup paths. Resetting onboarding or loading agents slowly could also leave Home incomplete or inconsistently framed.

**Solution:** Bundle the complete starter-agent set and derive first-run and reset Home from one canonical layout. The flow preserves customized Homes, safely handles user-owned filename collisions, recovers delayed starter pins and avatar media, and serializes persistence before marking onboarding complete.

<details>
<summary>File changes</summary>

**distro/agents/berdy.md**
Updates Berdy's bundled definition and black-cube avatar reference.

**distro/agents/choosey.md**
Adds Choosey to the default bundled starter-agent set.

**distro/agents/copycat.md**
Adds Copycat to the default bundled starter-agent set.

**distro/agents/pushback.md**
Adds Pushback to the default bundled starter-agent set.

**distro/agents/sprout.md**
Adds Agt. Builder's current instructions and avatar as a bundled starter agent.

**distro/agents/tinker.md**
Adds Tinker to the default bundled starter-agent set and Home composition.

**distro/agents/wildcard.md**
Adds Wildcard to the default bundled starter-agent set and Home composition.

**release-agents/README.md**
Documents that starter agents are sourced from the distro bundle.

**scripts/release/lib.sh**
Clarifies release staging behavior for the bundled starter-agent set.

**src-tauri/src/services/bundled_agents.rs**
Generalizes safe fallback installation and repair for user-owned filename collisions while preserving bundled identity and updates.

**src/app/AppShell.navigation.test.tsx**
Updates Berdy fixtures for the new avatar identity.

**src/features/agents/hooks/usePersonas.ts**
Coordinates persona loads and mutations across consumers so stale requests cannot overwrite edits.

**src/features/agents/hooks/__tests__/usePersonas.test.ts**
Covers shared initial loads and cross-consumer mutation races.

**src/features/experiments/ExperimentsSettings.tsx**
Consolidates onboarding reset controls into one global reset action.

**src/features/experiments/__tests__/ExperimentsSettings.test.tsx**
Updates reset-control coverage for the consolidated experience.

**src/features/home/onboarding/createStarterHomeWidgets.ts**
Defines the canonical first-run and reset Home composition.

[...]

PR #70, opened 2026-08-17
add clean-install starter agents

cynfria:
## Summary
- bundle and validate seven starter agents for clean Berd installations
- install them through a versioned allocation manifest that preserves collisions, edits, and deletions
- expose only backend-verified managed identities to the frontend
- feature verified Tinker and Wildcard together on eligible new Home canvases
- keep established Homes unchanged and retain one developer onboarding reset

## Deliberately narrow lifecycle
- the complete pending allocation plan is persisted before the first file write
- pending records adopt already-written matching bytes before consulting the current package
- missing pending targets install only while the package still has the exact recorded bytes
- repair restores only an existing managed allocation whose target is missing
- customized, occupied, legacy, corrupt, and unknown states are never repaired or migrated implicitly
- invalid or unsafe manifests fail closed without being renamed or rewritten
- packaged updates never rewrite established copies
- frontend metadata, names, and filenames are not ownership proof

## Scope
This is clean-install-only. It does not migrate established Homes, retrofit established agent installations, recover corrupt manifests, duplicate customized agents, or refresh agents during packaged updates.

## Validation
- `just check`
- `just tauri-check`
- `just clippy`
- bundled-agent Rust tests (18 passed)
- validator tests (19 passed)
- focused frontend tests (75 passed)
- Windows-safe bundle validation entry point
- `git diff --check`

Supersedes #43.


src-tauri/src/services/bundled_agents.rs
```diff
@@ -228,6 +228,10 @@ fn seed_bundled_agents_from_dir(
     entries.sort_by_key(|entry| entry.file_name());
 
     let mut marker = read_seed_marker(target_root)?;
+    let clean_install = !target_root.exists()
```
morgmart: 🤖 **P1 · Use durable clean-install state** (blocking)

`clean_install` is derived from whether the shared target directory currently has any entry. A person launching Berd for the first time after already creating a personal Goose/ACP agent is therefore classified as an existing install, so every new bundled agent has both `clean_install == false` and `was_previously_seeded == false` and is skipped. The loop nevertheless records those filenames in the marker, making the skip permanent. The same logic prevents recovery if first-run copying stops after writing one file but before the marker is committed: the next launch sees a nonempty directory and marks all remaining files as handled without installing them.

**User effect:** New Berd users who already have any agent, or whose first seed is interrupted, can permanently miss Berdy and the six starter agents; Tinker and Wildcard then cannot appear on the intended starter Home.

**Recommended fix:** Determine first-Berd-install eligibility from durable Berd-owned state rather than directory emptiness, and make seeding recoverable per file or stage the complete set transactionally before committing the marker. Continue preserving colliding user-owned files.

**Test:** Add regression cases where the target starts with an unrelated user-owned agent and where only the first bundled file exists without a marker; both retries should install every non-colliding starter agent and record only handled files.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking finding. The clean-install gate is inferred from whether the shared agents directory is empty, so a first Berd launch after any pre-existing personal agent receives none of the starter set; an interrupted first seed also cannot recover the remaining files. Static review covered every changed user flow, accessibility, localization, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, release packaging, and project/design-system rules. No additional supported issues or duplicate overlaps survived the final self-check. Supplied GitHub evidence was inspected: six checks succeeded and two checks (Frontend build smoke and Transcript virtualization) were in progress at capture time; required checks still govern merge readiness.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Fixed in 1ca24685 at the install-state boundary. Initial seeding now uses Berd-owned marker presence rather than shared directory contents. A marker-free pass handles each bundled filename independently, preserves collisions, resumes interrupted copies, and commits the marker only after the full pass. Added regressions for a pre-existing personal agent and a partial marker-free seed.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -227,6 +227,13 @@ fn seed_bundled_agents_from_dir(
         .collect::<Result<Vec<_>, _>>()?;
     entries.sort_by_key(|entry| entry.file_name());
 
+    // The agents directory is shared with Goose/ACP and may already contain
+    // personal agents before Berd's first launch. Only Berd's own marker is
+    // authoritative for whether initial seeding has completed. If a prior
+    // attempt copied some files but failed before writing the marker, another
+    // marker-free pass safely resumes the remaining non-colliding files.
+    let initial_seed =
```
morgmart: 🤖 **P1 · Recover interrupted marker writes** (blocking)

Initial seeding now treats the existence of either Berd marker file as authoritative proof that the first seed has completed. However, `write_seed_marker` still writes JSON directly to the final marker path with `fs::write`. If the process exits, storage fills, or the write otherwise fails after creating or truncating that file, the next launch sets `initial_seed` to false and then fails while parsing the partial marker. Every later launch repeats that failure, so the remaining starter agents are never installed and existing bundled agents are no longer refreshed without manual filesystem repair. The new interruption test only covers failure before a marker exists.

**User effect:** A clean installation interrupted while committing its seed marker can permanently lose some or all starter agents, including the agents expected on the starter Home, and cannot self-recover on later launches.

**Recommended fix:** Write the marker atomically using a same-directory temporary file, flush/sync it, and rename it into place. Also define safe recovery for an invalid existing marker, such as quarantining it and resuming the marker-free non-overwriting seed pass while continuing to preserve user-owned collisions.

**Test:** Add a regression that leaves a truncated or otherwise invalid marker after only part of the bundled set exists, then verifies the next seed pass recovers every missing non-colliding agent, preserves personal agents, and replaces the invalid marker with a complete valid marker.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking finding. The prior clean-install classification bug is fixed, but the replacement now treats marker-file existence as durable completion even though marker writes are non-atomic, so an interrupted write can permanently strand initial seeding. The fresh full comparison review covered every changed user flow; accessibility; i18n/localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No other supported issue survived the final self-check. Supplied GitHub evidence was inspected and all eight reported checks completed successfully; required checks still independently govern merge readiness.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Fixed in 60e5099b at the marker I/O boundary. Marker commits now use a same-directory synced temp file plus atomic replacement on Unix and Windows. Invalid Berd-owned markers are quarantined and treated as interrupted initial seeding, so missing non-colliding agents recover while personal agents remain untouched. Marker status and contents now come from one read. Added a truncated-marker recovery regression.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -451,18 +457,41 @@ fn legacy_marker_path(target_root: &Path) -> PathBuf {
...
-    if legacy_path.exists() {
-        return read_seed_marker_file(&legacy_path);
+fn read_seed_marker_for_seed(target_root: &Path) -> Result<(SeedMarker, bool), String> {
+    for path in [marker_path(target_root), legacy_marker_path(target_root)] {
+        if !path.exists() {
+            continue;
+        }
+        match read_seed_marker_file(&path) {
+            Ok(marker) => return Ok((marker, false)),
+            Err(_) => {
```
morgmart, line 701: 🤖 **P1 · Preserve deletions during recovery** (blocking)

`read_seed_marker_for_seed` treats every unreadable or unparsable marker as evidence that initial seeding never completed: it quarantines the marker and returns an empty marker with `initial_seed = true`. That is safe for an interrupted first install, but the same state can occur on an established installation whose marker is damaged, unreadable, or from an incompatible schema. If that user intentionally deleted a bundled agent, the corrupted marker was the only record distinguishing that deletion from a never-installed file; the marker-free pass now restores the missing agent. The new corrupt-marker test models only a partial first install and therefore does not exercise this ambiguous established-install case.

**User effect:** A user who deliberately removed a starter agent can have it silently reappear after marker corruption or incompatibility, overriding their persisted choice and changing their agent library without consent.

**Recommended fix:** Do not infer first-install eligibility from marker corruption alone. Fail closed for an invalid completion marker, or maintain separate durable lifecycle evidence that can distinguish an interrupted initial seed from an established install before allowing missing agents to be recreated. Keep quarantine for diagnostics without discarding intentional-deletion semantics.

**Test:** Add an established-install regression that seeds the full set, deletes one bundled agent, corrupts the marker, and restarts. Recovery must preserve the deletion and personal-agent collisions while still providing an explicit safe path for the partial-first-install case.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking finding. The previous clean-install and atomic marker-write failures are fixed, but corrupt-marker recovery now assumes every invalid marker belongs to an incomplete first seed. On an established installation this discards the record of intentionally deleted bundled agents and restores them. The fresh full-comparison review covered every changed user flow; accessibility; i18n/localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. No other supported issue survived the final self-check. The related prior thread was resolved without a substantive human reply, so the still-present recovery-policy issue is postable rather than suppressed. Supplied GitHub evidence was inspected and all eight reported checks completed successfully; required checks still independently govern merge readiness.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Fixed in 832ab1f8. Atomic marker writes and quarantine remain, but corruption now fails closed: an invalid completion marker is evidence of prior Berd state, so the conservative pass records handled filenames without recreating missing agents. Marker absence still supports interrupted pre-commit first-seed recovery, and explicit repair remains the user-authorized restoration path. Added an established-install deletion regression.

src/features/home/onboarding/starterAgents.ts
```diff
@@ -1,17 +1,16 @@
...
-export const STARTER_AGENT_NAMES = ["block.md", "Builderbot"] as const;
+// Berdy is already featured by the onboarding tour widget. These two pins
+// complete the three-agent starter Home.
+export const STARTER_AGENT_NAMES = ["Tinker", "Wildcard"] as const;
+const STARTER_AGENT_FILE_NAMES = ["tinker.md", "wildcard.md"] as const;
 const LEGACY_SEEDED_STARTER_AGENTS_STORAGE_KEY =
   "goose:home:starter-agent-pins-seeded";
 const SEEDED_STARTER_AGENTS_STORAGE_KEY =
-  "goose:home:starter-agent-pins-seeded-v2";
+  "goose:home:starter-agent-pins-seeded-v3";
```
morgmart: 🤖 **P1 · Keep Berdy until replacements exist** (blocking)

Changing the Home completion key to v3 makes an established installation with the legacy unsuffixed marker enter the Berdy-pin migration. That same installation already has a valid disk seed marker, so the new clean-install-only backend deliberately records `tinker.md` and `wildcard.md` as handled without copying them. Home nevertheless removes the existing Berdy pin before checking `selectStarterAgentPersonas`; when neither replacement persona exists, it persists the removal and completes the migration without adding Tinker or Wildcard. The current tests cover a legacy marker only at the storage-helper level and do not exercise this end-to-end mismatch between backend eligibility and Home migration.

**User effect:** An existing user can upgrade and silently lose Berdy from their customized Home while receiving neither Tinker nor Wildcard, leaving their starter-agent area emptier than before.

**Recommended fix:** Do not remove or commit migration of the Berdy pin until the complete replacement set is available and can be saved successfully. Alternatively, do not migrate established Homes in this clean-install-only release. Preserve the existing pin when replacements are unavailable.

**Test:** Add a Home migration regression with the legacy unsuffixed marker, an existing Berdy pin, and no Tinker or Wildcard personas. Verify Berdy remains pinned and the v3 completion marker is not committed; then provide both replacement personas and verify the migration completes only after their pins save successfully.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking finding. The previous clean-install, marker-write, and deletion-preservation issues are fixed. However, the new v3 Home migration can remove an existing Berdy pin even though the intentionally clean-install-only backend does not install Tinker or Wildcard for that established user, leaving no replacement starter pins. The fresh full-comparison review covered every changed user flow; accessibility; i18n/localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. Broader marker concurrency and corrupt-marker ambiguity were not reported because they either predate this comparison or are resolved by the chosen fail-closed policy. No other supported issue survived the final self-check. Supplied GitHub evidence was inspected and all eight reported checks completed successfully; required checks still independently govern merge readiness.

src/features/home/onboarding/starterAgents.ts
cynfria: 🤖 Fixed in e4081349 by removing established-Home migration from this clean-install-only PR. The existing v2 completion key remains authoritative, Home no longer removes Berdy or migrates legacy pins, and only newly eligible Homes add Tinker/Wildcard. Added an integration regression proving a legacy marker and existing Berdy pin remain untouched.

distro/agents/tinker.md
```diff
@@ -0,0 +1,52 @@
+---
+name: tinker
+display_name: Tinker
+description: Knows when you need an agent, a skill, or something else entirely, then builds it.
+avatar: app-avatar:gloopies-13
+good_for: making what you need, in Berd or out
+vibes: hands-on, resourceful
+metadata:
+  berdBundled: true
```
morgmart, line 9: 🤖 **P1 · Preserve customized bundled agents** (blocking)

Each new starter file stores `berdBundled: true` inside the same Markdown frontmatter users edit through Berd. The normal persona update path preserves existing source properties, so changing Tinker's name, prompt, avatar, or other fields leaves this marker intact. On a later package version, startup treats any marked file as still Berd-owned and replaces it whenever it differs from the packaged source. The marker proves only that the file originated in the bundle; it does not prove the current contents remain unmodified or owned by Berd. This directly contradicts the PR's stated scope that customized bundled files are left untouched.

**User effect:** A user can customize Tinker, Wildcard, or another new starter agent in Berd and silently lose those changes on a later app update.

**Recommended fix:** Track the digest or version of the last contents Berd installed in the external seed marker. Refresh only when the current file still matches that recorded digest; if it differs, treat it as user-owned and preserve it. Do not use editable frontmatter alone as proof that Berd may overwrite the current file.

**Test:** Replace the existing overwrite expectation with a regression that seeds a bundled agent, edits it while retaining all bundled metadata, changes the packaged source, and verifies startup preserves the edited file. Add a companion case proving an unchanged installed digest is refreshed to the new packaged version.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking finding. The four prior automated issues are fixed, including preservation of established Home pins. The newly bundled editable agents still embed Berd's ownership marker in their editable frontmatter, so ordinary edits retain that marker and a later package update silently replaces the user's customized file, contrary to the PR's stated collision/customization boundary. The fresh full-comparison review covered every changed user flow; accessibility; i18n/localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. The uncorrelated Home save acknowledgement and marker transaction concurrency were not reported because those mechanisms predate this comparison rather than being caused by it. No other supported issue survived the final self-check. Supplied GitHub evidence was inspected and all eight reported checks completed successfully; required checks still independently govern merge readiness.

distro/agents/tinker.md
cynfria: 🤖 Addressed in 37b1adf2 and follow-up hardening through 04646a30. Ownership now lives in a backend-managed allocation manifest with the exact installed digest; editable `berdBundled` frontmatter is no longer treated as overwrite authority. Startup preserves a file whose current digest differs from the recorded installed digest—even when bundled metadata remains—and does not create a duplicate. Home identity also comes only from backend-verified manifest allocations. Added a regression covering an edited managed agent. This clean-install-only PR deliberately does not refresh established copies during packaged updates, including unchanged ones.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -429,12 +616,18 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
                 ));
             }
         }
-        fs::rename(&temp_path, target).map_err(|err| {
+        fs::hard_link(&temp_path, target).map_err(|err| {
```
morgmart, line 637: 🤖 **P1 · Replace verified managed copies safely** (blocking)

When a packaged source digest changes and the installed target still matches the manifest's old digest, the seeder correctly decides that Berd still owns the file and calls `install_agent_file`. That helper now always creates the destination with `fs::hard_link(temp_path, target)`, which fails whenever the existing target path is present. Therefore the ordinary managed-update transition cannot succeed. The marker retains the old digest, and `verified_managed_agent_allocations` then rejects the agent because the new packaged source no longer matches that digest. Existing tests cover initial creation, unchanged copies, edits, and deletion, but not an unchanged managed copy followed by a package update.

**User effect:** The first release that updates a bundled agent's instructions or avatar can make startup seeding fail and remove Berdy's, Tinker's, or Wildcard's verified identity from Home even though the user never edited the agent.

**Recommended fix:** Separate no-replace creation from verified replacement. For an existing managed target, re-verify immediately before commit that it still matches the manifest's old digest, stage and sync the new file, atomically replace that owned path without clobbering an intervening user edit, and only then commit the new digest.

**Test:** Seed an agent, change only the packaged source, seed again, and verify the unchanged managed target updates, the manifest records the new digest, and verified allocation identity remains available. Include an intervening-edit case proving replacement fails closed and preserves the edit.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -117,39 +175,51 @@ fn repair_bundled_agent_from_dir(
...
-                        fallback_target.display()
-                    ));
-                }
+    let mut marker = read_current_seed_marker(target_root)?.unwrap_or(SeedMarker {
+        version: 1,
+        install_state: InstallState::Complete,
+        seeded_files: BTreeSet::new(),
+        allocations: BTreeMap::new(),
+    });
+    if marker.version != 1 {
```
morgmart: 🤖 **P1 · Keep corrupt manifests fail-closed** (blocking)

Invalid or unsafe manifests are quarantined and replaced with a durable version-0 marker so later startup remains fail-closed. However, explicit repair immediately converts any non-version-1 marker into a fresh version-1 manifest with no historical allocations. A Berdy click after corruption therefore reverses the fail-closed decision and promotes a partial reconstruction to authoritative ownership. Startup then sees an established version-1 manifest and deliberately does not restore allocations omitted by that one-agent repair.

**User effect:** After manifest corruption, trying to open Berdy can silently replace the durable safety state with incomplete ownership data, leaving the other starter agents permanently unmanaged and preventing reliable recovery.

**Recommended fix:** Do not reset a version-0 or otherwise unknown manifest inside feature-local repair. Route repair through the same manifest recovery state machine, and keep unknown ownership fail-closed unless an explicit recovery operation safely reconstructs the complete allocation plan without claiming or overwriting existing files.

**Test:** Create a corrupt manifest, trigger fail-closed quarantine, invoke Berdy repair, and verify the fail-closed marker remains authoritative and no partial version-1 manifest is written. Add a separate explicit recovery test if product-approved recovery is introduced.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -117,39 +175,51 @@ fn repair_bundled_agent_from_dir(
...
-    if target.file_name().and_then(|name| name.to_str()) != Some(file_name) {
-        marker.seeded_files.insert(
-            target
-                .file_name()
-                .and_then(|name| name.to_str())
-                .ok_or_else(|| "Bundled agent target is missing a valid filename".to_string())?
-                .to_string(),
-        );
+    let target = target_root.join(&target_file_name);
+    if !target.exists() || digest_file(&target)? != source_digest {
```
morgmart: 🤖 **P1 · Persist repair plans before install** (blocking)

Initial seeding persists every pending allocation before writing the first target, but explicit repair uses the opposite ordering: it chooses an allocation, writes the agent file, and only afterward inserts and commits the installed record. If the process stops between the file write and marker commit, the next repair sees an occupied but unclaimed filename and allocates another suffix. The first copy becomes an orphan that the manifest cannot distinguish from user content.

**User effect:** An interrupted Berdy repair can leave duplicate agent files and permanently lose ownership of the first restored copy, causing confusing duplicate personas and unreliable future repair or update behavior.

**Recommended fix:** Use the same resumable transition as initial seeding: atomically commit a pending repair allocation first, install to that exact allocation with collision protection, then transition the record to installed. Avoid maintaining a second non-resumable write protocol.

**Test:** Simulate interruption after the repaired target is created but before the installed transition commits, then retry. Verify the same allocation is resumed, exactly one restored file exists, and the manifest finishes with one installed record.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: three blocking findings. The previous five automated issues are fixed by the backend-managed allocation manifest and verified identity flow, but the new state machine cannot replace an existing unchanged managed target during package upgrades, explicit repair can erase the persistent corrupt-manifest fail-closed state, and repair installs before durably recording its allocation. The fresh full-comparison review covered all 30 changed files and every changed user flow; accessibility; i18n/localization; navigation and consent guards; async, error, never-completes, lifecycle, race and interruption behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. The concerns were consolidated by underlying lifecycle invariant; no other supported issue survived the final self-check. Supplied GitHub evidence was inspected: seven checks completed successfully and Tauri Rust checks (Linux) was in progress at capture time; required checks still independently govern merge readiness.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Addressed in fdcbb903. Explicit repair now requires a valid version-1 manifest and an existing managed allocation. It no longer converts corrupt, legacy, missing, or version-0 state into a partial manifest. Added `repair_keeps_corrupt_manifest_fail_closed` to verify the durable fail-closed marker remains authoritative and no agent is written.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Addressed in fdcbb903. Repair now writes the exact allocation as `pending` before creating the target, resumes or safely reallocates that recorded destination, and marks it `installed` only after the expected bytes exist. Added `repair_resumes_pending_allocation_without_duplicate` to verify retries converge on one installed record and one restored file.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Addressed in fdcbb903 with the narrower clean-install-only policy. Packaged updates no longer rewrite established managed copies. Verified identity now remains tied to the digest Berd originally installed, so an unchanged managed target stays verified when the packaged source changes, while an edited target fails verification and remains untouched. Added `packaged_source_change_preserves_managed_copy_and_verified_digest` to cover this transition.

justfile
```diff
@@ -313,6 +313,7 @@ bb-cli-docker-acceptance:
 
 # Stage the pinned Goose backend into src-tauri/binaries/goosed-<target> and build bundles.
 bundle:
+    pnpm validate:bundled-agents distro/agents/*.md
```
morgmart, line 316: 🤖 **P1 · Expand validation inputs on Windows** (blocking)

The cross-platform `bundle` recipe passes `distro/agents/*.md` directly to `pnpm`. The native Windows shell does not expand globs for external commands, so the TypeScript validator receives the literal path `distro/agents/*.md`; its `readFileSync` then reports a missing-file validation error before `_bundle-windows` can run. The successful Windows Rust check does not exercise this packaging recipe.

**User effect:** Every Windows release bundle is blocked before packaging begins, so the app cannot produce its Windows installer through the documented bundle command.

**Recommended fix:** Make the validator enumerate Markdown files when given a directory, or add a Windows-safe recipe that explicitly produces the file list rather than relying on shell glob expansion.

**Test:** Add a Windows packaging smoke test or platform-neutral validator test that invokes the same bundle-validation entry point without shell-expanded arguments and verifies all distro agent files are discovered.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -117,39 +172,75 @@ fn repair_bundled_agent_from_dir(
...
+
+    let target = target_root.join(&allocation.target_file_name);
+    let target_matches_recorded =
+        target.is_file() && digest_file(&target)? == allocation.installed_digest;
+    let target_file_name = if allocation.status == AllocationStatus::Pending
+        || !target.exists()
+        || target_matches_recorded
+    {
+        allocation.target_file_name
+    } else {
```
morgmart, line 193: 🤖 **P1 · Do not duplicate customized Berdy** (blocking)

A customized managed Berdy no longer matches the recorded digest, so verified identity is intentionally withheld and Home's Ask Berdy flow invokes repair. Repair sees the digest-changed allocated target as occupied, allocates another filename, records that new destination, and installs a stock copy. The user's edit is preserved, but a normal attempt to chat with it silently creates and selects a second Berdy instead of treating customization as relinquished ownership. The click is consent to open Berdy, not consent to add a replacement agent.

**User effect:** After changing Berdy's name, avatar, or instructions, clicking Ask Berdy can unexpectedly add a second stock Berdy and switch the interaction away from the customized agent.

**Recommended fix:** Distinguish a missing managed target from a digest-changed customized target. Repair may restore a missing allocation, but customization should remain relinquished and require an explicit restore/replace decision rather than automatic reallocation.

**Test:** Seed Berdy, customize its managed file, list personas so verified identity is absent, then trigger the Home resolution flow. Verify no second file or allocation is created and the UI presents the explicit recovery behavior chosen by product.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -201,118 +344,188 @@ fn avatar_ref_from_contents(contents: &str) -> Option<String> {
...
+    }
+    let source_digests = sources
+        .iter()
+        .map(|(name, _, digest)| (name.as_str(), digest.as_str()))
+        .collect::<BTreeMap<_, _>>();
+    if marker
+        .allocations
+        .keys()
+        .any(|name| !source_digests.contains_key(name.as_str()))
+        || marker.allocations.iter().any(|(name, allocation)| {
```
morgmart, line 449: 🤖 **P1 · Adopt pending files before drift checks** (blocking)

Seeding returns early whenever a pending allocation's recorded digest differs from the current packaged source. That check runs before inspecting the pending target. If digest-A bytes were successfully written and the process stopped before changing the record to Installed, then the app updated to package digest B, the next launch rejects the manifest without noticing that the target safely matches its durable recorded digest A. Every later launch repeats the same early return, leaving the installation permanently pending even though the risky file write completed successfully.

**User effect:** A first installation interrupted at the status-commit boundary and followed by an app update can permanently miss verified starter-agent identities and never finish its starter Home without manual filesystem repair.

**Recommended fix:** Inspect each pending target before applying package-drift policy. If it matches the allocation's recorded digest, adopt it as Installed regardless of the current package digest. Apply the product's fail-closed package-drift decision only when the pending target is absent or does not match the recorded bytes.

**Test:** Persist a pending allocation for digest A with an already-written matching target, change the packaged source to digest B, and rerun seeding. Verify the A target is adopted as Installed and remains verified without being rewritten.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -451,18 +662,51 @@ fn legacy_marker_path(target_root: &Path) -> PathBuf {
...
-        return read_seed_marker_file(&legacy_path);
+fn read_current_seed_marker(target_root: &Path) -> Result<Option<SeedMarker>, String> {
+    for path in [marker_path(target_root), legacy_marker_path(target_root)] {
+        if !path.exists() {
+            continue;
+        }
+        match read_seed_marker_file(&path) {
+            Ok(marker) if marker.version == 1 => {
+                if validate_manifest_shape(&marker).is_err() {
+                    quarantine_invalid_marker(&path)?;
```
morgmart, line 693: 🤖 **P1 · Persist fail-closed state before quarantine** (blocking)

Corrupt-manifest handling first renames the only authoritative marker to a quarantine filename and only afterward writes the version-0 fail-closed marker. A process, disk, or machine failure between those operations leaves neither recognized marker present. On the next launch, `read_current_seed_marker` returns None and the seeder treats the established installation as a clean install, reversing the exact safety decision the tombstone was meant to preserve.

**User effect:** An interruption while handling corruption can cause Berd to recreate deliberately deleted agents or allocate new bundled copies around existing files on the next launch.

**Recommended fix:** Never remove the recognized fail-closed fact before its replacement is durable. Copy and sync corrupt bytes to quarantine if diagnostics are needed, then atomically replace the authoritative marker with the version-0 tombstone; if quarantine copying fails, leave the corrupt marker present and return an error.

**Test:** Inject failure after quarantine preparation but before authoritative-marker replacement, restart, and verify the installation still fails closed and no clean-install allocation plan or agent files are created.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: four blocking findings. The previously reported managed-update, corrupt-manifest repair, and repair-plan ordering issues are fixed under the clarified clean-install-only policy. The fresh full comparison still has four concrete lifecycle and packaging failures: Windows bundle validation receives an unexpanded glob; asking Berdy after customization silently allocates a second stock copy; an app update can permanently strand an already-written pending install; and corrupt-manifest quarantine has a crash window that erases the fail-closed fact. The review covered all 30 changed files and every changed user flow; accessibility; i18n/localization; navigation and consent guards; async, error, never-completes, lifecycle, race and interruption behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. Broader concurrent-writer concerns were not included to avoid overlapping the same manifest lifecycle risks without a separately demonstrated user flow. Supplied GitHub evidence was inspected and all eight reported checks completed successfully; required checks still independently govern merge readiness.

morgmart:
🤖 This PR is superseded by #77. The replacement keeps Berd’s existing bundled-agent seeder, adds the public starter files through that one system, removes the unnecessary alternate Berdy filename, and limits Home pinning to eligible new Homes. The allocation/digest state machine is intentionally not carried forward.

justfile
cynfria: 🤖 Addressed in 34515554. The validator now accepts a directory and enumerates sorted `.md` files itself, so `bundle` passes `distro/agents` without relying on shell glob expansion. Empty directories remain validation failures. Added platform-neutral directory expansion and empty-directory regressions.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Addressed in 34515554 by narrowing repair. Repair now restores only an already-managed allocation whose target is missing. A present customized or otherwise occupied target returns an explicit-restore-required error; it is never reallocated or duplicated. Added `repair_does_not_duplicate_a_customized_agent`.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Addressed in 34515554. Startup now performs a pending-adoption pass before considering current package sources: any regular non-symlink target matching its recorded digest becomes Installed even after package drift or source removal. Missing pending targets are installed only when the current package still matches the recorded digest. Added package-drift and removed-source adoption regressions.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Addressed in 34515554 by removing quarantine/tombstone mutation entirely. Invalid, unsafe, or symlinked manifests now return an error while the authoritative marker remains untouched, so there is no rename-before-replacement crash window. Added corrupt-marker preservation and symlink-marker regressions.

PR #73, opened 2026-08-18
add a first-run landing experience

cynfria:
**Category:** new-feature
**User Impact:** New Berd installations now open with a responsive welcome experience before entering the main canvas.
**Problem:** Berd previously entered the product without a designed first-run introduction or clear usage-data disclosure. Existing installations also need to remain uninterrupted when the experience becomes the default.
**Solution:** Add an interactive, accessible landing page and graduate the old experiment into the normal first-run flow, backed by a durable installation-cohort marker that safely distinguishes new installs from upgrades.

> [!NOTE]
> The usage-data checkbox is intentionally awaiting the follow-up telemetry preference hookup and should not be treated as release-ready consent behavior yet.

<details>

**src-tauri/Cargo.toml**
Makes unique temporary-file support available to the installation cohort service.

**src-tauri/src/commands/installation.rs**
Exposes the resolved installation cohort to the renderer after startup initialization settles.

**src-tauri/src/commands/mod.rs**
Registers the installation command module.

**src-tauri/src/lib.rs**
Initializes cohort state before app-data migration and registers the command without blocking startup on classification failures.

**src-tauri/src/services/app_data_migration.rs**
Provides authoritative legacy-layout detection for upgrade classification.

**src-tauri/src/services/installation_cohort.rs**
Persists a versioned fresh-versus-established cohort marker with concurrent-launch and crash-safety handling.

**src-tauri/src/services/mod.rs**
Registers the installation cohort service.

**src/app/AppShell.tsx**
Makes incomplete onboarding the normal first-run route instead of an experimental route.

**src/features/experiments/ExperimentsSettings.tsx**
Removes obsolete hidden-experiment filtering after first-run onboarding graduation.

**src/features/experiments/__tests__/ExperimentsSettings.test.tsx**
Updates experiment settings coverage for the graduated flow.

**src/features/experiments/experimentDefinitions.ts**
Removes the retired first-run onboarding experiment definition.

**src/features/onboarding/api/installationCohort.ts**
Adds the typed renderer API for reading the native installation cohort.

**src/features/onboarding/model/onboardingStore.test.ts**
Covers fresh installs, upgrades, malformed records, unavailable storage, and persistence compatibility.

**src/features/onboarding/model/onboardingStore.ts**
Graduates established users without showing onboarding while preserving authored or newer-version state.

**src/features/onboarding/ui/OnboardingShell.tsx**
Allows the welcome layout to customize its content overflow without forking the shell.

**src/features/onboarding/ui/WelcomeStep.test.tsx**
Covers welcome actions, focus, reduced motion, project rendering, and disclosure behavior.

**src/features/onboarding/ui/WelcomeStep.tsx**
Implements the responsive landing page, interactive project cube, consent disclosure, and glass dialog.

**src/features/projects/artifact/ProjectArtifactPreview.tsx**
Passes optional camera framing through the artifact preview boundary.

**src/features/projects/artifact/ProjectArtifactRenderer.tsx**
Supports adjusted camera distance for oversized embedded project cubes.

**src/features/projects/artifact/types.ts**
Adds the typed camera-distance renderer contract.

**src/main.test.tsx**
Verifies cohort lookup and startup fallback behavior for main and session windows.

**src/main.tsx**
Resolves installation cohort before rendering the main app while preserving session-window startup.

**src/shared/i18n/locales/en/onboarding.json**
Adds English welcome and usage-data disclosure copy.

**src/shared/i18n/locales/en/settings.json**
Removes retired experiment copy.

**src/shared/i18n/locales/es/onboarding.json**
Adds Spanish welcome and usage-data disclosure copy.

**src/shared/i18n/locales/es/settings.json**
Removes retired experiment copy.

**src/shared/ui/dialog.tsx**
Uses a shared solid close icon and supports a localized accessible close label.

</details>

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -211,6 +211,13 @@ pub fn run() {
             app.manage(commands::pocket_voice::PocketVoiceState::default());
             app.manage(commands::native_voice::NativeVoiceState::default());
             app.manage(commands::voice_capture::VoiceCaptureState::default());
+            let installation_cohort =
+                services::installation_cohort::initialize_installation_cohort(
+                    &app_data_dir,
+                    services::app_data_migration::legacy_layout_database_exists(app.handle()),
+                )
+                .map_err(std::io::Error::other)?;
+            app.manage(installation_cohort);
             let release_channel_state = commands::updates::ReleaseChannelState::load(app.handle())?;
             app.manage(release_channel_state);
 
@@ -510,6 +517,7 @@ pub fn run() {
             commands::git::git_create_worktree,
             commands::git::git_remove_worktree,
             commands::home_widget_media::import_home_widget_photo,
+            commands::installation::get_installation_cohort,
             commands::layout::get_layout,
             commands::layout::save_layout_items,
             commands::layout::save_layout_camera,
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -211,13 +211,23 @@ pub fn run() {
             app.manage(commands::pocket_voice::PocketVoiceState::default());
             app.manage(commands::native_voice::NativeVoiceState::default());
             app.manage(commands::voice_capture::VoiceCaptureState::default());
+            let (installation_cohort_sender, installation_cohort_state) =
+                services::installation_cohort::installation_cohort_channel();
+            app.manage(installation_cohort_state);
             let installation_cohort =
                 services::installation_cohort::initialize_installation_cohort(
                     &app_data_dir,
                     services::app_data_migration::legacy_layout_database_exists(app.handle()),
                 )
-                .map_err(std::io::Error::other)?;
-            app.manage(installation_cohort);
+                .unwrap_or_else(|error| {
+                    log::warn!("Failed to initialize installation cohort: {error}");
+                    services::installation_cohort::InstallationCohort::Unknown
+                });
+            installation_cohort_sender.send_replace(
+                services::installation_cohort::InstallationCohortReadiness::Ready(
+                    installation_cohort,
+                ),
+            );
             let release_channel_state = commands::updates::ReleaseChannelState::load(app.handle())?;
             app.manage(release_channel_state);
 
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -214,10 +214,18 @@ pub fn run() {
             let (installation_cohort_sender, installation_cohort_state) =
                 services::installation_cohort::installation_cohort_channel();
             app.manage(installation_cohort_state);
+            let current_layout_exists =
+                services::installation_cohort::layout_database_exists(&app_data_dir);
+            let legacy_layout_exists = if app.try_state::<services::e2e_mode::E2eMode>().is_some() {
+                Ok(false)
+            } else {
+                services::app_data_migration::legacy_layout_database_exists(app.handle())
+            };
             let installation_cohort =
                 services::installation_cohort::initialize_installation_cohort(
                     &app_data_dir,
-                    services::app_data_migration::legacy_layout_database_exists(app.handle()),
+                    current_layout_exists,
+                    legacy_layout_exists,
                 )
                 .unwrap_or_else(|error| {
                     log::warn!("Failed to initialize installation cohort: {error}");
--- a/src-tauri/src/services/installation_cohort.rs
+++ b/src-tauri/src/services/installation_cohort.rs
@@ -31,6 +31,18 @@ struct InstallationCohortRecord {
     cohort: InstallationCohort,
 }
 
+pub fn layout_database_exists(app_data_dir: &Path) -> Result<bool, String> {
+    file_exists(&app_data_dir.join(CURRENT_LAYOUT_DATABASE))
+}
+
+pub(crate) fn file_exists(path: &Path) -> Result<bool, String> {
+    match fs::metadata(path) {
+        Ok(metadata) => Ok(metadata.is_file()),
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => Ok(false),
+        Err(error) => Err(format!("Failed to inspect {}: {error}", path.display())),
+    }
+}
+
 pub fn installation_cohort_channel() -> (
     watch::Sender<InstallationCohortReadiness>,
     InstallationCohortState,
@@ -41,6 +53,7 @@ pub fn installation_cohort_channel() -> (
 
 pub fn initialize_installation_cohort(
     app_data_dir: &Path,
+    current_layout_exists: Result<bool, String>,
     legacy_layout_exists: Result<bool, String>,
 ) -> Result<InstallationCohort, String> {
     fs::create_dir_all(app_data_dir)
@@ -59,8 +72,9 @@ pub fn initialize_installation_cohort(
         return Ok(record.cohort);
     }
 
+    let current_layout_exists = current_layout_exists?;
     let legacy_layout_exists = legacy_layout_exists?;
-    let cohort = if app_data_dir.join(CURRENT_LAYOUT_DATABASE).is_file() || legacy_layout_exists {
+    let cohort = if current_layout_exists || legacy_layout_exists {
         InstallationCohort::EstablishedBeforeLandingV1
     } else {
         InstallationCohort::FreshWithLandingV1
@@ -87,7 +101,11 @@ fn persist_marker(path: &Path, cohort: InstallationCohort) -> Result<Installatio
 
     match temporary.persist_noclobber(path) {
         Ok(_) => {
-            sync_parent_directory(path)?;
+            if let Err(error) = sync_parent_directory(path) {
+                log::warn!(
+                    "Installation cohort marker was published but directory sync failed: {error}"
+                );
+            }
             Ok(cohort)
         }
         Err(error) if error.error.kind() == std::io::ErrorKind::AlreadyExists => {
@@ -134,11 +152,11 @@ mod tests {
     #[test]
     fn classifies_and_persists_a_fresh_installation() {
         let root = tempdir().unwrap();
-        let first = initialize_installation_cohort(root.path(), Ok(false)).unwrap();
+        let first = initialize_installation_cohort(root.path(), Ok(false), Ok(false)).unwrap();
         assert_eq!(first, InstallationCohort::FreshWithLandingV1);
 
         fs::write(root.path().join(CURRENT_LAYOUT_DATABASE), b"later").unwrap();
-        let second = initialize_installation_cohort(root.path(), Ok(false)).unwrap();
+        let second = initialize_installation_cohort(root.path(), Ok(false), Ok(false)).unwrap();
         assert_eq!(second, InstallationCohort::FreshWithLandingV1);
     }
 
@@ -147,13 +165,13 @@ mod tests {
         let current = tempdir().unwrap();
         fs::write(current.path().join(CURRENT_LAYOUT_DATABASE), b"existing").unwrap();
         assert_eq!(
-            initialize_installation_cohort(current.path(), Ok(false)).unwrap(),
+            initialize_installation_cohort(current.path(), Ok(true), Ok(false)).unwrap(),
             InstallationCohort::EstablishedBeforeLandingV1
         );
 
         let legacy = tempdir().unwrap();
         assert_eq!(
-            initialize_installation_cohort(legacy.path(), Ok(true)).unwrap(),
+            initialize_installation_cohort(legacy.path(), Ok(false), Ok(true)).unwrap(),
             InstallationCohort::EstablishedBeforeLandingV1
         );
     }
@@ -170,7 +188,7 @@ mod tests {
             let barrier = Arc::clone(&barrier);
             std::thread::spawn(move || {
                 barrier.wait();
-                initialize_installation_cohort(&path, Ok(legacy_exists)).unwrap()
+                initialize_installation_cohort(&path, Ok(false), Ok(legacy_exists)).unwrap()
             })
         });
         barrier.wait();
@@ -179,7 +197,7 @@ mod tests {
         assert_eq!(cohorts[0], InstallationCohort::FreshWithLandingV1);
         assert_eq!(cohorts[1], cohorts[0]);
         assert_eq!(
-            initialize_installation_cohort(&path, Ok(false)).unwrap(),
+            initialize_installation_cohort(&path, Ok(false), Ok(false)).unwrap(),
             cohorts[0]
         );
     }
@@ -187,7 +205,12 @@ mod tests {
     #[test]
     fn detection_failure_does_not_publish_a_fresh_marker() {
         let root = tempdir().unwrap();
-        assert!(initialize_installation_cohort(root.path(), Err("unavailable".into())).is_err());
+        assert!(initialize_installation_cohort(
+            root.path(),
+            Err("metadata unavailable".into()),
+            Ok(false),
+        )
+        .is_err());
         assert!(!root.path().join(MARKER_FILE_NAME).exists());
     }
 
@@ -202,7 +225,7 @@ mod tests {
         .unwrap();
 
         assert_eq!(
-            initialize_installation_cohort(root.path(), Ok(false)).unwrap(),
+            initialize_installation_cohort(root.path(), Ok(false), Ok(false)).unwrap(),
             InstallationCohort::Unknown
         );
         assert!(String::from_utf8(fs::read(marker).unwrap())
--- a/src-tauri/src/services/installation_cohort.rs
+++ b/src-tauri/src/services/installation_cohort.rs
@@ -1,7 +1,7 @@
 use serde::{Deserialize, Serialize};
 use std::fs;
 use std::io::Write;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use tokio::sync::watch;
 
 const MARKER_FILE_NAME: &str = "installation-cohort-v1.json";
@@ -65,32 +65,50 @@ pub fn initialize_installation_cohort(
     } else {
         InstallationCohort::FreshWithLandingV1
     };
-    persist_marker(&marker_path, cohort)?;
-    Ok(cohort)
+    persist_marker(&marker_path, cohort)
 }
 
-fn persist_marker(path: &Path, cohort: InstallationCohort) -> Result<(), String> {
+fn persist_marker(path: &Path, cohort: InstallationCohort) -> Result<InstallationCohort, String> {
     let record = InstallationCohortRecord {
         version: MARKER_VERSION,
         cohort,
     };
+    let parent = path
+        .parent()
+        .ok_or_else(|| "Installation cohort marker has no parent directory".to_string())?;
     let bytes = serde_json::to_vec(&record)
         .map_err(|error| format!("Failed to serialize installation cohort marker: {error}"))?;
-    let part_path = part_path(path);
-    let result = (|| {
-        let mut file = fs::File::create(&part_path)
-            .map_err(|error| format!("Failed to create installation cohort marker: {error}"))?;
-        file.write_all(&bytes)
-            .and_then(|_| file.sync_all())
-            .map_err(|error| format!("Failed to write installation cohort marker: {error}"))?;
-        fs::rename(&part_path, path)
-            .map_err(|error| format!("Failed to publish installation cohort marker: {error}"))?;
-        sync_parent_directory(path)
-    })();
-    if result.is_err() {
-        let _ = fs::remove_file(&part_path);
+    let mut temporary = tempfile::NamedTempFile::new_in(parent)
+        .map_err(|error| format!("Failed to create installation cohort marker: {error}"))?;
+    temporary
+        .write_all(&bytes)
+        .and_then(|_| temporary.as_file().sync_all())
+        .map_err(|error| format!("Failed to write installation cohort marker: {error}"))?;
+
+    match temporary.persist_noclobber(path) {
+        Ok(_) => {
+            sync_parent_directory(path)?;
+            Ok(cohort)
+        }
+        Err(error) if error.error.kind() == std::io::ErrorKind::AlreadyExists => {
+            read_published_cohort(path)
+        }
+        Err(error) => Err(format!(
+            "Failed to publish installation cohort marker: {}",
+            error.error
+        )),
     }
-    result
+}
+
+fn read_published_cohort(path: &Path) -> Result<InstallationCohort, String> {
+    let bytes = fs::read(path)
+        .map_err(|error| format!("Failed to read published installation cohort marker: {error}"))?;
+    let record: InstallationCohortRecord = serde_json::from_slice(&bytes)
+        .map_err(|error| format!("Invalid published installation cohort marker: {error}"))?;
+    if record.version != MARKER_VERSION || record.cohort == InstallationCohort::Unknown {
+        return Err("Published installation cohort marker is unsupported".to_string());
+    }
+    Ok(record.cohort)
 }
 
 #[cfg(unix)]
@@ -106,10 +124,6 @@ fn sync_parent_directory(path: &Path) -> Result<(), String> {
 #[cfg(not(unix))]
 fn sync_parent_directory(_path: &Path) -> Result<(), String> {
     Ok(())
-}
-
-fn part_path(path: &Path) -> PathBuf {
-    path.with_extension("json.part")
 }
 
 #[cfg(test)]
@@ -141,6 +155,32 @@ mod tests {
         assert_eq!(
             initialize_installation_cohort(legacy.path(), Ok(true)).unwrap(),
             InstallationCohort::EstablishedBeforeLandingV1
+        );
+    }
+
+    #[test]
+    fn concurrent_initializers_use_the_published_winner() {
+        use std::sync::{Arc, Barrier};
+
+        let root = tempdir().unwrap();
+        let path = Arc::new(root.path().to_path_buf());
+        let barrier = Arc::new(Barrier::new(3));
+        let handles = [false, false].map(|legacy_exists| {
+            let path = Arc::clone(&path);
+            let barrier = Arc::clone(&barrier);
+            std::thread::spawn(move || {
+                barrier.wait();
+                initialize_installation_cohort(&path, Ok(legacy_exists)).unwrap()
+            })
+        });
+        barrier.wait();
+
+        let cohorts = handles.map(|handle| handle.join().unwrap());
+        assert_eq!(cohorts[0], InstallationCohort::FreshWithLandingV1);
+        assert_eq!(cohorts[1], cohorts[0]);
+        assert_eq!(
+            initialize_installation_cohort(&path, Ok(false)).unwrap(),
+            cohorts[0]
         );
     }
 
--- a/src-tauri/src/services/installation_cohort.rs
+++ b/src-tauri/src/services/installation_cohort.rs
@@ -1,0 +1,136 @@
+use serde::{Deserialize, Serialize};
+use std::fs;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+const MARKER_FILE_NAME: &str = "installation-cohort-v1.json";
+const MARKER_VERSION: u32 = 1;
+const CURRENT_LAYOUT_DATABASE: &str = "berd.sqlite";
+
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum InstallationCohort {
+    FreshWithLandingV1,
+    EstablishedBeforeLandingV1,
+}
+
+#[derive(Clone, Debug)]
+pub struct InstallationCohortState(pub InstallationCohort);
+
+#[derive(Deserialize, Serialize)]
+struct InstallationCohortRecord {
+    version: u32,
+    cohort: InstallationCohort,
+}
+
+pub fn initialize_installation_cohort(
+    app_data_dir: &Path,
+    legacy_layout_exists: bool,
+) -> Result<InstallationCohortState, String> {
+    fs::create_dir_all(app_data_dir)
+        .map_err(|error| format!("Failed to create app data directory: {error}"))?;
+    let marker_path = app_data_dir.join(MARKER_FILE_NAME);
+
+    if marker_path.exists() {
+        let bytes = fs::read(&marker_path)
+            .map_err(|error| format!("Failed to read installation cohort marker: {error}"))?;
+        let Ok(record) = serde_json::from_slice::<InstallationCohortRecord>(&bytes) else {
+            return Ok(InstallationCohortState(
+                InstallationCohort::EstablishedBeforeLandingV1,
+            ));
+        };
+        if record.version != MARKER_VERSION {
+            return Ok(InstallationCohortState(
+                InstallationCohort::EstablishedBeforeLandingV1,
+            ));
+        }
+        return Ok(InstallationCohortState(record.cohort));
+    }
+
+    let cohort = if app_data_dir.join(CURRENT_LAYOUT_DATABASE).is_file() || legacy_layout_exists {
+        InstallationCohort::EstablishedBeforeLandingV1
+    } else {
+        InstallationCohort::FreshWithLandingV1
+    };
+    persist_marker(&marker_path, cohort)?;
+    Ok(InstallationCohortState(cohort))
+}
+
+fn persist_marker(path: &Path, cohort: InstallationCohort) -> Result<(), String> {
+    let record = InstallationCohortRecord {
+        version: MARKER_VERSION,
+        cohort,
+    };
+    let bytes = serde_json::to_vec(&record)
+        .map_err(|error| format!("Failed to serialize installation cohort marker: {error}"))?;
+    let part_path = part_path(path);
+    let mut file = fs::File::create(&part_path)
+        .map_err(|error| format!("Failed to create installation cohort marker: {error}"))?;
+    file.write_all(&bytes)
+        .and_then(|_| file.sync_all())
+        .map_err(|error| format!("Failed to write installation cohort marker: {error}"))?;
+    fs::rename(&part_path, path)
+        .map_err(|error| format!("Failed to publish installation cohort marker: {error}"))
+}
+
+fn part_path(path: &Path) -> PathBuf {
+    path.with_extension("json.part")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tempfile::tempdir;
+
+    #[test]
+    fn classifies_and_persists_a_fresh_installation() {
+        let root = tempdir().unwrap();
+        let first = initialize_installation_cohort(root.path(), false).unwrap();
+        assert_eq!(first.0, InstallationCohort::FreshWithLandingV1);
+
+        fs::write(root.path().join(CURRENT_LAYOUT_DATABASE), b"later").unwrap();
+        let second = initialize_installation_cohort(root.path(), false).unwrap();
+        assert_eq!(second.0, InstallationCohort::FreshWithLandingV1);
+    }
+
+    #[test]
+    fn classifies_current_or_legacy_layouts_as_established() {
+        let current = tempdir().unwrap();
+        fs::write(current.path().join(CURRENT_LAYOUT_DATABASE), b"existing").unwrap();
+        assert_eq!(
+            initialize_installation_cohort(current.path(), false)
+                .unwrap()
+                .0,
+            InstallationCohort::EstablishedBeforeLandingV1
+        );
+
+        let legacy = tempdir().unwrap();
+        assert_eq!(
+            initialize_installation_cohort(legacy.path(), true)
+                .unwrap()
+                .0,
+            InstallationCohort::EstablishedBeforeLandingV1
+        );
+    }
+
+    #[test]
+    fn treats_an_unsupported_marker_as_established_and_preserves_it() {
+        let root = tempdir().unwrap();
+        let marker = root.path().join(MARKER_FILE_NAME);
+        fs::write(
+            &marker,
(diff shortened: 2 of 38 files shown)


src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +23,203 @@ const reveal = {
...
-            </Button>
-          </motion.div>
-        </div>
+            {t("welcome.getStarted")}
+          </Button>
+
+          <div className="mt-6 flex w-[calc(100%-72px)] items-start gap-2.5 text-left text-xs leading-[1.15] text-muted-foreground max-[760px]:mx-auto">
+            <Checkbox
+              id={checkboxId}
+              checked={shareUsageData}
```
morgmart, line 186: 🤖 **P1 · Consent choice does not control telemetry** (blocking)

The new consent checkbox only updates component-local state, while main startup calls initTelemetry() and trackAppLaunched() before the welcome UI is rendered. Advancing onboarding also does not read or persist the checkbox value. The disclosure therefore presents a privacy choice that the implementation does not honor; the PR description also explicitly says this hookup is not release-ready.

**User effect:** A fresh user can have launch telemetry initialized before seeing the disclosure, and unchecking “Share anonymous usage data” changes nothing even though the UI implies that it opts them out.

**Recommended fix:** Introduce one authoritative persisted telemetry-consent preference and gate telemetry initialization and event emission on it, including a deliberate policy for established users. If that cannot ship in this PR, remove the checkbox and consent claim until the preference is wired.

**Test:** Add an integration test with telemetry spies that starts from an unset preference, proves no telemetry initializes before consent, then proves unchecked remains disabled across restart and checked enables telemetry only after confirmation.

src/main.tsx
```diff
@@ -161,30 +163,41 @@ if (bootError) {
       renderBootError("The session window bundle could not be loaded.");
     });
 } else {
-  initTelemetry();
-  trackAppLaunched();
+  getInstallationCohort()
```
morgmart, line 174: 🤖 **P1 · Cohort lookup can blank startup forever** (blocking)

The renderer does not mount any loading or application UI until getInstallationCohort() settles. Rejection is handled, but a command that never resolves prevents the finally block from running, with no timeout or visible degraded path.

**User effect:** If the bookkeeping IPC stalls, the user sees a permanently blank main window and has no retry or recovery action.

**Recommended fix:** Render an existing startup surface immediately and bound cohort resolution with a fallback to unknown, or move cohort readiness into the app's existing startup lifecycle so failure and retry are visible without making first paint depend on an unbounded promise.

**Test:** Use a deferred invoke promise and fake timers to prove a pending lookup first shows startup UI and then reaches the intended unknown/retry path after the bound instead of remaining blank.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +23,203 @@ const reveal = {
...
+            hoverX: 0,
+            hoverY: 0,
+          }
+        : motion,
+    );
+  };
+
   return (
-    <OnboardingShell>
+    <OnboardingShell contentClassName="max-[760px]:overflow-visible">
```
morgmart: 🤖 **P1 · Narrow welcome actions can be unreachable** (blocking)

At widths up to 760px the welcome screen overrides its content region from overflow-y-auto to overflow-visible, while OnboardingShell still clips the full-screen main surface with overflow-hidden. The CTA section is absolutely positioned and contributes no scroll height. In a short window, with text scaling, or with longer localized copy, controls extending outside the viewport cannot be scrolled into view.

**User effect:** Users on a small or zoomed window can lose access to “Let’s go,” the consent choice, or its disclosure and become stuck on first launch.

**Recommended fix:** Keep a real vertical scroll container at narrow sizes and place the actionable content in normal flow; position only the decorative artifact independently. Verify intermediate widths and short-height windows as well as the wide composition.

**Test:** Add a responsive browser/component test at a narrow, short viewport with enlarged text and assert the CTA, checkbox, and Learn more control can all be reached by scrolling and activated.

src-tauri/src/services/installation_cohort.rs
```diff
@@ -0,0 +1,212 @@
...
+            return Ok(InstallationCohort::Unknown);
+        };
+        if record.version != MARKER_VERSION || record.cohort == InstallationCohort::Unknown {
+            return Ok(InstallationCohort::Unknown);
+        }
+        return Ok(record.cohort);
+    }
+
+    let legacy_layout_exists = legacy_layout_exists?;
+    let cohort = if app_data_dir.join(CURRENT_LAYOUT_DATABASE).is_file() || legacy_layout_exists {
```
morgmart: 🤖 **P1 · Inspection errors persist the wrong cohort** (blocking)

Both current and legacy database detection use Path::is_file(), which collapses metadata failures into false. A permissions or transient filesystem error can therefore be interpreted as “no existing installation,” after which the code durably writes the fresh-install marker. The added Result error path and its test do not protect real metadata failures because those failures have already been swallowed.

**User effect:** An established user can be permanently misclassified as new and unexpectedly forced through onboarding after a transient filesystem problem.

**Recommended fix:** Use fallible metadata inspection for every candidate path, treat only NotFound as absence, and propagate all other errors so no cohort marker is published until classification is trustworthy.

**Test:** Add a discriminating filesystem abstraction test where metadata returns a non-NotFound error and assert initialization returns an error and writes no marker; retain separate absent-path and existing-file cases.

src-tauri/src/lib.rs
```diff
@@ -211,6 +211,23 @@ pub fn run() {
             app.manage(commands::pocket_voice::PocketVoiceState::default());
             app.manage(commands::native_voice::NativeVoiceState::default());
             app.manage(commands::voice_capture::VoiceCaptureState::default());
+            let (installation_cohort_sender, installation_cohort_state) =
+                services::installation_cohort::installation_cohort_channel();
+            app.manage(installation_cohort_state);
+            let installation_cohort =
+                services::installation_cohort::initialize_installation_cohort(
+                    &app_data_dir,
+                    services::app_data_migration::legacy_layout_database_exists(app.handle()),
```
morgmart: 🤖 **P1 · E2E launches always get unknown cohort** (blocking)

Cohort initialization unconditionally calls legacy_layout_database_exists(), whose path resolver supports only production and development identifiers. Isolated app-driver runs use xyz.block.berd.e2e.<run-id>, so legacy detection errors and startup publishes Unknown. Since onboarding is now unconditional for an incomplete lifecycle, clean E2E launches no longer receive the expected fresh-install classification and are diverted into onboarding.

**User effect:** App-driver startup and navigation scenarios can open the onboarding flow instead of the requested product surface, breaking automation and preventing E2E from exercising production-equivalent first-run behavior.

**Recommended fix:** Handle active E2E mode explicitly with its isolated app-data directory and no legacy predecessor, or make unsupported identifiers safely mean no legacy layout only when the mode contract proves that assumption.

**Test:** Add a startup integration test using an xyz.block.berd.e2e.<run-id> identifier and isolated empty directory, then assert the command returns fresh-with-landing-v1 and the intended E2E navigation remains reachable.

src-tauri/src/services/installation_cohort.rs
```diff
@@ -0,0 +1,212 @@
...
+    let mut temporary = tempfile::NamedTempFile::new_in(parent)
+        .map_err(|error| format!("Failed to create installation cohort marker: {error}"))?;
+    temporary
+        .write_all(&bytes)
+        .and_then(|_| temporary.as_file().sync_all())
+        .map_err(|error| format!("Failed to write installation cohort marker: {error}"))?;
+
+    match temporary.persist_noclobber(path) {
+        Ok(_) => {
+            sync_parent_directory(path)?;
```
morgmart: 🤖 **P2 · Published cohort can be reported unknown** (blocking)

After persist_noclobber successfully publishes the cohort marker, a failure from sync_parent_directory is propagated. Startup converts that error to Unknown even though the durable marker already contains a valid cohort, so persisted truth and renderer-visible truth disagree for that launch.

**User effect:** A transient directory-sync failure can show onboarding to an established user; if they make progress, that authored onboarding state is then preserved on later launches even though the marker says they should have been graduated.

**Recommended fix:** Once publication succeeds, return the published cohort and log directory-sync failure as a durability warning, or otherwise structure publication so the persisted record and returned classification cannot disagree.

**Test:** Inject a parent-directory sync failure after successful marker publication and assert initialization still returns the published cohort, the marker remains readable, and subsequent initialization returns the same value.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Static review of the exact three-dot comparison found six blocking issues and no non-blocking issues. The usage-data control does not govern telemetry; cohort lookup can leave startup blank forever; the narrow welcome layout can clip its only actions without scrolling; fallible filesystem classification can permanently mark an established install as fresh; isolated E2E identifiers always fall back to an unknown cohort; and a post-publication directory-sync error can make the returned cohort disagree with the durable marker. Supplied GitHub evidence was inspected: combined status was pending and no check runs were reported, so required CI still governs merge readiness. No local code, tests, builds, package managers, hooks, scripts, or generated binaries were run.

Deterministic publication result: 6 blocking and 0 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +28,210 @@ const reveal = {
...
-  const projectThumbnail = selectCollectionImageUrl(
-    artifacts,
-    "onboarding",
-    "project-cube",
+  const { t } = useTranslation(["onboarding", "common"]);
+  const reduceMotion = useReducedMotion() === true;
+  const checkboxId = useId();
+  const headingRef = useRef<HTMLHeadingElement>(null);
+  const [shareUsageData, setShareUsageData] = useState(
+    () => getTelemetryConsent() ?? true,
```
morgmart, line 41: 🤖 **P1 · Progression is treated as telemetry consent** (blocking)

When no telemetry preference exists, the checkbox defaults to checked. Clicking the general “Let’s go” onboarding action persists that default and immediately starts telemetry, even if the user never interacted with or affirmatively selected the sharing control. The new telemetry owner now honors the stored value, but the UI still infers consent from proceeding through onboarding rather than receiving an explicit positive choice.

**User effect:** A user who simply continues setup begins sharing usage data without distinctly opting in, despite the screen presenting this as a privacy choice.

**Recommended fix:** Fail closed when the preference is unset: initialize the control unchecked, or require a separate explicit affirmative sharing choice before storing true. Keep onboarding progression independent from telemetry consent.

**Test:** Start with no stored preference, click “Let’s go” without touching the checkbox, and assert consent is not true and telemetry does not start; then explicitly opt in and assert the preference and one-time startup behavior.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +28,210 @@ const reveal = {
...
+              gestureFreezeActive={reduceMotion}
+              className="h-full w-full"
+            />
+          </div>
+        </motion.div>
+
+        <motion.section
+          variants={reveal}
+          transition={{ duration: reduceMotion ? 0 : 0.35 }}
+          className="w-full max-w-[390px] justify-self-start max-[760px]:absolute max-[760px]:top-[42%] max-[760px]:left-1/2 max-[760px]:-translate-x-1/2 max-[760px]:-translate-y-1/2 max-[760px]:text-center"
```
morgmart, line 152: 🤖 **P1 · Narrow content remains outside scroll flow** (blocking)

The latest change restores vertical overflow on the narrow container, but at widths up to 760px the actionable section is still absolutely positioned at top 42% with a negative half-height translation. It contributes no scroll height, and content extending above the scroll origin cannot be reached by scrolling. Short windows, increased text size, zoom, or longer localized copy can therefore still clip the heading or controls.

**User effect:** At the supported narrow window size or with enlarged text, users can lose access to the welcome heading, primary action, consent control, or disclosure and become stuck during first run.

**Recommended fix:** Keep the actionable narrow layout in normal document flow with sufficient top and bottom spacing; position only the decorative artifact independently. Validate the minimum window, 200% text/zoom, and Spanish copy.

**Test:** Add a constrained responsive test at the minimum narrow and short viewport with enlarged text and Spanish localization, and assert the heading, CTA, checkbox, and Learn more control are all reachable by scrolling and operable.

src-tauri/src/lib.rs
```diff
@@ -211,6 +211,31 @@ pub fn run() {
...
+                services::installation_cohort::installation_cohort_channel();
+            app.manage(installation_cohort_state);
+            let current_layout_exists =
+                services::installation_cohort::layout_database_exists(&app_data_dir);
+            let legacy_layout_exists = if app.try_state::<services::e2e_mode::E2eMode>().is_some() {
+                Ok(false)
+            } else {
+                services::app_data_migration::legacy_layout_database_exists(app.handle())
+            };
+            let installation_cohort =
```
morgmart, line 224: 🤖 **P2 · Cohort classification races migration** (blocking)

Startup inspects current and legacy databases and permanently publishes the cohort marker before running legacy migration. Atomic no-clobber publication coordinates marker writers, but no inter-process lock covers the database observations and migration. A concurrently running Berd or legacy Goose Internal process can publish established data after inspection but before migration, allowing this process to persist fresh-with-landing-v1 and then copy/open established data.

**User effect:** An installation with established data can be permanently treated as new and unexpectedly routed through onboarding because of launch timing between processes.

**Recommended fix:** Serialize classification and legacy migration with an inter-process startup/migration lock. Under the lock, inspect pre-migration evidence, perform migration, and publish the marker only after the evidence and resulting layout are stable.

**Test:** Add a lock-boundary or multi-process test that pauses after an absent observation, creates a valid legacy layout from another writer, resumes startup, and asserts the final marker is established rather than fresh.

src-tauri/src/services/installation_cohort.rs
```diff
@@ -0,0 +1,235 @@
...
+    cohort: InstallationCohort,
+}
+
+pub fn layout_database_exists(app_data_dir: &Path) -> Result<bool, String> {
+    file_exists(&app_data_dir.join(CURRENT_LAYOUT_DATABASE))
+}
+
+pub(crate) fn file_exists(path: &Path) -> Result<bool, String> {
+    match fs::metadata(path) {
+        Ok(metadata) => Ok(metadata.is_file()),
```
morgmart, line 40: 🤖 **P2 · Ambiguous database paths become fresh** (blocking)

file_exists returns false for every existing path whose metadata is not a regular file. A directory, FIFO, or other unexpected entry at either database path is therefore treated exactly like NotFound, and startup may permanently publish the fresh cohort even though layout initialization or migration cannot safely use that path. This bypasses the otherwise fail-closed metadata-error behavior.

**User effect:** After a filesystem conflict or malformed installation path, Berd can save the wrong permanent cohort and later force an established user through onboarding once the path problem is repaired.

**Recommended fix:** Return an error for an existing non-file path and explicitly define symlink handling, so ambiguous evidence resolves to unknown without publishing a marker.

**Test:** Create a directory at the current and legacy database paths and assert detection errors and no cohort marker is written; cover the intended symlink policy separately.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +28,210 @@ const reveal = {
...
+            </p>
+          </div>
+        </motion.section>
       </motion.div>
+
+      <Dialog open={detailsOpen} onOpenChange={setDetailsOpen}>
+        <DialogContent
+          size="xl"
+          closeLabel={t("actions.close", { ns: "common" })}
+          className="gap-5 rounded-[28px] p-10 pr-12"
```
morgmart, line 209: 🤖 **P2 · Welcome dialog bypasses shared chrome** (non-blocking)

The disclosure applies feature-local radius, padding, scrim color, and backdrop filters directly to DialogContent. Project guidance assigns modal surface geometry and visual behavior to the shared Dialog API, so this creates a second recipe that can drift across themes and future design-system changes.

**User effect:** The first-run disclosure can look and behave inconsistently with other Berd dialogs as the shared system evolves.

**Recommended fix:** Use the existing Dialog surface defaults, or add a named shared semantic surface/overlay option with documented theme mappings and explorer coverage.

**Test:** Cover the semantic option in the design-system explorer or component tests and assert WelcomeStep selects it without feature-owned surface color, radius, or backdrop classes.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +28,210 @@ const reveal = {
...
-        </div>
+            {t("welcome.getStarted")}
+          </Button>
+
+          <div className="mt-6 flex w-[calc(100%-72px)] items-start gap-2.5 text-left text-xs leading-[1.15] text-muted-foreground max-[760px]:mx-auto">
+            <Checkbox
+              id={checkboxId}
+              checked={shareUsageData}
+              onCheckedChange={(checked) => setShareUsageData(checked === true)}
+              aria-describedby={`${checkboxId}-description`}
```
morgmart, line 188: 🤖 **P2 · Checkbox description repeats controls** (non-blocking)

The checkbox gets its accessible name from the associated label, then aria-describedby points to the enclosing paragraph containing that same label and the separate Learn more button. Assistive technology can announce the consent sentence twice and include the neighboring button label as checkbox description.

**User effect:** Screen-reader users may hear a confusing duplicated consent announcement that blends the checkbox with a separate action.

**Recommended fix:** Remove aria-describedby unless there is distinct supporting copy. Keep the checkbox associated only with its label and leave Learn more as an independent adjacent control.

**Test:** Add an accessibility assertion for the checkbox's computed name and description, verifying the sentence is announced once and Learn more is exposed only as its own button.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found four blocking and two non-blocking findings. The previously reported telemetry plumbing and renderer-side never-completes path are substantially improved, but the final flow still treats a preselected checkbox plus the general onboarding CTA as affirmative telemetry consent, and the narrow layout still absolutely positions the actionable section so overflow scrolling cannot reliably recover content above the scroll origin. Cohort classification can also race legacy migration across processes, and an existing non-file database path is permanently classified as fresh rather than unknown. Two additional design-system/accessibility issues remain in the welcome disclosure. Supplied GitHub evidence was inspected: Rust and clean-install checks passed, while Frontend checks and unit tests failed and some jobs were skipped; required CI still governs merge readiness. No local code, tests, builds, package managers, hooks, scripts, or generated binaries were run.

Deterministic publication result: 4 blocking and 2 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/shared/telemetry/startup.ts
```diff
@@ -0,0 +1,16 @@
+import { initTelemetry, trackAppLaunched } from "@/shared/telemetry/client";
+import { getTelemetryConsent } from "@/shared/telemetry/consentPreference";
+
+let started = false;
+
+export function startTelemetryIfConsented(): boolean {
+  if (started || getTelemetryConsent() !== true) return false;
+  started = true;
+  initTelemetry();
```
morgmart: 🤖 **P1 · Telemetry failures block core flows** (blocking)

startTelemetryIfConsented calls the distribution-provided telemetry implementation without containing exceptions. On the welcome screen, an exception prevents onStart from running after consent has already been persisted. On subsequent launches, the same helper runs inside the cohort promise's finally callback before the main React tree is rendered, so an exception leaves the application on StartupLoadingView. Because started is set before initialization, the current renderer lifecycle cannot retry either.

**User effect:** A telemetry SDK failure can make the “Let’s go” action stop working or prevent Berd from opening on every later launch for a user who opted in.

**Recommended fix:** Make telemetry strictly best-effort inside the shared startup owner: catch and report initialization and launch-event failures, preserve onboarding and rendering in a finally-safe path, and define retry/idempotence behavior without allowing telemetry to own application availability.

**Test:** Mock initTelemetry and trackAppLaunched to throw separately and assert the welcome action still calls onStart, main startup still renders App, the error is reported, and repeated attempts follow the intended retry/idempotence policy.

src/main.tsx
```diff
@@ -161,30 +164,47 @@ if (bootError) {
...
     </React.StrictMode>,
   );
+  getInstallationCohort()
+    .then((cohort) => {
+      initializeOnboardingGraduation(cohort);
+    })
+    .catch((error) => {
+      console.error("Failed to resolve installation cohort:", error);
+      reportRendererError("installation_cohort_failed", error);
+      initializeOnboardingGraduation("unknown");
```
morgmart, line 181: 🤖 **P2 · Unknown cohort forces first-run onboarding** (blocking)

Metadata errors, unreadable or unsupported markers, backend initialization failures, and the renderer timeout all resolve to the unknown cohort. initializeOnboardingGraduation deliberately leaves onboarding pending for unknown, while AppShell now shows OnboardingFlow for every lifecycle other than completed. The implementation therefore converts “classification unavailable” into “definitely fresh,” contrary to the PR's stated requirement that established installations remain uninterrupted.

**User effect:** A transient slow disk, permissions issue, timeout, or damaged marker can force an established user through first-run provider and agent setup, potentially on every launch until they complete it.

**Recommended fix:** Keep unknown distinct from fresh. Use an explicit retryable recovery state or conservatively bypass onboarding for unknown so first-run UI is shown only for an affirmative fresh-with-landing-v1 classification.

**Test:** Exercise timeout, metadata failure, malformed marker, and unsupported marker through the real main-to-store-to-AppShell boundary and assert the chosen unknown recovery behavior does not silently render fresh-install onboarding for an established profile.

src/main.tsx
```diff
@@ -161,30 +164,47 @@ if (bootError) {
...
-              <ThemeProvider>
-                <UpdaterProvider>
-                  <App />
-                </UpdaterProvider>
-              </ThemeProvider>
-            </I18nProvider>
-          </QueryClientProvider>
-        </RendererErrorBoundary>
-      </TooltipProvider>
+      <I18nProvider>
```
morgmart, line 169: 🤖 **P2 · Startup loader stays hidden** (non-blocking)

The bootstrap path mounts StartupLoadingView before cohort resolution, but the main Tauri window is configured initially invisible. Window.show remains in App's effect, and App is not mounted until the cohort request settles or times out. The new loading surface therefore does not provide visible feedback during the wait it was added to cover.

**User effect:** During a slow or stalled cohort lookup, users see no Berd window for up to five seconds and then see the app appear abruptly instead of seeing startup progress.

**Recommended fix:** Move initial window visibility into the bootstrap owner and show the main window after the startup loading surface mounts, while retaining App's later startup diagnostics and normal lifecycle behavior.

**Test:** Hold the cohort promise pending and spy on the Tauri window API, asserting the loading surface is mounted and the main window is shown before cohort resolution or timeout.

src/shared/ui/dialog.tsx
```diff
@@ -1,7 +1,18 @@
 import type * as React from "react";
 import * as DialogPrimitive from "@radix-ui/react-dialog";
 import { Slot } from "@radix-ui/react-slot";
-import { XIcon } from "lucide-react";
+function SolidXIcon({ className }: { className?: string }) {
```
morgmart, line 4: 🤖 **P3 · Dialog forks the close icon system** (non-blocking)

The shared Dialog primitive replaces the established icon-library close affordance with a new inline filled SVG defined inside dialog.tsx. This globally changes every dialog and creates an undocumented one-off icon recipe rather than using or extending the app's shared icon system.

**User effect:** Every dialog receives a visually different close affordance that can drift in weight, sizing, and future accessibility treatment from the rest of Berd's icons.

**Recommended fix:** Keep the established shared icon-family component, or introduce a reusable system-owned close icon and document the intentional global change instead of embedding a private SVG in Dialog.

**Test:** Add shared Dialog visual/explorer coverage for the close affordance and assert it uses the approved icon primitive at all supported dialog sizes and themes.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found two new blocking and two new non-blocking findings. Telemetry implementation exceptions can prevent onboarding progression or leave later launches stuck on startup, and cohort uncertainty still silently routes established users into first-run onboarding. The new startup loader is mounted while the native window remains hidden, and the shared dialog now owns a one-off inline close icon outside the established icon system. Six still-present findings were suppressed because their prior automated threads remain unresolved. The final self-check covered the first-run and established-install flows, telemetry opt-in/out and feedback tracking, startup success/error/timeout/never-completes behavior, migration and concurrent-writer races, responsive/reflow behavior, accessibility, i18n, navigation gates, test honesty, design-system rules, and duplicate overlap. Supplied GitHub evidence was inspected: DCO, clean-room install, and completed Rust checks passed while frontend and Windows checks were in progress; required CI still governs merge readiness. No local code, tests, builds, package managers, hooks, scripts, or generated binaries were run.

Deterministic publication result: 2 blocking and 2 non-blocking finding(s) publishable; 6 duplicate(s) suppressed.

matt2e (dismissed):

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +28,215 @@ const reveal = {
...
-  const projectThumbnail = selectCollectionImageUrl(
-    artifacts,
-    "onboarding",
-    "project-cube",
+  const { t } = useTranslation(["onboarding", "common"]);
+  const reduceMotion = useReducedMotion() === true;
+  const checkboxId = useId();
+  const headingRef = useRef<HTMLHeadingElement>(null);
+  const [shareUsageData, setShareUsageData] = useState(
+    () => getTelemetryConsent() ?? true,
```
morgmart, line 41: 🤖 **P1 · Progression still implies telemetry consent** (blocking)

When no preference exists, the sharing checkbox initializes checked. Clicking the general “Let’s go” onboarding action persists that untouched default and starts telemetry. The system now respects the stored value, but it still derives affirmative consent from proceeding through onboarding rather than from a distinct positive sharing choice. The earlier automated thread was resolved without a substantive human reply, and the issue remains in the current comparison, so it is eligible to be reopened.

**User effect:** A user who simply continues setup begins sharing usage data without distinctly opting in, despite the screen presenting sharing as a choice.

**Recommended fix:** Fail closed when consent is unset: initialize the control unchecked, or require a separate explicit affirmative sharing action. Keep the onboarding progression action independent from telemetry consent.

**Test:** Start with no stored preference, click “Let’s go” without interacting with sharing, and assert telemetry remains disabled; then explicitly opt in and assert the durable preference and one-time startup behavior.

src/features/onboarding/ui/WelcomeStep.tsx
```diff
@@ -15,79 +28,215 @@ const reveal = {
...
+
+          <Button
+            type="button"
+            size="lg"
+            className="mt-9 w-[calc(100%-80px)] max-[760px]:mx-auto"
+            onClick={() => {
+              // Consent is persisted before telemetry starts, so onboarding has
+              // to advance regardless of how the telemetry seam behaves.
+              try {
+                if (setTelemetryConsent(shareUsageData) && shareUsageData) {
```
morgmart, line 172: 🤖 **P1 · Consent cannot be changed later** (blocking)

WelcomeStep is the only production surface that writes the new app-wide telemetry preference. Startup and feedback treat it as authoritative, but there is no persistent settings surface or shared lifecycle for changing it after onboarding, and the process-level started latch has no disable path. Fresh users who opt in cannot revoke consent; established users are graduated past the only choice surface and cannot opt in. Onboarding has therefore become the accidental owner of a system-wide privacy lifecycle.

**User effect:** Users cannot later withdraw data-sharing consent, and existing users who skipped onboarding cannot express a preference at all.

**Recommended fix:** Give telemetry an app-wide consent owner that reads and updates the preference, starts collection when enabled, stops or disables it when revoked, and notifies consumers. Use that owner from both onboarding and a persistent Privacy/Telemetry setting available to every user.

**Test:** Cover the full lifecycle: established user opts in from settings, fresh user revokes after onboarding, telemetry stops or disables according to the SDK contract, feedback tracking follows the live preference, and the choice persists across restart.

src-tauri/src/services/installation_cohort.rs
```diff
@@ -0,0 +1,235 @@
...
+        let Ok(record) = serde_json::from_slice::<InstallationCohortRecord>(&bytes) else {
+            return Ok(InstallationCohort::Unknown);
+        };
+        if record.version != MARKER_VERSION || record.cohort == InstallationCohort::Unknown {
+            return Ok(InstallationCohort::Unknown);
+        }
+        return Ok(record.cohort);
+    }
+
+    let current_layout_exists = current_layout_exists?;
```
morgmart, line 75: 🤖 **P2 · Definitive established evidence is discarded** (blocking)

The current and legacy database probe Results are unwrapped sequentially. If either probe is Ok(true) but the other probe errors, initialization returns Unknown instead of using the already definitive evidence that an established database exists. For example, a valid current berd.sqlite is ignored when permissions prevent inspecting the obsolete legacy directory.

**User effect:** An established user can be sent into first-run onboarding because an irrelevant secondary filesystem probe failed, even though Berd already found their existing database.

**Recommended fix:** Aggregate the probes as three-state evidence: classify established if either is Ok(true), classify fresh only when both are Ok(false), and return unknown/error only when no positive evidence exists and at least one probe is uncertain.

**Test:** Add cases for Ok(true) plus Err and Err plus Ok(true), asserting both classify and persist established; retain both-false and uncertainty-without-positive-evidence cases.

src/main.tsx
```diff
@@ -161,30 +164,48 @@ if (bootError) {
...
     </React.StrictMode>,
   );
+  getInstallationCohort()
+    .then((cohort) => {
+      initializeOnboardingGraduation(cohort);
+    })
+    .catch((error) => {
+      console.error("Failed to resolve installation cohort:", error);
+      reportRendererError("installation_cohort_failed", error);
+      initializeOnboardingGraduation("unknown");
```
morgmart, line 181: 🤖 **P2 · Unknown cohort still forces onboarding** (blocking)

Metadata errors, unreadable or unsupported markers, backend initialization failures, and renderer timeout all resolve to Unknown. Graduation leaves onboarding pending for Unknown, while AppShell shows onboarding for every lifecycle except completed. This converts “classification unavailable” into “definitely fresh.” The prior automated thread was resolved without a substantive human reply, and the behavior remains, so it is eligible to be reopened.

**User effect:** A transient disk, permissions, marker, or timeout problem can force an established user through first-run provider and agent setup, potentially on every launch.

**Recommended fix:** Keep unknown distinct from fresh. Use a retryable recovery state or conservatively bypass onboarding for unknown so first-run UI appears only after affirmative fresh classification.

**Test:** Exercise timeout, metadata failure, malformed marker, and unsupported marker through the main-to-store-to-AppShell boundary and assert the chosen recovery behavior does not silently show fresh-install onboarding to an established profile.

src/main.tsx
```diff
@@ -161,30 +164,48 @@ if (bootError) {
...
-              <ThemeProvider>
-                <UpdaterProvider>
-                  <App />
-                </UpdaterProvider>
-              </ThemeProvider>
-            </I18nProvider>
-          </QueryClientProvider>
-        </RendererErrorBoundary>
-      </TooltipProvider>
+      <I18nProvider>
```
morgmart, line 169: 🤖 **P2 · Startup loader remains hidden** (non-blocking)

Bootstrap mounts StartupLoadingView before cohort resolution, but the main Tauri window is configured initially invisible. Window.show remains in App's effect, and App is not mounted until the cohort request settles or times out. The loading surface therefore cannot provide visible feedback during that wait. The prior automated thread was resolved without a substantive human reply, and the behavior remains, so it is eligible to be reopened.

**User effect:** During a slow cohort lookup, users see no Berd window for up to five seconds and then see it appear abruptly instead of seeing startup progress.

**Recommended fix:** Make bootstrap own initial window visibility and show the main window immediately after mounting the loading surface, while preserving normal App startup and diagnostic behavior.

**Test:** Hold the cohort promise pending and spy on the Tauri window API, asserting the loading surface is mounted and the main window is shown before resolution or timeout.

src/shared/ui/dialog.tsx
morgmart, line 4: 🤖 **P3 · Dialog still forks icon system** (non-blocking)

The shared Dialog primitive replaces the established icon-library close affordance with a private inline filled SVG. This changes every dialog globally and creates an undocumented one-off icon recipe rather than using or extending the shared icon system. The prior automated thread was resolved without a substantive human reply, and the implementation remains, so it is eligible to be reopened.

**User effect:** Every dialog receives a close affordance that can drift in weight, sizing, and future treatment from the rest of Berd's icons.

**Recommended fix:** Use the established shared icon-family component, or introduce and document a reusable system-owned close icon rather than embedding a private SVG in Dialog.

**Test:** Add shared Dialog explorer or visual coverage asserting the approved close icon primitive at supported sizes and themes.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison found four blocking and two non-blocking publishable findings. The telemetry failure-containment change fixes the prior core-flow failure, but telemetry consent still defaults on through the general progression action and has no app-wide lifecycle for later opt-in or revocation. Cohort aggregation can discard definitive established-install evidence when the other probe fails, while unknown classification still silently enters first-run onboarding. The startup loader remains mounted behind an initially hidden native window, and the shared dialog still forks the icon system. Five still-present issues were suppressed because their prior automated threads remain unresolved; the prior telemetry-failure issue was verified fixed. The final self-check covered fresh and established onboarding flows, consent and revocation, feedback telemetry, startup success/error/timeout/visibility, migration and concurrent writers, responsive/reflow behavior, accessibility, i18n, navigation gates, tests, design-system rules, and duplicate overlap. Supplied GitHub evidence was inspected: all eight reported checks completed successfully. No local code, tests, builds, package managers, hooks, scripts, or generated binaries were run.

Deterministic publication result: 4 blocking and 2 non-blocking finding(s) publishable; 5 duplicate(s) suppressed.

src/shared/i18n/locales/en/onboarding.json
```diff
@@ -3,12 +3,19 @@
...
+    "getStarted": "Let’s go",
+    "usageConsent": "Share anonymous usage data to help improve Berd.",
+    "learnMore": "Learn more",
+    "usageDialog": {
+      "title": "Sharing usage data",
+      "intro": "We use a random installation ID, not your identity, to see how Berd is used and fix what’s broken.",
+      "collectTitle": "What we collect",
+      "collectBody": "Usage events, such as features used, errors, and performance, tied to a random installation ID.",
+      "notCollectTitle": "What we don’t collect",
+      "notCollectBody": "Your name, email, or account. Berd has no login, so there’s nothing to tie data to."
```
morgmart, line 17: 🤖 **P1 · Disclosure overstates telemetry anonymity** (blocking)

The disclosure says usage events are tied to a persistent random installation ID, then claims that because Berd has no login there is nothing to tie data to. The events remain linkable to an installation even when they are not linked to an account.

**User effect:** People may agree to usage sharing based on a stronger anonymity promise than Berd actually provides.

**Recommended fix:** Explain that Berd does not send account identity fields while usage events remain associated with a random installation identifier, and verify categorical collection claims against the actual payload.

**Test:** Add copy coverage confirming every supported locale distinguishes account identity from installation-level linkage.

src/features/feedback/submitFeedbackReport.ts
```diff
@@ -58,7 +59,9 @@ export async function submitFeedbackReport(
     doctorReport,
     labelIds: input.labelIds,
   });
-  trackFeedbackSubmitted();
+  if (getTelemetryConsent() === true) {
+    trackFeedbackSubmitted();
```
morgmart, line 63: 🤖 **P1 · Telemetry can invalidate successful feedback** (blocking)

The feedback issue is created before trackFeedbackSubmitted runs, but a telemetry exception rejects the whole function after the issue already exists, causing the UI to report failure for a completed submission.

**User effect:** Someone can be told their feedback failed and retry even though it was submitted, creating duplicate tickets and uncertainty.

**Recommended fix:** Treat issue creation as the transaction boundary and run feedback telemetry afterward through best-effort failure containment without rejecting the successful result.

**Test:** Make feedback telemetry throw after successful issue creation and assert the report still resolves, the telemetry error is reported, and the UI does not show submission failure.

src/shared/i18n/locales/en/onboarding.json
```diff
@@ -3,12 +3,19 @@
...
   },
   "welcome": {
-    "title": "Welcome to Berd",
-    "projects": "A place to build projects",
-    "agents": "work with agents",
-    "and": "and",
-    "done": "get work done",
-    "getStarted": "Get started"
+    "title": "Welcome to Berd.",
+    "subtitle": "Your place for doing.",
```
morgmart, line 7: 🤖 **P2 · Welcome does not explain Berd** (non-blocking)

The new subtitle, Your place for doing, gives no concrete model of Berd as an ongoing workspace for real work with agents, so setup begins before the user understands what they are configuring.

**User effect:** A first-time user sees a polished but generic introduction and must continue before learning what Berd is or why agent setup matters.

**Recommended fix:** Add a concise sentence identifying Berd as a continuous workspace for doing real work with AI agents.

**Test:** Update English and Spanish copy coverage to confirm the welcome introduces the workspace and agents before setup.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The completed four-lens review found two blocking findings and one non-blocking finding. The telemetry disclosure overstates anonymity, optional feedback telemetry can make a successful submission appear to fail, and the welcome copy does not explain Berd's agent-workspace purpose. Eleven still-present issues are suppressed by authoritative unresolved automated threads. This PR changes graphical onboarding, consent, startup, settings, and shared dialog surfaces, but the supplied evidence contains no screenshots or recording; please add screenshots or a short screen recording. One Atlas escalation remains for telemetry consent policy, durable ownership, and revocation behavior. All eight supplied checks passed; required checks still govern merge readiness.

Deterministic publication result: 2 blocking and 1 non-blocking inline finding(s) publishable; 11 duplicate(s) suppressed; 1 blocking screenshot-evidence requirement(s) in this review body.

PR #82, opened 2026-08-18
feat(chat): default to gender-neutral pronouns via an always-on interaction norms preamble

delkc:
## Why
Agents and chats were assigning he/him or she/her pronouns to people and agents without any prior instruction. The default for anyone — the user, people they mention, or agents — should be they/them unless pronouns are stated or clearly established.

## What
- Add a tiny always-on `[Defaults]` interaction-norms preamble (~50 tokens) in `interactionNorms.ts`
- Inject it on every send for goose-managed sessions under a new keyed section (`berd_interaction_norms`)
- Fold it into the in-band persona handoff for external harnesses (Claude Code, Codex, …), ahead of the persona
- Place it before user-authored content on both paths, and state explicitly that user-given pronouns always win
- Add an authoring rule to the builder persona (`agt-builder.md`) and the `agent-builder` skill: generated agents get no gender or gendered pronouns unless the creator asks — needed because personas override the runtime preamble

## Risk Assessment
Low — adds one short system-prompt block per send and two authoring-guidance lines; no behavior change beyond the stated default, and users override it by simply stating pronouns.

Generated with Goose

### DIFF
--- a/distro/agents/agt-builder.md
+++ b/distro/agents/agt-builder.md
@@ -26,7 +26,7 @@ Build in the open. Draft and show the concrete proposed agent or diff first. Wai
 
 **Ask before you assume, especially on voice.** If they haven't said how the agent should sound, don't invent a personality and hope it's close. Ask directly, or offer two contrasting options and let them react. An agent's voice is the hardest thing to get right by guessing.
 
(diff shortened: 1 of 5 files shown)


tulsi-builder (approved):
APPROVED

PR #83, opened 2026-08-18
import agents from zip files

cynfria:
**Category:** new-feature
**User Impact:** Users can now import an agent directly from a ZIP file created by Berd’s agent export flow.

**Problem:** Exported agent ZIPs could be shared conveniently, but importing them required manually extracting the archive first. **Solution:** Accept ZIPs in the existing reviewed import flow, safely locate one supported agent payload, enforce archive and payload limits, and present localized validation errors.

<details>

**src/features/agents/lib/agentZipImport.ts**
Adds bounded ZIP extraction for one supported PNG, Markdown, or JSON agent payload, with typed validation failures for the UI.

**src/features/agents/lib/agentZipImport.test.ts**
Covers supported payloads, macOS metadata, malformed and ambiguous archives, filename handling, and nested size limits.

**src/features/agents/ui/AgentImportDialog.tsx**
Allows ZIP files through the existing file picker and drag-and-drop import surface.

**src/features/agents/ui/AgentsView.tsx**
Routes ZIP payloads through the same preview and confirmation flows as direct agent files and localizes ZIP-specific errors.

**src/features/agents/ui/__tests__/AgentsView.entry.test.tsx**
Verifies portable agent image ZIPs preview and continue into configuration, and ambiguous archives fail visibly.

**src/shared/i18n/locales/en/agents.json**
Updates supported-format guidance and adds English ZIP validation messages.

**src/shared/i18n/locales/es/agents.json**
Updates supported-format guidance and adds Spanish ZIP validation messages.

</details>

### DIFF
--- a/src/features/agents/lib/agentZipImport.ts
+++ b/src/features/agents/lib/agentZipImport.ts
@@ -60,6 +60,7 @@ export function extractAgentFileFromZip(
   archiveBytes: Uint8Array,
 ): ExtractedAgentFile {
   let entryCount = 0;
+  let supportedEntryCount = 0;
   let totalUncompressedBytes = 0;
   let archive: Record<string, Uint8Array>;
   try {
@@ -74,14 +75,19 @@ export function extractAgentFileFromZip(
           throw new AgentZipImportError("tooLarge", MAX_SNAPSHOT_PNG_BYTES);
         }
         const name = entry.name.split("/").at(-1) ?? "";
-        if (
+        const isSupportedEntry =
           !entry.name.endsWith("/") &&
           !entry.name.split("/").includes("__MACOSX") &&
           !name.startsWith(".") &&
-          isSupportedAgentFileName(name) &&
-          entry.originalSize > maxAgentFileBytes(name)
-        ) {
-          throw new AgentZipImportError("tooLarge", maxAgentFileBytes(name));
+          isSupportedAgentFileName(name);
+        if (isSupportedEntry) {
+          supportedEntryCount += 1;
+          if (supportedEntryCount > 1) {
+            throw new AgentZipImportError("multipleAgents");
+          }
+          if (entry.originalSize > maxAgentFileBytes(name)) {
+            throw new AgentZipImportError("tooLarge", maxAgentFileBytes(name));
+          }
         }
         return !entry.name.endsWith("/");
       },
--- a/src/features/agents/lib/agentZipImport.ts
+++ b/src/features/agents/lib/agentZipImport.ts
@@ -1,0 +1,115 @@
+import { unzipSync } from "fflate";
+import { MAX_SNAPSHOT_PNG_BYTES } from "@/features/agents/agent-snapshot";
+import { MAX_PERSONA_IMPORT_BYTES } from "@/features/agents/lib/personaImport";
+
+const MAX_ARCHIVE_ENTRIES = 32;
+
+export type AgentZipImportErrorCode =
+  | "invalid"
+  | "tooManyFiles"
+  | "tooLarge"
+  | "missingAgent"
+  | "multipleAgents";
+
+export class AgentZipImportError extends Error {
+  constructor(
+    public readonly code: AgentZipImportErrorCode,
+    public readonly maxBytes?: number,
+  ) {
+    super(code);
+    this.name = "AgentZipImportError";
+  }
+}
+
+export interface ExtractedAgentFile {
+  bytes: Uint8Array;
+  name: string;
+}
+
+export function isAgentZipFileName(fileName: string): boolean {
+  return fileName.trim().toLowerCase().endsWith(".zip");
+}
+
+function isAgentImageFileName(fileName: string): boolean {
+  return fileName.toLowerCase().endsWith(".png");
+}
+
+function isSupportedAgentFileName(fileName: string): boolean {
+  const lowerName = fileName.toLowerCase();
+  return (
+    isAgentImageFileName(lowerName) ||
+    lowerName.endsWith(".md") ||
+    lowerName.endsWith(".json")
+  );
+}
+
+function maxAgentFileBytes(fileName: string): number {
+  return isAgentImageFileName(fileName)
+    ? MAX_SNAPSHOT_PNG_BYTES
+    : MAX_PERSONA_IMPORT_BYTES;
+}
+
+function validateExtractedSize(fileName: string, size: number): void {
+  const maxBytes = maxAgentFileBytes(fileName);
+  if (size > maxBytes) {
+    throw new AgentZipImportError("tooLarge", maxBytes);
+  }
+}
+
+export function extractAgentFileFromZip(
+  archiveBytes: Uint8Array,
+): ExtractedAgentFile {
+  let entryCount = 0;
+  let totalUncompressedBytes = 0;
+  let archive: Record<string, Uint8Array>;
+  try {
+    archive = unzipSync(archiveBytes, {
+      filter(entry) {
+        entryCount += 1;
+        if (entryCount > MAX_ARCHIVE_ENTRIES) {
+          throw new AgentZipImportError("tooManyFiles");
+        }
+        totalUncompressedBytes += entry.originalSize;
+        if (totalUncompressedBytes > MAX_SNAPSHOT_PNG_BYTES) {
+          throw new AgentZipImportError("tooLarge", MAX_SNAPSHOT_PNG_BYTES);
+        }
+        const name = entry.name.split("/").at(-1) ?? "";
+        if (
+          !entry.name.endsWith("/") &&
+          !entry.name.split("/").includes("__MACOSX") &&
+          !name.startsWith(".") &&
+          isSupportedAgentFileName(name) &&
+          entry.originalSize > maxAgentFileBytes(name)
+        ) {
+          throw new AgentZipImportError("tooLarge", maxAgentFileBytes(name));
+        }
+        return !entry.name.endsWith("/");
+      },
+    });
+  } catch (error) {
+    if (error instanceof AgentZipImportError) throw error;
+    throw new AgentZipImportError("invalid");
+  }
+
+  const candidates = Object.entries(archive).filter(([path]) => {
+    const parts = path.split("/");
+    const name = parts.at(-1) ?? "";
+    return (
+      !parts.includes("__MACOSX") &&
+      !name.startsWith(".") &&
+      isSupportedAgentFileName(name)
+    );
+  });
+
+  if (candidates.length === 0) {
+    throw new AgentZipImportError("missingAgent");
+  }
+  if (candidates.length > 1) {
+    throw new AgentZipImportError("multipleAgents");
+  }
+
+  const [path, bytes] = candidates[0];
+  const name = path.split("/").at(-1) ?? path;
+  validateExtractedSize(name, bytes.length);
+  return { bytes, name };
+}
--- a/src/features/agents/lib/agentZipImport.ts
+++ b/src/features/agents/lib/agentZipImport.ts
@@ -56,6 +56,61 @@ function validateExtractedSize(fileName: string, size: number): void {
   }
 }
 
+export const AGENT_ZIP_IMPORT_TIMEOUT_MS = 15_000;
+
+export function extractAgentFileFromZipInWorker(
+  archiveBytes: Uint8Array,
+  signal?: AbortSignal,
+  timeoutMs = AGENT_ZIP_IMPORT_TIMEOUT_MS,
+): Promise<ExtractedAgentFile> {
+  return new Promise((resolve, reject) => {
+    const worker = new Worker(
+      new URL("./agentZipImport.worker.ts", import.meta.url),
+      { type: "module" },
+    );
+    let settled = false;
+    const finish = (operation: () => void) => {
+      if (settled) return;
+      settled = true;
+      window.clearTimeout(timeout);
+      signal?.removeEventListener("abort", handleAbort);
+      worker.terminate();
+      operation();
+    };
+    const handleAbort = () =>
+      finish(() => reject(new DOMException("Aborted", "AbortError")));
+    const timeout = window.setTimeout(
+      () => finish(() => reject(new AgentZipImportError("invalid"))),
+      timeoutMs,
+    );
+    worker.onmessage = (
+      event: MessageEvent<
+        | ExtractedAgentFile
+        | {
+            error: { code: AgentZipImportErrorCode; maxBytes?: number };
+          }
+      >,
+    ) => {
+      if ("error" in event.data) {
+        const { code, maxBytes } = event.data.error;
+        finish(() => reject(new AgentZipImportError(code, maxBytes)));
+      } else {
+        const extracted = event.data;
+        finish(() => resolve(extracted));
+      }
+    };
+    worker.onerror = () =>
+      finish(() => reject(new AgentZipImportError("invalid")));
+    if (signal?.aborted) {
+      handleAbort();
+      return;
+    }
+    signal?.addEventListener("abort", handleAbort, { once: true });
+    const workerBytes = new Uint8Array(archiveBytes);
+    worker.postMessage({ archiveBytes: workerBytes }, [workerBytes.buffer]);
+  });
+}
+
 export function extractAgentFileFromZip(
   archiveBytes: Uint8Array,
 ): ExtractedAgentFile {
--- a/src/features/agents/ui/AgentImportDialog.tsx
+++ b/src/features/agents/ui/AgentImportDialog.tsx
@@ -77,6 +77,7 @@ export function AgentImportDialog({
     null,
   );
   const preparationRef = useRef<AbortController | null>(null);
+  const [preparing, setPreparing] = useState(false);
   const [prepared, setPrepared] = useState<{
     bytes: Uint8Array;
     name: string;
@@ -144,6 +145,7 @@ export function AgentImportDialog({
     onImportFile: async (bytes, name) => {
       const controller = new AbortController();
       preparationRef.current = controller;
+      setPreparing(true);
       try {
         const result = await prepareImport(bytes, name, controller.signal);
         if (controller.signal.aborted) return;
@@ -159,6 +161,7 @@ export function AgentImportDialog({
       } finally {
         if (preparationRef.current === controller) {
           preparationRef.current = null;
+          setPreparing(false);
         }
       }
     },
@@ -232,6 +235,8 @@ export function AgentImportDialog({
           ) : (
             <div
               {...dropHandlers}
+              role="status"
+              aria-busy={preparing}
               className={cn(
                 "flex min-h-56 flex-col items-center justify-center gap-4 rounded-md border border-dashed border-border bg-muted/40 px-6 text-center",
                 isDragOver && "border-ring bg-muted/70",
@@ -241,11 +246,16 @@ export function AgentImportDialog({
                 className="size-10 text-muted-foreground"
                 aria-hidden="true"
               />
-              <p className="text-sm">{t("importDialog.dropTitle")}</p>
+              <p className="text-sm">
+                {preparing
+                  ? t("importDialog.preparing")
+                  : t("importDialog.dropTitle")}
+              </p>
               <Button
                 type="button"
                 variant="outline"
                 leftIcon={<IconUpload />}
+                feedbackState={preparing ? "loading" : "idle"}
                 onClick={openFilePicker}
               >
                 {t("importDialog.openFinder")}
--- a/src/features/agents/ui/AgentImportDialog.tsx
+++ b/src/features/agents/ui/AgentImportDialog.tsx
@@ -222,7 +222,7 @@ export function AgentImportDialog({
           <input
             ref={fileInputRef}
             type="file"
-            accept=".agent.png,.png,.persona.md,.md,.json,image/png,text/markdown,text/plain,application/json"
+            accept=".agent.zip,.zip,.agent.png,.png,.persona.md,.md,.json,application/zip,application/x-zip-compressed,image/png,text/markdown,text/plain,application/json"
             className="hidden"
             onChange={handleFileChange}
           />
--- a/src/features/agents/ui/AgentImportDialog.tsx
+++ b/src/features/agents/ui/AgentImportDialog.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { IconPhotoPlus, IconUpload } from "@tabler/icons-react";
 
 import { useTranslation } from "react-i18next";
@@ -37,6 +37,12 @@ export interface AgentImportPreview extends PersonaImportPreview {
 interface AgentImportDialogProps {
   open: boolean;
   onOpenChange: (open: boolean) => void;
+  /**
+   * A file already validated and read by another import surface (the gallery
+   * drop zone). The dialog prepares it through the same flow as picker files
+   * so every entry point shares one preview/confirmation owner.
+   */
+  initialFile?: { bytes: Uint8Array; name: string } | null;
   onImportFile: (
     fileBytes: Uint8Array,
     fileName: string,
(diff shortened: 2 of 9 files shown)


src/features/agents/ui/AgentsView.tsx
```diff
@@ -480,16 +522,27 @@ export function AgentsView({
...
           onImportFile={handleImportFileBytes}
           prepareImport={(bytes, name) => {
-            if (isAgentImageFileName(name)) {
-              const snapshot = decodeAgentImage(bytes);
-              const { width, height } = getPngDimensions(bytes);
+            let extracted: ExtractedAgentFile;
+            try {
+              extracted = isAgentZipFileName(name)
+                ? extractAgentFileFromZip(bytes)
+                : { bytes, name };
```
morgmart: 🤖 **P1 · Rejected ZIP leaves stale import armed** (blocking)

The new ZIP preparation path can throw for malformed, oversized, missing-agent, or ambiguous archives, but that failure only reports an error; it does not invalidate AgentImportDialog's previously prepared file. After previewing agent A, selecting rejected ZIP B therefore leaves A's preview and enabled Import agent button in place. This makes the confirmation surface disagree with the user's latest file selection and crosses the import consent boundary with stale state.

**User effect:** A user who selects a replacement ZIP, sees an error, and then presses Import agent can import the previously selected agent instead of the file they most recently chose.

**Recommended fix:** Clear or replace the prepared import as soon as a new selection attempt begins, before validation, reading, or ZIP extraction, so every rejected replacement removes the old preview and disables confirmation. Keep the selected file and prepared payload under one generation/source of truth.

**Test:** Add a dialog-level regression test that previews valid agent A, then selects a malformed or ambiguous ZIP B, verifies the localized error, and verifies A's preview and the Import agent button are no longer available and no import can be confirmed.

src/features/agents/lib/agentZipImport.ts
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return !entry.name.endsWith("/");
+      },
+    });
+  } catch (error) {
+    if (error instanceof AgentZipImportError) throw error;
+    throw new AgentZipImportError("invalid");
+  }
+
+  const candidates = Object.entries(archive).filter(([path]) => {
```
morgmart, line 155: 🤖 **P2 · Duplicate ZIP entries evade ambiguity check** (non-blocking)

Candidate counting happens after unzipSync has converted entries into a record keyed by path. A valid ZIP may contain repeated entries with the same path; those entries collapse to one record key before candidates.length is checked, so an archive containing multiple supported agent payload entries can be accepted as exactly one. The existing ambiguity test only covers distinct filenames.

**User effect:** For a ZIP with duplicate agent filenames, Berd silently chooses the payload retained by the ZIP library instead of explaining that the archive is ambiguous, so the preview/import result can differ from what the archive creator intended.

**Recommended fix:** Count supported entry occurrences while enumerating the ZIP directory, before extraction collapses entries by pathname, and reject the archive when more than one supported occurrence exists (including duplicate paths). Then extract only the single validated occurrence.

**Test:** Add a handcrafted ZIP fixture containing two supported entries with the same pathname and assert extractAgentFileFromZip throws AgentZipImportError with code multipleAgents.

morgmart (changes requested):
🤖 Automated code review

Fresh static review completed for the exact three-dot comparison f31e8e02790924b1d64b783d52517e95b1d72544...04a2e0baebf48905b7c088d6f5cd3f09a05cc3de using both the project code-review and wes-review lenses. Final self-check covered the ZIP picker/drop, preview, confirmation, PNG/persona import, malformed/ambiguous/oversized archive flows; accessibility; English and Spanish localization; navigation and consent guards; async reads, errors, never-completes, lifecycle, and races; test honesty; project/design-system rules; duplicate overlap; and blocking evidence plus user effect. Two findings remain: one blocking consent/state-consistency defect and one non-blocking archive-boundary defect. Recommend REQUEST_CHANGES. Supplied GitHub evidence was inspected: checks were structurally available for the exact head; Frontend checks and unit tests were still in progress while the reported Rust, frozen-install, close, and DCO checks had succeeded. Required checks independently govern merge readiness.

src/features/agents/lib/agentZipImport.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export function extractAgentFileFromZip(
+  archiveBytes: Uint8Array,
+): ExtractedAgentFile {
+  let entryCount = 0;
+  let supportedEntryCount = 0;
+  let totalUncompressedBytes = 0;
+  let archive: Record<string, Uint8Array>;
+  try {
+    archive = unzipSync(archiveBytes, {
```
morgmart, line 122: 🤖 **P1 · ZIP extraction blocks the renderer** (blocking)

The import boundary passes the entire untrusted archive to synchronous unzipSync on the renderer thread. The filter inflates every non-directory entry, including irrelevant files, and its resource accounting trusts each entry's declared originalSize before decompression. A crafted DEFLATE stream can make the decoder perform far more expansion work than those declarations imply, while an ordinary archive can still force up to the full allowed payload to be decompressed synchronously. The reviewed dialog path also repeats extraction when confirmation reopens the original ZIP, doubling this work. Because no worker, cancellation, or deadline exists, the desktop UI can become unresponsive while parsing attacker-controlled input.

**User effect:** Opening a malicious or expensive ZIP can freeze Berd before the user sees a preview or has a chance to consent, potentially requiring the app to be force-quit.

**Recommended fix:** Move ZIP inspection and extraction off the renderer thread into a worker with cancellation when the file is replaced or the dialog closes, enforce limits against actual emitted bytes rather than only ZIP metadata, and inflate only the single validated supported entry. Retain that extracted payload as the prepared source of truth so confirmation does not decompress the archive again.

**Test:** Add a worker-level regression test using an archive whose declared size does not provide the effective work bound, assert extraction terminates at the actual-byte/deadline limit, and add a dialog test proving replacement or close cancels an in-flight extraction and a stale result cannot update the preview.

src/features/agents/ui/AgentsView.tsx
```diff
@@ -413,6 +432,19 @@ export function AgentsView({
 
   const validateImportFile = useCallback(
     (file: Pick<File, "name" | "type" | "size">) => {
+      if (isAgentZipFileName(file.name)) {
+        if (file.size > MAX_SNAPSHOT_PNG_BYTES) {
+          return t("view.importTooLarge", { maxSize: "10 MB" });
+        }
+        if (
+          file.type &&
```
morgmart: 🤖 **P2 · Valid ZIP MIME types are rejected** (non-blocking)

ZIP validation accepts an empty MIME type and only two specific ZIP MIME strings. File.type is supplied by the platform rather than derived authoritatively from archive contents, and valid ZIPs are commonly exposed as application/octet-stream or application/x-zip. Those files have an accepted .zip filename and would be safely validated by the ZIP parser, but this changed MIME gate rejects them before parsing.

**User effect:** Some users can select a valid Berd agent ZIP and receive an unsupported-file error solely because their operating system or WebView labels the file differently.

**Recommended fix:** Treat the .zip extension plus successful archive parsing as authoritative, or allow the common generic ZIP MIME variants, including application/octet-stream and application/x-zip, while retaining parser validation.

**Test:** Add validation tests for valid .zip files reported as application/octet-stream and application/x-zip and verify they proceed to archive parsing, while a non-ZIP extension remains rejected.

morgmart (changes requested):
🤖 Automated code review

Fresh static review completed for the full exact three-dot comparison f31e8e02790924b1d64b783d52517e95b1d72544...b7d4c27111313a70e89737f95e67b7f5117b1378 using the project code-review and user wes-review lenses. The two previously reported issues were rechecked and are fixed: replacement selection now clears stale prepared state, and supported entries are counted before duplicate paths collapse. Final self-check covered file-picker and drag/drop entry, ZIP/PNG/Markdown/JSON validation, preview and confirmation, malformed/missing/ambiguous/oversized archives, accessibility, English and Spanish localization, navigation and consent guards, async reads, errors, never-completes, lifecycle and races, test honesty, design-system/project rules, duplicate overlap, and concrete blocking evidence plus user effect. Two new publishable findings remain, including a blocking availability issue from synchronous untrusted archive processing. Recommend REQUEST_CHANGES. Supplied GitHub evidence was inspected and was structurally valid for the exact head: Frontend checks and unit tests and Linux Rust checks were in progress; Windows/macOS Rust, clean-room frozen install, and DCO had succeeded. Required checks independently govern merge readiness.

src/features/agents/ui/AgentsView.tsx
```diff
@@ -457,16 +483,26 @@ export function AgentsView({
       preview?: { snapshot?: ReturnType<typeof decodeAgentImage> },
     ) => {
       try {
-        if (isAgentImageFileName(fileName)) {
+        const extracted = isAgentZipFileName(fileName)
```
morgmart: 🤖 **P1 · Gallery ZIP drops bypass safe preparation** (blocking)

The validator shared with PersonaGallery now accepts .zip files, so a ZIP dropped on the gallery is passed directly to handleImportFileBytes. That changed handler still calls synchronous extractAgentFileFromZip on the renderer instead of the new worker-backed dialog preparation path. It also sends extracted Markdown or JSON directly to importPersonas without the preview and explicit Import action used by the dialog. Thus the same untrusted ZIP has two security and consent models depending on where it is dropped, and the worker timeout, cancellation, prepared source of truth, and confirmation guard do not protect the gallery path. The earlier renderer-blocking thread was resolved without a substantive human reply, but this underlying issue still exists at this current changed-line path, so it is reopenable rather than suppressed.

**User effect:** Dropping an expensive ZIP on the agents gallery can freeze Berd, and a ZIP containing Markdown or JSON can create agents immediately without letting the user review and confirm the extracted agent first.

**Recommended fix:** Route gallery ZIP drops into the same import dialog/coordinator that owns worker extraction, parsing, preview, cancellation, and confirmation, then commit the exact prepared bytes and inner filename. As a temporary safety measure, reject ZIPs from the gallery drop path until that unified flow exists.

**Test:** Add a gallery drag-and-drop integration test for a ZIP that asserts worker-backed preparation is used, no persona is created before the preview's Import action, and confirming imports the exact prepared inner payload. Also cover replacement or cancellation while preparation is pending.

src/features/agents/ui/AgentImportDialog.tsx
```diff
@@ -123,17 +141,32 @@ export function AgentImportDialog({
     handleFileChange,
     openFilePicker,
   } = useFileImportZone({
-    onImportFile: (bytes, name) => {
+    onImportFile: async (bytes, name) => {
+      const controller = new AbortController();
+      preparationRef.current = controller;
       try {
-        const preview = prepareImport(bytes, name);
+        const result = await prepareImport(bytes, name, controller.signal);
```
morgmart: 🤖 **P2 · ZIP preparation has no pending state** (non-blocking)

Dialog ZIP extraction is now asynchronous and can remain pending until the 15-second timeout, but the dialog tracks only the eventual prepared result. During that wait it continues to show the unchanged drop target and enabled picker with no visible status, disabled state, aria-busy attribute, or live announcement. Cancellation prevents stale completion, but users receive no feedback that work is in progress.

**User effect:** After choosing a larger ZIP, users may think nothing happened, repeatedly select files, or wait without knowing whether Berd is processing the archive or has become stuck; assistive-technology users receive no progress signal at all.

**Recommended fix:** Track a preparation state in the dialog, show localized loading copy or a progress treatment, expose the region as busy/status to assistive technology, and define whether choosing a replacement remains available as an intentional cancel-and-restart action.

**Test:** Add a deferred-preparation dialog test that verifies a visible localized pending message and accessible busy/status semantics before resolution, then verifies they clear on success, error, replacement, and close.

morgmart (changes requested):
🤖 Automated code review

Fresh static review completed for the full exact three-dot comparison f31e8e02790924b1d64b783d52517e95b1d72544...59420e596adf900fd0b8edbbe85f45126803bba3 using the project code-review and user wes-review lenses. The resolved stale-preparation, duplicate-entry, and ZIP MIME issues are fixed. The resolved renderer-blocking issue remains present through a different current changed-line entry path and is therefore reopenable under the supplied duplicate policy: gallery drag-and-drop accepts ZIPs but bypasses the new worker-backed preview coordinator. Final self-check covered dialog picker and gallery drop entry, ZIP/PNG/Markdown/JSON validation, worker success/error/timeout/abort behavior, preview and confirmation, malformed/missing/ambiguous/oversized archives, accessibility, English and Spanish localization, navigation and consent guards, async errors and never-completes behavior, lifecycle and races, test honesty, design-system/project rules, duplicate overlap, and blocking evidence plus user effect. Two publishable findings remain. Recommend REQUEST_CHANGES. Supplied GitHub evidence was inspected and structurally valid for the exact head: Frontend checks and unit tests and Linux Rust checks were in progress; Windows/macOS Rust, clean-room frozen install, and DCO had succeeded. Required checks independently govern merge readiness.

src/features/agents/ui/AgentsView.tsx
```diff
@@ -457,16 +482,28 @@ export function AgentsView({
       preview?: { snapshot?: ReturnType<typeof decodeAgentImage> },
     ) => {
       try {
-        if (isAgentImageFileName(fileName)) {
+        // A .zip name only reaches this handler from the gallery drop zone;
+        // the import dialog always passes the extracted inner file.
+        const extracted = isAgentZipFileName(fileName)
+          ? await extractAgentFileFromZipInWorker(fileBytes)
```
morgmart: 🤖 **P1 · Gallery ZIP preparations can race** (blocking)

The gallery path now moves ZIP extraction into a worker, but it calls extractAgentFileFromZipInWorker without an AbortSignal or generation check and still commits extracted text directly instead of routing through the dialog's preparation owner. If ZIP A is dropped and ZIP B is dropped before A finishes, both workers remain live: A can resolve last and replace B's image preview, or a text payload can import after the newer selection and without any preview confirmation. Navigation away likewise does not cancel this work. The existing resolved gallery safe-preparation thread had no substantive human reply; its underlying lifecycle and consent issue remains at this current changed line, so it may be reopened.

**User effect:** A user's latest dropped file does not reliably win: an earlier ZIP can surface for confirmation or create agents after a later drop or navigation, and Markdown/JSON ZIPs are still imported without the review-and-confirm step shown by the import dialog.

**Recommended fix:** Give all import entry surfaces one coordinator. Route gallery drops into the same dialog preparation flow so one owner validates, starts a cancellable worker, ignores stale generations, previews the extracted agent, and commits the exact prepared bytes only after confirmation. At minimum, add gallery-scoped abort/generation ownership and do not directly commit extracted ZIP text.

**Test:** Add deferred worker tests for gallery drops A then B resolving out of order and for unmount/navigation during extraction. Assert A cannot update UI or persist after B, pending work is cancelled on lifecycle exit, and a Markdown/JSON ZIP does not call importPersonas until the user confirms its preview.

src/features/agents/ui/AgentImportDialog.tsx
```diff
@@ -208,11 +246,16 @@ export function AgentImportDialog({
...
+              <p className="text-sm">
+                {preparing
+                  ? t("importDialog.preparing")
+                  : t("importDialog.dropTitle")}
+              </p>
               <Button
                 type="button"
                 variant="outline"
                 leftIcon={<IconUpload />}
+                feedbackState={preparing ? "loading" : "idle"}
```
morgmart, line 308: 🤖 **P2 · Loading button bypasses localization** (non-blocking)

The pending surface adds localized importDialog.preparing copy, but the Button is placed in feedbackState=loading without a loadingLabel. The shared Button defaults that feedback label to the hard-coded English string “Loading...”, replacing the localized Open Finder label while preparation is pending. The resolved pending-state thread had no substantive human reply, and localization of the pending control remains incomplete.

**User effect:** Spanish users see an English “Loading...” button inside an otherwise Spanish import dialog while a ZIP is being opened.

**Recommended fix:** Pass a localized loadingLabel, using importDialog.preparing or a dedicated localized button label, while retaining the localized status text and aria-busy state.

**Test:** Extend the pending preparation test to render with a non-English translation and assert the loading button exposes the translated loading label rather than the Button default.

morgmart (changes requested):
🤖 Automated code review

Fresh static review completed for the full exact three-dot comparison f31e8e02790924b1d64b783d52517e95b1d72544...e412b74b08b4ae5f9dbc44aea0acd3cb1e939237 using the project code-review and user wes-review lenses. The stale dialog preparation, duplicate ZIP entry, MIME rejection, renderer-thread dialog extraction, and absence of a visible pending state are substantially fixed. The resolved gallery safe-preparation thread had no substantive human reply and its underlying ownership issue still exists: gallery ZIP extraction now uses a worker, but remains outside the dialog's replacement, cancellation, preview, and consent coordinator, so it is reopenable rather than suppressed. Final self-check covered dialog picker and gallery drop flows; ZIP/PNG/Markdown/JSON validation; worker success, error, timeout, abort, replacement, unmount, and race behavior; preview and confirmation; accessibility; English and Spanish localization; navigation and consent guards; test honesty; design-system/project rules; duplicate overlap; and concrete blocking evidence plus user effect. Two publishable findings remain. Recommend REQUEST_CHANGES. Supplied GitHub evidence was inspected and structurally valid for the exact head: Frontend checks and unit tests were in progress; Windows, macOS, and Linux Rust checks, clean-room frozen install, and DCO had succeeded. Required checks independently govern merge readiness.

src/features/agents/ui/AgentImportDialog.tsx
```diff
@@ -116,24 +142,59 @@ export function AgentImportDialog({
...
 
+  const startPreparation = useCallback(
+    async (bytes: Uint8Array, name: string) => {
+      preparationRef.current?.abort();
+      const controller = new AbortController();
+      preparationRef.current = controller;
+      setPreparing(true);
+      try {
+        const result = await prepareImport(bytes, name, controller.signal);
+        if (controller.signal.aborted) return;
```
morgmart: 🤖 **P2 · Stale previews can leak blob URLs** (non-blocking)

prepareImport creates a cardImageUrl before its async result returns to AgentImportDialog. If replacement or close aborts that preparation after the URL is created but before the dialog continuation accepts the result, the aborted guard discards the result. The existing cleanup effect only revokes URLs that reached prepared state, so the discarded preview URL has no owner and is never revoked. This can occur through microtask ordering when another file read completes between the async preparation returning and the dialog continuation resuming.

**User effect:** Repeatedly replacing or closing image or image-ZIP imports at this race boundary can accumulate renderer-managed blob resources for the rest of the app session.

**Recommended fix:** Create the object URL only after the current preparation wins inside the dialog, or give every preparation result an explicit disposal path and revoke cardImageUrl whenever a result is aborted, stale, rejected, replaced, closed, or unmounted before becoming prepared.

**Test:** Add a race test where an aborted or replaced preparation resolves with a cardImageUrl after the abort, then assert URL.revokeObjectURL is called for that discarded result and the stale preview never renders.

morgmart (commented):
🤖 Automated code review

Fresh static review completed for the full exact three-dot comparison f31e8e02790924b1d64b783d52517e95b1d72544...671149a8dd558ce8c3d51641f79b9863af36885b using the project code-review and user wes-review lenses. All prior automated issues were rechecked and are fixed: rejected replacements clear stale state; duplicate supported ZIP entries are counted before path collapse; platform MIME does not wrongly reject ZIPs; extraction runs in a bounded, cancellable worker; gallery imports route into the same preview-and-confirm coordinator; latest selection wins and close/unmount aborts pending work; pending status is visible, accessible, and localized. Final self-check covered picker selection and gallery drag/drop for ZIP, PNG, Markdown, and JSON; malformed, missing, ambiguous, oversized, and never-completing archives; preview-to-commit source-of-truth consistency; navigation and consent guards; async success, error, timeout, replacement, close, unmount, and race behavior; accessibility; English and Spanish localization; test honesty and discriminating coverage; design-system and project rules; duplicate/overlapping findings; and blocking evidence plus user effect. One new non-blocking resource-lifecycle finding remains. Recommend COMMENT. Supplied GitHub evidence was inspected and structurally valid for the exact head: DCO had succeeded and the combined status was pending; required checks independently govern merge readiness.

src/features/agents/ui/AgentImportDialog.tsx
```diff
@@ -116,24 +142,66 @@ export function AgentImportDialog({
...
+        if (preparationRef.current === controller) {
+          preparationRef.current = null;
+          setPreparing(false);
+        }
+      }
+    },
+    [onImportError, prepareImport],
+  );
+
+  const consumedInitialFileRef = useRef<typeof initialFile>(null);
```
morgmart, line 185: 🤖 **P1 · StrictMode aborts gallery preparation** (blocking)

The new initialFile effect marks a gallery-dropped file as consumed before starting its preparation. A separate mount effect cleanup aborts the active preparation. Berd renders under React.StrictMode, whose development effect replay runs mount cleanups and then reruns effects: the first preparation is aborted, but consumedInitialFileRef still equals the same file, so the replayed initialFile effect refuses to restart it. The dialog remains open with no preview or error. Existing tests render this component without StrictMode and therefore do not exercise the app's actual development lifecycle.

**User effect:** In Berd's normal development build, dropping an agent file on the gallery can open the import dialog but never produce a preview or Import action, leaving the user stuck until they choose the file again through the dialog.

**Recommended fix:** Make initial-file preparation replay-safe: an effect cleanup may abort its current attempt, but an open and still-current initialFile must be eligible to restart after cleanup. Track successfully accepted input generations rather than marking the file consumed before preparation begins, or reset the consumption marker when that attempt is aborted by lifecycle cleanup without the file changing.

**Test:** Render the gallery-drop flow under React.StrictMode and assert that effect replay aborts the obsolete attempt but restarts preparation for the still-current initial file, ultimately showing its preview; also assert stale attempts cannot win or leak a preview URL.

morgmart (changes requested):
🤖 Automated code review

Fresh static review completed for the full exact three-dot comparison 8c06ce5723ae114988f6e579dc98e971ae3f76b0...bc693a22a6436f920a7dda7139fea9086e9ea21d using the project code-review and user wes-review lenses. All prior automated issues were rechecked and are fixed, including stale preparation state, duplicate ZIP entries, MIME handling, renderer blocking, gallery consent and race ownership, pending localization, and disposal of late aborted preview URLs. Final self-check covered picker and gallery selection/drop for ZIP, PNG, Markdown, and JSON; accessibility; English and Spanish localization; navigation and consent guards; async success, error, timeout, never-completes, replacement, close, unmount, StrictMode lifecycle, and races; archive boundaries; preview URL ownership; test honesty; design-system/project rules; duplicate overlap; and concrete blocking evidence plus user effect. One new blocking lifecycle finding remains. Recommend REQUEST_CHANGES. Supplied GitHub evidence was inspected and structurally valid for the exact head: Frontend checks and unit tests were in progress; clean-room frozen install, Windows/macOS/Linux Rust checks, and DCO had succeeded. Required checks independently govern merge readiness.

morgmart (approved):
🤖 Automated code review

Fresh static review completed for the full exact three-dot comparison 8c06ce5723ae114988f6e579dc98e971ae3f76b0...912c3b157ff7c08f4508c7851d3b722e2f49d1ab using the project code-review and user wes-review lenses. All prior automated issues were rechecked and are fixed, including stale replacement state, duplicate ZIP entries, platform MIME handling, renderer-thread extraction, gallery preview/consent ownership and races, pending feedback and localization, late blob URL disposal, and StrictMode replay. Final self-check covered picker and gallery selection/drop for ZIP, PNG, Markdown, and JSON; accessibility; English and Spanish localization; navigation and consent guards; async success, errors, timeout, never-completes, replacement, close, unmount, StrictMode, and races; archive boundaries; preview-to-commit consistency and resource ownership; test honesty and discriminating coverage; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. Potential concerns that were already present in the exact base rather than caused by this comparison were excluded. No publishable findings or unresolved suppressions remain. Recommend APPROVE. Supplied GitHub evidence was inspected and structurally valid for the exact head: all reported check runs had completed successfully; the combined status remained pending. Required checks independently govern merge readiness, and the publisher refreshes checks before approval.

PR #93, opened 2026-08-18
fix sidebar empty section behavior

cynfria:
**Category:** fix
**User Impact:** Empty sidebar sections now behave consistently, and newly created projects remain immediately visible when their first chat opens.

**Problem:** The empty Chats header used different ordering and collapse behavior from its populated state, including a caret when there was nothing to collapse. Creating a project could also open its first chat without exposing the project in a previously collapsed Projects section.

**Solution:** Reuse the shared section-header treatment for empty Chats while making empty chat lists non-collapsible. Carry an explicit, window-local project-created revision to the session list so only local creation expands Projects, without treating ordinary backend list refreshes as creation.

<details>

**src/app/AppShell.tsx**
Records explicit local project creation and passes that signal into sidebar navigation while preserving the existing project-save workflow.

**src/app/views/NavigationPanesView.tsx**
Forwards the project-created revision to the session-list capability.

**src/app/views/__tests__/NavigationPanesView.test.tsx**
Covers the static empty Chats label and verifies that only explicit local project creation expands Projects.

**src/features/sessions/capabilities/SessionListCapability.tsx**
Responds to the explicit creation signal by opening Projects without inferring intent from project-list changes.

**src/features/sessions/ui/session-list/SidebarProjectsSection.tsx**
Replaces the bespoke empty Chats header with the shared non-interactive section header.

**src/features/sessions/ui/session-list/SidebarRecentsSection.tsx**
Keeps empty chat actions visible and removes collapse interaction when there are no chats.

**src/features/sessions/ui/session-list/__tests__/SidebarRecentsSection.test.tsx**
Adds regression coverage for the non-collapsible empty Chats state.

</details>

### DIFF
--- a/src/features/sessions/capabilities/SessionListCapability.tsx
+++ b/src/features/sessions/capabilities/SessionListCapability.tsx
@@ -161,6 +161,7 @@ export interface SessionListCapabilityProps {
   onReorderProject?: (fromId: string, toId: string) => void;
   onSelectSession?: (sessionId: string) => void;
   onSessionSelectForScroll?: (sessionId: string) => void;
+  onProjectCreatedRevisionHandled?: (revision: number) => void;
   projectCreatedRevision?: number;
   projects: ProjectInfo[];
   surface: SessionListSurfaceOptions;
@@ -480,6 +481,7 @@ export function SessionListCapability({
   onReorderProject,
   onSelectSession,
   onSessionSelectForScroll,
+  onProjectCreatedRevisionHandled,
   projectCreatedRevision = 0,
   projects,
   surface,
@@ -554,15 +556,17 @@ export function SessionListCapability({
     () => new Set(projects.map((project) => project.id)),
     [projects],
   );
-  const initialProjectCreatedRevisionRef = useRef(projectCreatedRevision);
   useEffect(() => {
-    if (projectCreatedRevision === initialProjectCreatedRevisionRef.current) {
-      return;
-    }
+    if (projectCreatedRevision === 0) return;
     setSectionVisibility((current) =>
       current.projects ? current : { ...current, projects: true },
     );
-  }, [projectCreatedRevision, setSectionVisibility]);
+    onProjectCreatedRevisionHandled?.(projectCreatedRevision);
+  }, [
+    onProjectCreatedRevisionHandled,
+    projectCreatedRevision,
+    setSectionVisibility,
+  ]);
   const projectsById = useMemo(
     () => new Map(projects.map((project) => [project.id, project])),
     [projects],
--- a/src/features/sessions/capabilities/SessionListCapability.tsx
+++ b/src/features/sessions/capabilities/SessionListCapability.tsx
@@ -161,6 +161,7 @@ export interface SessionListCapabilityProps {
   onReorderProject?: (fromId: string, toId: string) => void;
   onSelectSession?: (sessionId: string) => void;
   onSessionSelectForScroll?: (sessionId: string) => void;
+  projectCreatedRevision?: number;
   projects: ProjectInfo[];
   surface: SessionListSurfaceOptions;
 }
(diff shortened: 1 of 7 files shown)


src/features/sessions/capabilities/SessionListCapability.tsx
```diff
@@ -552,6 +554,15 @@ export function SessionListCapability({
     () => new Set(projects.map((project) => project.id)),
     [projects],
   );
+  const initialProjectCreatedRevisionRef = useRef(projectCreatedRevision);
```
morgmart: 🤖 **P1 · Preserve creation across remounts** (blocking)

SessionListCapability is not mounted while the sidebar is collapsed, but a project can still be created from the Home surface. If projectCreatedRevision increments during that interval, this ref initializes to the already-incremented value when the sidebar is expanded. The effect then considers that creation already observed and never opens Projects. The added test only changes the revision while the component remains mounted, so it does not exercise this lifecycle path.

**User effect:** A user who creates a project while the sidebar is collapsed can open the sidebar after the first project chat starts and still find Projects collapsed, so the newly created project is not exposed as this fix promises.

**Recommended fix:** Keep the creation intent in an owner that survives SessionListCapability unmounts, or initialize/consume the monotonic revision so a nonzero revision received on remount is handled rather than adopted as the baseline. Add a discriminating test that creates or increments the revision while the session list is absent, remounts it with Projects persisted closed, and verifies the new project is visible and the persisted section is open.

**Test:** Render with the sidebar/session list absent and Projects persisted closed, increment projectCreatedRevision, then render or expand NavigationPanesView and assert the project row is visible; the current implementation should fail this test.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The exact three-dot comparison was reviewed by static inspection at head e25465ca8fcff9ae05f6a1cbfe721d33b769493c. One blocking lifecycle bug remains in the core project-creation flow: a creation signal emitted while the session list is unmounted is treated as the component's baseline and discarded when the sidebar returns. The supplied GitHub evidence was inspected; Rust, DCO, clean-room install, and close checks passed, while Frontend checks and unit tests were still in progress at capture time. Required checks independently govern merge readiness.

src/features/sessions/capabilities/SessionListCapability.tsx
cynfria: 🤖 Fixed in 3d137698. The creation revision now remains pending until SessionListCapability handles it, so a nonzero signal received on mount opens Projects and is then acknowledged by AppShell. I also added a regression test that mounts the session list with Projects persisted closed and a pending creation revision.

cynfria:
🤖 The blocking lifecycle finding is fixed in 3d137698, with regression coverage for a pending creation signal received on mount. The review thread is resolved and all required checks pass. Please re-review the latest head.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static inspection of the complete exact three-dot comparison at head 3d1376982a5eaddc1295321296c1005fd9f8ecf7 found no publishable findings. The prior automated remount-lifecycle issue is fixed: AppShell now retains a nonzero pending creation revision while the session list is absent, SessionListCapability opens Projects when it mounts with that signal, and acknowledgement clears only the revision actually handled so a concurrent newer creation remains pending. Regression coverage exercises both ordinary project-list refreshes and initial mount with a pending creation signal. The empty Chats states use localized copy and shared design-system primitives, remain keyboard-accessible, expose no false disclosure control, and keep the start-chat action visible. Navigation/consent behavior, async and failure paths, lifecycle and race behavior, accessibility, localization, design-system rules, test honesty, and changed-flow coverage were rechecked. Supplied GitHub evidence was inspected; all listed check runs completed successfully, while the legacy combined commit status was pending with no status contexts. Required checks independently govern merge readiness.

PR #124, opened 2026-08-19
fix agent share card details

cynfria:
**Category:** fix
**User Impact:** Shared agent cards now show accurate, concise descriptions and real “Good for” and “Vibes” details without exposing private instructions.
**Problem:** Agent cards could show placeholder or inferred data, truncate long instruction text, and lose curated metadata across imports and downloads. A stale installed Agt. Builder could also continue shadowing the corrected bundled definition.
**Solution:** Carry validated card metadata through every supported format, derive safe concise fallback descriptions, offer explicit model-assisted generation, and narrowly migrate only the exact known stale bundled artifact.

<details>

**scripts/validate-bundled-agents.ts**
Requires short, grapheme-bounded `good_for` and `vibes` metadata for bundled agents.

**src-tauri/src/services/bundled_agents.rs**
Recognizes and replaces only the exact historical stale Agt. Builder file while preserving modified and user-owned files.

**src-tauri/test-fixtures/legacy-agt-builder.md**
Provides the exact historical artifact used to verify the migration signature.

**src/features/agents/agent-snapshot/mapping.test.ts**
Covers portable card metadata and Unicode-safe round trips.

**src/features/agents/agent-snapshot/mapping.ts**
Carries `goodFor` and `vibes` through portable agent snapshots.

**src/features/agents/agent-snapshot/schema.ts**
Validates portable card metadata with grapheme-aware bounds.

**src/features/agents/ui/AgentImportDialog.tsx**
Uses public preview descriptions and real card metadata without exposing imported instructions.

**src/features/agents/ui/__tests__/AgentImportDialog.test.tsx**
Covers authored and generic privacy-safe import preview descriptions.

**src/features/agents/ui/share-card/AgentShareDialog.test.tsx**
Covers description generation lifecycle, cancellation, reopening, and preview/export consistency.

**src/features/agents/ui/share-card/AgentShareDialog.tsx**
Displays real metadata, supports explicit model-generated descriptions, and renders reviewed copy into downloads.

**src/features/agents/ui/share-card/agentShareCard.test.ts**
Verifies instruction-derived card descriptions and existing rendering behavior.

**src/features/agents/ui/share-card/agentShareCard.ts**
Accepts the reviewed description when rendering downloadable cards.

**src/features/agents/ui/share-card/agentShareCardCopy.test.ts**
Covers metadata precedence, length limits, and readable localization fallbacks.

**src/features/agents/ui/share-card/agentShareCardCopy.ts**
Prefers validated agent metadata while retaining short classified fallbacks.

**src/features/agents/ui/share-card/agentShareCardDescription.test.ts**
Covers safe extraction and generic fallback behavior.

**src/features/agents/ui/share-card/agentShareCardDescription.ts**
Derives concise public descriptions from recognized role statements without copying arbitrary instructions.

**src/features/agents/ui/share-card/agentShareCardDescriptionInference.test.ts**
Covers hidden tool-free inference, caching, provider readiness, and malformed-output fallback.

**src/features/agents/ui/share-card/agentShareCardDescriptionInference.ts**
Generates optional descriptions through a hidden, cancellable, tool-free model session after explicit user action.

**src/shared/api/__tests__/agents.test.ts**
Covers metadata hydration, precedence, sanitization, imports, exports, and native JSON promotion.

**src/shared/api/agents.ts**
Normalizes and preserves card metadata across source listing, Markdown, JSON, import previews, and exports.

**src/shared/i18n/locales/en/agents.json**
Adds English copy for optional description generation.

**src/shared/i18n/locales/es/agents.json**
Adds Spanish copy for optional description generation.

**src/shared/lib/graphemeCount.ts**
Provides shared Unicode grapheme counting for card limits.

**src/shared/runtime-config/validateBundledAgents.test.ts**
Covers required and bounded bundled-agent card metadata.

**src/shared/types/agents.ts**
Adds typed card metadata to personas and creation requests.

</details>

### Verification

- `just check`
- `just tauri-check`
- focused frontend test suites
- bundled-agent migration tests
- review-fix loop with final general and Wes reviews clean


src/features/agents/ui/share-card/agentShareCard.ts
```diff
@@ -25,7 +25,7 @@ export function getAgentShareCardBase(_personaId: string): string {
 }
 
 export function getAgentShareDescription(persona: Persona): string {
-  return getRealPersonaDescription(persona) ?? persona.systemPrompt.trim();
+  return deriveAgentCardDescription(persona.systemPrompt, persona.displayName);
```
morgmart: 🤖 **P1 · Preserve authored public descriptions** (blocking)

The changed implementation unconditionally derives the card description from systemPrompt. Before this PR, getAgentShareDescription preferred sourceDescription when it was a real authored description. This now discards the explicit public-facing source of truth and may extract a role or purpose from operational instructions instead; the changed test explicitly makes unrelated sourceDescription metadata acceptable, so it preserves the regression rather than catching it.

**User effect:** An agent owner can author a deliberate public description but see a generic or instruction-derived sentence on the preview and downloaded public card, potentially revealing details they intentionally kept in the agent instructions rather than the public description.

**Recommended fix:** Restore the real authored sourceDescription as the first choice. Only derive from instructions when no real public description exists, and use the same resolved description for preview and rendering.

**Test:** Add a regression test with a real sourceDescription and a different sensitive job/purpose in systemPrompt; assert that both the visible preview and rendered/downloaded card use only the authored description. Keep a separate fallback test for personas without a real description.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -372,6 +395,7 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
         })?;
         match installed_agent_path_state(target)? {
             InstalledAgentPathState::Missing | InstalledAgentPathState::Bundled => {}
+            InstalledAgentPathState::UserOwned if is_known_legacy_agt_builder(target)? => {}
```
morgmart, line 402: 🤖 **P1 · Bind migration check to replacement** (blocking)

The migration authorizes replacing agt-builder.md by hashing it, but after this second hash check it performs an unconditional rename. Another Berd process or external editor can modify the file between the check and rename, so the newly user-owned contents are overwritten. This changed exception is the path that permits replacing a file otherwise classified as user-owned.

**User effect:** A user editing the historical Agent Builder file during startup can lose their edits when Berd replaces the file with the bundled version.

**Recommended fix:** Make the replacement conditional on the same file identity/content that was verified, using an atomic or concurrency-safe compare-and-replace strategy. If the target changes at any point, preserve it and skip migration.

**Test:** Add a deterministic race test that changes the target after legacy recognition but before replacement and asserts the modified user file remains untouched; retain the exact historical-file replacement test.

src/features/agents/ui/share-card/agentShareCardDescriptionInference.ts
```diff
@@ -0,0 +1,205 @@
...
+    const readiness = await readDefaultProviderReadiness({ coalesce: true });
+    if (
+      readiness.status !== "ready" ||
+      !readiness.modelId ||
+      options.signal?.aborted
+    ) {
+      return fallback;
+    }
+    if (
+      (options.providerId && options.providerId !== readiness.providerId) ||
```
morgmart: 🤖 **P2 · Use the configured generation model** (non-blocking)

Generation compares persona.provider and persona.model directly with the default readiness provider/model and returns the fallback on any mismatch. Berd personas can identify the harness as provider "goose" while carrying the concrete provider separately in modelProviderId, and an agent can use a model different from the configured default. In those normal cases the Generate action silently performs no inference even though the UI presents it as available.

**User effect:** Clicking Generate description can appear to complete but leave the same fallback copy unchanged, with no explanation that the configured model was never called.

**Recommended fix:** Choose one supported model source for this best-effort action—preferably the ready configured default—or compare against the persona's concrete modelProviderId rather than its harness ID. If generation is intentionally unavailable, disable or explain the action instead of silently returning unchanged copy.

**Test:** Add an integration-level dialog/inference test for a Goose persona with modelProviderId different from persona.provider and assert that clicking Generate creates the hidden session with the intended configured provider/model; also cover an intentional unavailable state.

src/shared/api/agents.ts
```diff
@@ -948,12 +1017,22 @@ export async function exportPersona(id: string): Promise<ExportResult> {
...
   /** Only local/data-backed media is safe to render before import consent. */
   avatar?: string;
   identity: string;
 }
 
+function previewDescription(displayName: string, description: unknown): string {
+  const authored = trimmedPropertyString(description);
+  return hasRealAgentDescription(authored)
+    ? (authored as string)
+    : `${displayName} is ready to help.`;
```
morgmart: 🤖 **P2 · Localize import fallback copy** (non-blocking)

previewDescription constructs the user-visible fallback `${displayName} is ready to help.` inside the API layer. That English sentence is then shown by the localized import dialog, so Spanish and future locales cannot translate it.

**User effect:** Users running Berd in Spanish see English fallback copy on an imported agent preview when the file has no authored description.

**Recommended fix:** Return an absent/typed fallback state from the API and let AgentImportDialog resolve a localized i18n key, or pass localized fallback copy into the preview boundary. Remove the duplicate hard-coded component fallback as part of the same source-of-truth cleanup.

**Test:** Render an import without a real description under the Spanish locale and assert that the preview contains the Spanish translation and no English fallback sentence.

src/features/agents/agent-snapshot/schema.ts
```diff
@@ -152,6 +156,20 @@ export function validateSnapshotV1(value: unknown): SnapshotV1 {
   const profile = isRecord(value.profile) ? value.profile : undefined;
   if (profile) {
     optionalString(profile, "displayName", MAX_SNAPSHOT_NAME_LENGTH);
+    optionalString(profile, "goodFor");
+    optionalString(profile, "vibes");
+    if (
+      typeof profile.goodFor === "string" &&
+      graphemeCount(profile.goodFor) > 44
```
morgmart: 🤖 **P3 · Validate normalized snapshot metadata** (non-blocking)

Snapshot validation counts goodFor and vibes before the mapping layer trims them. A value with exactly 44 or 32 meaningful graphemes plus incidental surrounding whitespace is rejected even though the normalized value that would be persisted is within the documented limit.

**User effect:** An otherwise valid shared agent image can fail to import solely because its bounded card metadata contains leading or trailing spaces.

**Recommended fix:** Apply the same normalization before enforcing grapheme limits at snapshot ingress, or explicitly reject non-normalized whitespace with a clear contract and ensure exports always produce that canonical form.

**Test:** Add snapshot validation and round-trip cases for max-length Unicode metadata surrounded by whitespace; assert the normalized bounded values import successfully.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the exact three-dot comparison introduces two blocking regressions: public share cards discard authored public descriptions in favor of instruction-derived text, and the legacy bundled-agent replacement can overwrite a concurrently edited user file. Three non-blocking issues remain around model-backed generation compatibility, localization, and snapshot normalization. Supplied GitHub evidence was inspected; frontend checks were still in progress at capture time, while the other reported checks had completed successfully. Required checks independently govern merge readiness.

Deterministic publication result: 2 blocking and 3 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/features/agents/ui/share-card/agentShareCard.ts
cynfria: 🤖 Authored public descriptions now take precedence; instruction derivation is only the fallback.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Legacy replacement now atomically claims the verified file and re-verifies its exact contents before installation.

src/features/agents/ui/share-card/agentShareCardDescriptionInference.ts
cynfria: 🤖 Generation now consistently uses the ready configured default provider and model.

src/shared/api/agents.ts
cynfria: 🤖 The API now returns no fallback description; the import dialog supplies localized fallback copy.

src/features/agents/agent-snapshot/schema.ts
cynfria: 🤖 Snapshot limits now count normalized, trimmed metadata.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -370,22 +401,64 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
...
+        ) && is_known_legacy_agt_builder(target)?;
+        if claimed_legacy {
+            fs::rename(target, &legacy_claim_path).map_err(|err| {
+                format!(
+                    "Failed to claim legacy bundled agent '{}': {err}",
+                    target.display()
+                )
+            })?;
+            if !has_known_legacy_agt_builder_contents(&legacy_claim_path)? {
+                let _ = fs::rename(&legacy_claim_path, target);
```
morgmart: 🤖 **P1 · Restore migration without replacement** (blocking)

The new claim-and-reverify flow closes the original overwrite window, but its changed-content recovery calls fs::rename(legacy_claim_path, target) and ignores the result. On platforms where rename replaces an existing destination, another writer can recreate agt-builder.md after the claim and before this recovery; the recovery then overwrites that newly user-owned file. This is the same data-preservation invariant as the prior resolved automation thread, but a different remaining race, and that thread has no substantive human reply.

**User effect:** A user or second Berd process saving Agent Builder during startup can still lose the newly saved file if the migration detects a concurrent change and restores its claimed copy over the new target.

**Recommended fix:** Restore the claimed legacy file with no-replace semantics, just as the publish path preserves a recreated target. If target exists, leave it untouched and retain or safely clean up the claim without overwriting user data; handle restoration failure explicitly rather than discarding it.

**Test:** Add a deterministic concurrent-writer test that recreates target after the claim but before the failed re-verification recovery, then assert the recreated user file remains byte-for-byte intact. Cover both restoration into an empty path and conflict with an occupied path.

src/features/agents/ui/share-card/AgentShareDialog.tsx
```diff
@@ -305,6 +360,7 @@ export function AgentShareDialog({
           cardBase,
           cardCopy,
           locale,
+          description,
```
morgmart, line 317: 🤖 **P2 · Persist the reviewed description** (non-blocking)

The generated description is supplied to renderAgentShareCard, so it appears in the downloaded PNG artwork, but the embedded snapshot is still built from the unchanged persona and personaToSnapshot writes profile.about as null. The reviewed/generated value therefore is not part of the portable agent data even though goodFor and vibes now are.

**User effect:** A recipient can import a card whose artwork shows the generated description, but re-sharing that imported agent produces different fallback copy because the reviewed description was lost in the round trip.

**Recommended fix:** Make the resolved public description part of the snapshot profile and map it back to CreatePersonaRequest.description on import, with the same validation and size limits as other public card metadata. Ensure authored descriptions remain the source of truth unless the user explicitly generated a replacement.

**Test:** Generate a description, export the PNG snapshot, map it back to a create request, and assert the exact reviewed description is retained and used by a subsequent share-card render. Include an authored-description round trip as well.

src/features/agents/ui/share-card/agentShareCardDescriptionInference.ts
```diff
@@ -0,0 +1,197 @@
...
+export async function generateAgentCardDescription(
+  instructions: string,
+  displayName: string,
+  options: GenerateDescriptionOptions = {},
+): Promise<string> {
+  const fallback = deriveAgentCardDescription(instructions, displayName);
+  if (!instructions.trim() || options.signal?.aborted) return fallback;
+
+  try {
+    const readiness = await readDefaultProviderReadiness({ coalesce: true });
```
morgmart: 🤖 **P2 · Bound the full generation setup** (non-blocking)

The 15-second timeout only applies once promptForText starts. The preceding readiness lookup, hidden-session creation, model selection, extension listing/removal, and system-prompt setup are awaited without a shared deadline; aborting the dialog does not make all of those promises settle either. If any setup call never completes, the effect never clears generationPending while the dialog remains open.

**User effect:** After clicking Generate description, the control can remain disabled on “Generating…” forever, with no fallback or retry path until the dialog is closed.

**Recommended fix:** Apply one abort-aware deadline to the entire generation operation, not only prompting, and return the existing extractive fallback when setup times out. Cleanup any session that was created, and keep completion idempotent if timeout and a late ACP response race.

**Test:** Use a never-resolving readiness call and a never-resolving setup call in separate tests; advance the overall deadline and assert the fallback resolves, pending UI clears, retry is enabled, and any created session is cancelled/deleted without accepting late completion.

src/features/agents/ui/share-card/AgentShareDialog.tsx
```diff
@@ -211,18 +212,72 @@ export function AgentShareDialog({
...
   const cardBase = getAgentShareCardBase(persona.id);
-  const description = getAgentShareDescription(persona);
+  const fallbackDescription = getAgentShareDescription(persona);
+  const [generatedDescriptionState, setGeneratedDescription] = useState<{
+    identity: string;
+    value: string;
+  }>();
+  const [generationRequested, setGenerationRequested] = useState(false);
+  const [generationPending, setGenerationPending] = useState(false);
+  const descriptionIdentity = [
```
morgmart: 🤖 **P2 · Key generated copy by locale** (non-blocking)

descriptionIdentity includes the persona fields but not locale. After generation completes, generationRequested is false, so changing the application language leaves generatedDescriptionState eligible and continues displaying the sentence produced for the previous locale even though cardContentIdentity and card labels update.

**User effect:** A user who switches languages while the Share dialog is open can see and download a mixed-language card: translated labels with the previously generated description.

**Recommended fix:** Include locale in the generated-description identity, or clear the generated result when locale changes so the localized fallback is shown until the user generates again. Avoid automatically invoking a model solely because the app language changed.

**Test:** Generate under Spanish, switch to English without closing the dialog, and assert the Spanish generated sentence is no longer displayed or rendered; then generate again and assert the English result is used.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found one blocking migration race and three non-blocking share-card generation lifecycle/consistency issues. All five prior automated threads were inspected; their original issues are fixed except that the migration's replacement path still has a different concurrent-writer overwrite window, which is eligible to be raised again because the resolved thread has no substantive human reply. Supplied GitHub evidence was inspected: frontend checks and unit tests, macOS/Windows Rust checks, clean-room install, and DCO had succeeded; Linux Rust checks were in progress and two additional checks were queued at capture time. Required checks independently govern merge readiness.

Deterministic publication result: 1 blocking and 3 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Legacy recovery now also uses no-replace semantics: it restores only into an empty target path and preserves any concurrently recreated user file.

src/features/agents/ui/share-card/AgentShareDialog.tsx
cynfria: 🤖 The reviewed/generated public description now round-trips through snapshot profile.about, and import maps it back to the persona description.

src/features/agents/ui/share-card/AgentShareDialog.tsx
cynfria: 🤖 Generated description identity now includes locale, so switching languages clears prior generated copy without automatically calling the model.

src/features/agents/ui/share-card/agentShareCardDescriptionInference.ts
cynfria: 🤖 I’m leaving the full-operation deadline open as a follow-up rather than expanding this PR’s inference lifecycle. The current user escape hatch is closing the dialog, which aborts the request and restores the synchronous fallback; prompt execution itself remains bounded to 15 seconds.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -370,22 +401,80 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
...
+                    target.display()
+                ));
+            }
+            fs::remove_file(&temp_path).map_err(|err| {
+                format!(
+                    "Failed to remove temporary bundled agent '{}': {err}",
+                    temp_path.display()
+                )
+            })?;
+            let _ = fs::remove_file(&legacy_claim_path);
```
morgmart: 🤖 **P1 · Retain the claimed legacy inode** (blocking)

The migration renames the verified historical file to legacy_claim_path, publishes the replacement, and immediately removes the claim. A process that opened the original file before the rename can still write through that descriptor after the hash check because it references the same claimed inode. Removing the claim then unlinks the only path to those concurrent edits. This is a remaining form of the same user-data preservation invariant discussed in the resolved migration threads; those threads contain no substantive human reply, and this distinct open-descriptor race is still present.

**User effect:** A user or second process editing Agent Builder during startup can successfully write changes that are then silently deleted when the migration removes its claimed file.

**Recommended fix:** Do not delete the claimed inode during the same migration. Preserve it as a recoverable backup/claim and reconcile or clean it only after the system can establish that no concurrent edits landed, with a policy that always prefers preserving ambiguous user data over cleanup.

**Test:** Open the legacy target for writing before migration, pause after verification/publication, write through the original descriptor, then complete migration and assert the edited claimed contents remain recoverable and are never silently unlinked.

src/features/agents/agent-snapshot/schema.ts
```diff
@@ -152,6 +156,21 @@ export function validateSnapshotV1(value: unknown): SnapshotV1 {
   const profile = isRecord(value.profile) ? value.profile : undefined;
   if (profile) {
     optionalString(profile, "displayName", MAX_SNAPSHOT_NAME_LENGTH);
+    optionalString(profile, "about", 110);
```
morgmart: 🤖 **P1 · Validate descriptions by grapheme** (blocking)

The new profile.about limit uses JavaScript string length, while generated descriptions are accepted up to 110 graphemes. Emoji and many non-Latin graphemes use multiple UTF-16 code units, so a description accepted and displayed by generation can fail validateSnapshotV1 when the user downloads the card. The same 110-code-unit cap also makes existing authored public descriptions over that size newly fail personaToSnapshot instead of exporting.

**User effect:** Users can review a valid-looking description and then have PNG or ZIP download fail, especially for emoji-rich or non-Latin copy; agents with longer authored descriptions can also lose the ability to download a share card.

**Recommended fix:** Give profile.about a documented boundary consistent with the card description contract: apply a cheap input-size guard, normalize it, then enforce the intended grapheme limit. For longer authored descriptions, derive or safely truncate the portable card copy rather than rejecting the whole download.

**Test:** Round-trip and download a 110-grapheme description made of multi-code-unit emoji and assert success. Add an authored description longer than the card limit and assert download still succeeds with deterministic bounded portable copy.

src/features/agents/ui/share-card/AgentShareDialog.tsx
```diff
@@ -211,18 +212,73 @@ export function AgentShareDialog({
...
+    vibes: persona.vibes,
+  });
   const cardContentIdentity = [
     locale,
     persona.id,
     persona.avatar,
     persona.displayName,
     persona.systemPrompt,
     persona.sourceDescription,
+    description,
```
morgmart, line 230: 🤖 **P1 · Keep the generated card ready** (non-blocking)

A generated description changes cardContentIdentity, causing the layout effect to clear avatarReadySrc. The hidden preload image still has the same src, key, and callback ref because only the description changed, so its load event and ref callback do not run again to restore readiness. The card falls back to its loader until the ten-second avatar timeout forces another candidate. The generation test does not first make the card ready, so it misses this transition.

**User effect:** Clicking Generate on an already visible card makes the preview disappear and disables card download behind a loader for roughly ten seconds, even though the avatar is already decoded.

**Recommended fix:** Do not reset avatar readiness for text-only card-content changes, or immediately re-confirm an already complete preload image when content identity changes. Keep avatar identity/readiness separate from rendered text identity.

**Test:** Mark the initial card ready, generate a new description without changing the avatar, and assert the updated preview remains visible and download stays enabled without advancing the avatar timeout.

src/features/agents/ui/share-card/agentShareCardDescription.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  const purpose = withoutIdentity.match(
+    /Your purpose is\s+(.+?)(?:\.\s|$)/iu,
+  )?.[1];
+  const purposeSentence = purpose
+    ? sentenceWithinLimit(toThirdPerson(purpose))
+    : undefined;
+  if (purposeSentence) return purposeSentence;
+
+  return `${displayName.trim() || "This agent"} helps with focused work.`;
```
morgmart: 🤖 **P2 · Localize description fallback** (non-blocking)

The new extractive fallback recognizes English instruction patterns and otherwise returns the hard-coded English sentence `${displayName} helps with focused work.`. This fallback is used whenever the configured model is unavailable, generation fails, or non-English instructions do not match the English patterns, even though the surrounding share UI and exported card labels are localized.

**User effect:** A Spanish user can see and download a mixed-language card whose labels are Spanish but whose generated fallback description is English.

**Recommended fix:** Separate extraction from presentation and resolve the generic fallback through the agents i18n namespace for the active locale. Keep a locale-independent helper result or pass a localized fallback into the derivation boundary.

**Test:** With Spanish active and non-matching Spanish instructions, force model unavailability and assert the preview and rendered/downloaded card use a Spanish fallback with no English sentence.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found two blocking issues and two non-blocking issues. The legacy migration can still discard edits written through an already-open file descriptor, and the new snapshot description boundary rejects valid generated or authored descriptions based on UTF-16 length, breaking card downloads. The description-generation setup deadline remains an existing unresolved automated thread and is suppressed from new findings. Final self-check covered the share preview/generate/download/import flows, migration lifecycle and concurrency, accessibility, localization, consent boundaries, async/error/never-completes/race behavior, tests, project rules, anchors, overlap, and severity evidence. Supplied GitHub evidence was inspected; all listed check runs had completed successfully, while combined status was still reported as pending at capture time. Required checks independently govern merge readiness.

Deterministic publication result: 2 blocking and 2 non-blocking finding(s) publishable; 1 duplicate(s) suppressed.

src/features/agents/ui/share-card/agentShareCardDescriptionInference.ts
cynfria: 🤖 Model-backed generation and its async lifecycle were removed; cards now always use immediate deterministic copy.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 The exact historical inode is now retained as a recovery backup after migration, so writes through an already-open descriptor remain recoverable.

src/features/agents/agent-snapshot/schema.ts
cynfria: 🤖 Portable descriptions now use a grapheme-aware 110-character boundary, and longer authored descriptions are deterministically bounded before snapshot validation.

src/features/agents/ui/share-card/AgentShareDialog.tsx
cynfria: 🤖 Model generation was removed, so text changes no longer reset avatar readiness or hide an already-ready card.

src/features/agents/ui/share-card/agentShareCardDescription.ts
cynfria: 🤖 The deterministic generic fallback is now supplied by localized share-card copy in the dialog.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -370,22 +401,75 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
...
-        })
+            ));
+        }
+        if claimed_legacy {
+            // Publish without replacement: if another writer recreates the
+            // target after our claim, preserve their file and restore the
+            // claimed legacy file only when the path is still empty.
+            if let Err(err) = fs::hard_link(&temp_path, target) {
+                if !target.exists() {
+                    let _ = fs::rename(&legacy_claim_path, target);
```
morgmart: 🤖 **P1 · Restore backup without replacement** (blocking)

If publishing the replacement hard link fails, recovery checks that target does not exist and then calls fs::rename(legacy_claim_path, target). The existence check and rename are not atomic, and on platforms where rename replaces the destination, another writer can create target between them and have that new user-owned file overwritten. This is a remaining instance of the migration data-preservation invariant from the resolved automation threads, which contain no substantive human reply.

**User effect:** A user or second Berd process saving Agent Builder during the recovery window can lose the newly saved file when migration restores the historical backup over it.

**Recommended fix:** Use atomic no-replace semantics for recovery, such as hard-linking the retained backup into an empty target path and treating an existing destination as a conflict. Never separate an existence check from a replacement-capable rename when preserving concurrent writers.

**Test:** Force publication failure, create a user target immediately before recovery, and assert that recovery returns an error while preserving both the recreated target and retained backup byte-for-byte.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -347,6 +374,10 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
         ".berd-agent-install-{}-{sequence}.tmp",
         std::process::id()
     ));
+    let legacy_claim_path = parent.join(format!(
+        ".berd-agent-legacy-backup-{}-{sequence}.bak",
```
morgmart: 🤖 **P1 · Reserve unique recovery backups** (blocking)

Recovery backup names contain only the process ID and an in-process sequence that restarts at zero. Because successful migrations intentionally retain these backups, a later process can reuse the same PID and sequence. fs::rename(target, legacy_claim_path) can then replace the existing retained backup, destroying the very user-data recovery artifact this change is designed to preserve.

**User effect:** A later app launch can silently delete an older recoverable Agent Builder backup before creating the new one, leaving users unable to recover edits captured in that file.

**Recommended fix:** Atomically reserve a genuinely unique backup path with collision handling and never rename onto an existing recovery backup. A random/UUID component or create-new reservation loop can provide uniqueness while preserving every prior backup.

**Test:** Pre-create the exact backup name the process would otherwise choose, run migration, and assert that the existing backup remains unchanged and a separate unique backup is created.

src/features/agents/agent-snapshot/schema.ts
```diff
@@ -152,6 +156,27 @@ export function validateSnapshotV1(value: unknown): SnapshotV1 {
   const profile = isRecord(value.profile) ? value.profile : undefined;
   if (profile) {
     optionalString(profile, "displayName", MAX_SNAPSHOT_NAME_LENGTH);
+    optionalString(profile, "about");
```
morgmart: 🤖 **P2 · Pre-bound snapshot metadata size** (non-blocking)

profile.about, goodFor, and vibes are passed to graphemeCount without a cheap code-unit or byte bound. Untrusted snapshots can contain megabytes of combining marks that count as one grapheme, forcing full Intl.Segmenter allocation on the UI thread and allowing an extremely large string to pass the grapheme limit into persisted agent metadata.

**User effect:** Opening a crafted agent image can stall the import UI and persist unexpectedly huge public metadata even though these fields are intended to be short card copy.

**Recommended fix:** Apply conservative raw code-unit/byte caps before grapheme segmentation, then normalize and enforce the display-oriented grapheme limits. Keep the raw cap high enough for legitimate multi-code-unit grapheme clusters.

**Test:** Validate a large combining-mark payload that forms one grapheme and assert it is rejected before persistence; retain positive tests for legitimate emoji clusters at the grapheme limits.

src/features/agents/ui/share-card/agentShareCardDescription.ts
```diff
@@ -0,0 +1,116 @@
...
+      .replace(/^write\b/iu, "writes"),
+  );
+}
+
+function sentenceWithinLimit(value: string): string | undefined {
+  const sentence = value.trim().replace(/[,;:]?\s*$/u, "");
+  if (!sentence || graphemeCount(sentence) > MAX_CARD_DESCRIPTION_CHARACTERS) {
+    return undefined;
+  }
+  return /[.!?]$/u.test(sentence) ? sentence : `${sentence}.`;
```
morgmart: 🤖 **P2 · Count appended punctuation** (non-blocking)

sentenceWithinLimit accepts an unpunctuated sentence at exactly 110 graphemes and then appends a period, returning 111 graphemes. The preview and rendered artwork use that value, while personaToSnapshot truncates the embedded profile.about back to 110, so the portable description no longer matches the card the user reviewed.

**User effect:** After importing and re-sharing a card with boundary-length copy, the description can lose its final character or punctuation and differ from the original artwork.

**Recommended fix:** Validate the final punctuated sentence or reserve one grapheme before appending punctuation, so every returned description is within the same 110-grapheme portable contract.

**Test:** Use an unpunctuated 110-grapheme extracted sentence and assert the returned, rendered, snapshotted, and re-imported description are identical and no longer than 110 graphemes.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found two blocking migration races and two non-blocking boundary/round-trip issues. Recovery can still overwrite a target created between an existence check and rename, and intentionally retained recovery backups can be overwritten after a PID/name collision. All supplied review threads were inspected; the recovery race is eligible to be raised again because the related resolved automation threads contain no substantive human reply, while the backup-name collision is a distinct issue. Final self-check covered every changed share/import/download and migration flow, accessibility, i18n, consent/navigation boundaries, async/error/never-completes/lifecycle/races, test honesty, design-system/project rules, duplicate overlap, anchors, and blocking user-effect evidence. Supplied GitHub evidence was inspected: all listed checks except the macOS Rust check had completed successfully, and that check plus combined status were still in progress/pending at capture time. Required checks independently govern merge readiness.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Recovery now uses no-replace hard linking, so a concurrently recreated target is always preserved alongside the retained backup.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Backup directories are atomically reserved with create_dir and collision retry; existing backups are never replaced.

src/features/agents/agent-snapshot/schema.ts
cynfria: 🤖 Snapshot card metadata now has a conservative raw 4,096-code-unit cap before grapheme segmentation.

src/features/agents/ui/share-card/agentShareCardDescription.ts
cynfria: 🤖 Description bounds now validate the final punctuated sentence, so returned and snapshotted copy share the same 110-grapheme contract.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -370,22 +417,80 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
...
-            InstalledAgentPathState::Missing | InstalledAgentPathState::Bundled => {}
-            InstalledAgentPathState::UserOwned => {
+        let claimed_legacy = matches!(
+            installed_agent_path_state(target)?,
+            InstalledAgentPathState::UserOwned
+        ) && is_known_legacy_agt_builder(target)?;
+        let legacy_claim_path = if claimed_legacy {
+            let backup_dir = reserve_legacy_backup_dir(parent)?;
+            let legacy_claim_path = backup_dir.join(LEGACY_AGT_BUILDER_FILE_NAME);
+            fs::rename(target, &legacy_claim_path).map_err(|err| {
```
morgmart: 🤖 **P1 · Recover interrupted legacy claims** (blocking)

The migration moves the canonical agt-builder.md into a hidden backup directory before publishing its replacement. A crash, process termination, or post-rename error before the hard-link publication leaves the canonical path missing. No startup path discovers or restores abandoned backup directories. If the seed marker already records this file, the next startup interprets the missing target as a user deletion and skips reinstalling it, so the agent remains absent even though its bytes survive in a hidden directory. The backup test verifies retention only after successful publication and does not establish recovery.

**User effect:** An interrupted startup can make Agent Builder disappear permanently from the app, with recovery requiring manual discovery of an internal hidden backup path.

**Recommended fix:** Make the claim a recoverable transaction: record enough state before moving the canonical file and reconcile abandoned claims on startup before honoring the seed marker's deletion semantics. Recovery must preserve a concurrently recreated target and either restore the claim or safely complete publication.

**Test:** Simulate interruption immediately after the rename and after post-rename verification, restart seeding with a marker that already contains agt-builder.md, and assert the canonical agent is automatically restored or safely completed without overwriting a concurrent user file.

src/features/agents/agent-snapshot/schema.ts
```diff
@@ -152,6 +157,27 @@ export function validateSnapshotV1(value: unknown): SnapshotV1 {
   const profile = isRecord(value.profile) ? value.profile : undefined;
   if (profile) {
     optionalString(profile, "displayName", MAX_SNAPSHOT_NAME_LENGTH);
+    optionalString(profile, "about", MAX_SNAPSHOT_CARD_COPY_RAW_LENGTH);
+    if (
+      typeof profile.about === "string" &&
```
morgmart: 🤖 **P1 · Preserve v1 snapshot compatibility** (blocking)

profile.about already existed in snapshot version 1 and previously had no 110-grapheme validation limit. This PR adds a hard rejection while leaving the wire version unchanged, so previously valid Berd/Buzz cards with longer public descriptions now fail validation as invalid snapshots. A presentation constraint for the new card should not retroactively invalidate the entire portable agent payload.

**User effect:** Users can no longer import existing version-1 agent cards that worked before this update solely because their public description exceeds the new card-copy length.

**Recommended fix:** Keep v1 decoding backward-compatible: accept the existing about field at its prior safe raw bound, then normalize/truncate or omit only the presentation copy when mapping it into Berd. If hard rejection is truly required as a wire contract, introduce a new snapshot version with explicit compatibility handling.

**Test:** Import a version-1 fixture whose profile.about exceeds 110 graphemes and assert the agent still imports with safely bounded presentation copy; retain rejection tests for genuinely unsafe raw payload sizes.

src/shared/api/agents.ts
```diff
@@ -1008,8 +1097,10 @@ export function previewPersonaImport(
   }
 
   validateLegacyPersonaImport(parsed);
+  const displayName = parsed.displayName as string;
   return {
-    displayName: parsed.displayName as string,
+    displayName,
+    description: previewDescription(parsed.description),
```
morgmart, line 1107: 🤖 **P2 · Persist the previewed legacy description** (non-blocking)

The legacy JSON preview now exposes parsed.description as authored public copy, but the matching import path still constructs the source with the fixed placeholder description "Agent". The confirmation surface and committed result therefore use different sources of truth. The added tests cover the preview but not the post-import persona.

**User effect:** A user can confirm an import after reviewing a meaningful description, then find that the imported agent has lost it and shows generic or derived copy when shared again.

**Recommended fix:** Normalize and preserve the same validated legacy description in legacyPersonaToCreateRequest, or omit it from the preview if legacy import intentionally does not support that field.

**Test:** Preview and import the same legacy JSON with a real description, then assert the returned and persisted persona retains exactly the description shown before confirmation.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found two blocking regressions and one non-blocking preview/import inconsistency. The new legacy-file transaction has no recovery for interruption after moving the canonical file, and the unchanged v1 snapshot version now rejects previously valid long profile.about values. All 17 supplied review threads, their resolution states, replies, authors, bodies, and URLs were inspected; these are distinct current issues rather than duplicates of the resolved automated findings. Final self-check covered every changed share preview/download/import and migration flow, accessibility, i18n, navigation and consent boundaries, async/error/never-completes/lifecycle/races, test honesty, design-system/project rules, duplicate overlap, valid anchors, and concrete blocking user effects. Supplied GitHub evidence was inspected; every listed check run had completed successfully, although combined status remained pending at capture time. Required checks independently govern merge readiness.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Product direction now permits direct replacement of the exact known stale Agt. Builder, so backup/claim transaction state and interruption recovery were removed entirely. Customized files still fail the exact hash match.

src/features/agents/agent-snapshot/schema.ts
cynfria: 🤖 Snapshot v1 again accepts long profile.about values within the raw safety cap; presentation copy is grapheme-bounded only during mapping.

src/shared/api/agents.ts
cynfria: 🤖 Legacy JSON import now persists the same validated authored description shown in preview.

src-tauri/src/services/bundled_agents.rs
```diff
@@ -372,6 +399,7 @@ fn install_agent_file(source: &Path, target: &Path) -> Result<(), String> {
         })?;
         match installed_agent_path_state(target)? {
             InstalledAgentPathState::Missing | InstalledAgentPathState::Bundled => {}
+            InstalledAgentPathState::UserOwned if is_known_legacy_agt_builder(target)? => {}
```
morgmart, line 402: 🤖 **P1 · Bind verification to replacement** (blocking)

The legacy exception re-checks that target still has the historical hash and then calls fs::rename(temp_path, target). Those operations are not atomic. A user or another Berd process can modify or replace agt-builder.md after the hash check but before rename; the rename then replaces the newly user-owned file. This is the original migration data-preservation issue reintroduced by removing the claim flow, and the related resolved automation threads contain no substantive human reply.

**User effect:** A user saving Agent Builder during startup can lose their edits when Berd replaces the file after validating an earlier version of its contents.

**Recommended fix:** Use an atomic compare-and-replace strategy that binds authorization to the exact inode/content being replaced, or preserve the target and skip migration whenever it changes. Do not follow a hash check with an unconditional replacement-capable rename.

**Test:** Deterministically modify or replace target after the final historical-hash check but before publication, then assert the concurrent user file remains byte-for-byte intact and migration does not report success.

src/features/agents/ui/share-card/AgentShareDialog.tsx
```diff
@@ -305,13 +314,15 @@ export function AgentShareDialog({
...
           locale,
+          description,
         );
         if (operationGeneration !== cardOperationGenerationRef.current) return;
         const embeddedAvatar = await avatarSourceToDataUrl(cardAvatarSrc);
         if (operationGeneration !== cardOperationGenerationRef.current) return;
         const snapshot = personaToSnapshot({
           ...persona,
           avatar: embeddedAvatar ?? persona.avatar,
+          sourceDescription: description,
```
morgmart: 🤖 **P2 · Keep fallback copy presentational** (non-blocking)

The download path always overwrites sourceDescription with the resolved card description. For agents without an authored description, that resolved value can be the localized share.descriptionFallback or an English heuristic extracted from instructions. The snapshot then treats temporary presentation copy as authored profile.about, making portable agent metadata depend on the sharer's current locale and fallback behavior.

**User effect:** Sharing an undescribed agent in Spanish can permanently give the recipient a Spanish fallback sentence as the agent's authored description, even when they use another language or re-share it later.

**Recommended fix:** Only write profile.about from a real authored sourceDescription. If deterministic card copy must round-trip, represent it as explicitly generated/presentation metadata with provenance rather than promoting a localized fallback into authored agent data.

**Test:** Export an agent with no real description under English and Spanish, import both snapshots, and assert neither gains an authored sourceDescription while the artwork still uses the correct localized fallback.

src/shared/lib/graphemeCount.ts
```diff
@@ -0,0 +1,5 @@
+export function graphemeCount(value: string, locale = "en"): number {
+  return Array.from(
```
morgmart, line 3: 🤖 **P3 · Guard grapheme segmentation support** (non-blocking)

The new shared graphemeCount helper constructs Intl.Segmenter unconditionally. The existing share-card text utility explicitly handles runtimes where Intl.Segmenter is unavailable, but snapshot validation and agent metadata normalization now call this unguarded helper. An older or constrained WebView can therefore throw while listing or importing agents instead of using a safe fallback.

**User effect:** On a runtime without Intl.Segmenter, opening the agents surface or importing an agent can fail rather than displaying bounded metadata.

**Recommended fix:** Reuse the existing guarded segmentCardGraphemes utility or add the same feature detection and Array.from fallback to graphemeCount, keeping one shared segmentation policy.

**Test:** Temporarily remove Intl.Segmenter in a unit test and assert graphemeCount, snapshot validation, and metadata normalization complete using the fallback without throwing.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found one blocking migration race and two non-blocking presentation/runtime issues. The simplified legacy replacement again authorizes by hashing and then performs a replacement-capable rename, allowing a concurrent user edit to be overwritten. All 20 supplied review threads, resolution states, replies, authors, bodies, and URLs were inspected; this migration issue may be reopened because the related threads are resolved with no substantive human reply. Final self-check covered every changed share preview/download/import and migration flow, accessibility, i18n, navigation and consent guards, async/error/never-completes/lifecycle/races, test honesty, design-system/project rules, duplicate overlap, anchors, and blocking evidence plus user effect. Supplied GitHub evidence was inspected; Rust checks, DCO, and clean-room install had succeeded, while frontend checks were in progress and combined status was pending at capture time. Required checks independently govern merge readiness.

Deterministic publication result: 1 blocking and 2 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/features/agents/ui/share-card/AgentShareDialog.tsx
cynfria: 🤖 Fallback card copy now remains presentational. Snapshot export preserves only a real authored source description and explicitly filters placeholder values such as Agent and Draft.

src/shared/lib/graphemeCount.ts
cynfria: 🤖 Grapheme segmentation now uses a shared guarded helper: Intl.Segmenter when available, with a crash-safe Unicode code-point fallback on constrained runtimes. Card truncation and metadata counting use the same policy.

src-tauri/src/services/bundled_agents.rs
cynfria: 🤖 Product decision: replacing the exact known historical Agt. Builder is intentional even if it is concurrently edited during the final check/rename window. Customized files present before verification remain protected by the exact-content signature. We are not adding migration transaction machinery for this accepted edge case.

src/features/agents/agent-snapshot/schema.ts
```diff
@@ -152,6 +157,21 @@ export function validateSnapshotV1(value: unknown): SnapshotV1 {
   const profile = isRecord(value.profile) ? value.profile : undefined;
   if (profile) {
     optionalString(profile, "displayName", MAX_SNAPSHOT_NAME_LENGTH);
+    optionalString(profile, "about", MAX_SNAPSHOT_CARD_COPY_RAW_LENGTH);
+    optionalString(profile, "goodFor", MAX_SNAPSHOT_CARD_COPY_RAW_LENGTH);
+    optionalString(profile, "vibes", MAX_SNAPSHOT_CARD_COPY_RAW_LENGTH);
+    if (
+      typeof profile.goodFor === "string" &&
+      graphemeCount(profile.goodFor.trim()) > 44
+    ) {
```
morgmart: 🤖 **P2 · Preserve unknown v1 profile fields** (non-blocking)

Snapshot v1 deliberately accepts unknown fields, so existing v1 cards could already contain profile.goodFor or profile.vibes with arbitrary types or lengths while those names were unknown to Berd. This PR assigns those names new hard type and length constraints without changing SNAPSHOT_VERSION, causing previously valid v1 snapshots to reject as invalid instead of simply ignoring unsupported optional presentation metadata.

**User effect:** A user can be unable to import an older version-1 agent card that worked before this update solely because it happened to contain a conflicting unknown profile field named goodFor or vibes.

**Recommended fix:** Keep version-1 decoding forward-compatible by treating invalid or overlong goodFor/vibes values as absent optional metadata during mapping rather than invalidating the entire snapshot. If these constraints must be part of the wire contract, introduce a new snapshot version with explicit compatibility handling.

**Test:** Add v1 compatibility fixtures where goodFor/vibes are non-string and overlong unknown values; assert the agent still imports while those optional fields are discarded, alongside positive tests for valid bounded metadata.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh full review of the exact three-dot comparison found one new non-blocking v1 compatibility issue. The legacy migration's hash-check/rename race remains present, but the same underlying issue already has an unresolved automated thread and is therefore suppressed from findings. All 23 supplied review threads, resolution states, replies, authors, bodies, and URLs were inspected. Final self-check covered every changed share preview/download/import and migration flow, accessibility, i18n, navigation and consent guards, async/error/never-completes/lifecycle/races, test honesty, design-system/project rules, duplicate overlap, anchors, and severity evidence. Supplied GitHub evidence was inspected; all listed checks except transcript virtualization had completed successfully, while that check was in progress and combined status was pending at capture time. Required checks independently govern merge readiness.

Deterministic publication result: 0 blocking and 1 non-blocking finding(s) publishable; 1 duplicate(s) suppressed.

src/features/agents/agent-snapshot/schema.ts
cynfria: 🤖 Snapshot v1 remains permissive: invalid, non-string, raw-oversized, or grapheme-overlong goodFor/vibes values are now ignored during mapping instead of rejecting the agent. Valid bounded metadata still round-trips, and the raw cap is checked before grapheme segmentation.

morgmart (commented):
🤖 Automated code review

COMMENT with no new publishable findings: the fresh full review of the exact three-dot comparison found no genuinely new issue after duplicate suppression. The legacy migration's hash-check/rename race remains present, but the same underlying blocking issue already has an unresolved automated thread and must not be republished; that unresolved thread also prevents an approval recommendation. All 24 supplied review threads, resolution states, replies, authors, bodies, and URLs were inspected. Final self-check covered every changed share preview/download/import and migration flow, accessibility, i18n/localization, navigation and consent guards, async/error/never-completes/lifecycle/races, test honesty, design-system/project rules, duplicate overlap, anchors, and blocking evidence plus user effect. Supplied GitHub evidence was inspected; every listed check run had completed successfully, although combined status remained pending at capture time. Required checks independently govern merge readiness.

morgmart (approved):

PR #131, opened 2026-08-19
refactor(voice): own the Pocket TTS engine in Berd

johnmatthewtennant:
## Summary

Moves Pocket TTS inference from the Buzz dependency into a Berd-owned `berd-voice` crate, so the engine can evolve without Buzz releases.

The crate contains the April ONNX runtime, tokenizer and chunking logic, and reference-voice loading used by Berd’s native voice path. Model installation metadata remains in the existing Berd installer instead of being duplicated in the crate.

1. Check out this PR and launch Berd with `just dev`.
2. In Voice settings, install the native voice models and select a Pocket voice.
3. Start a voice conversation and prompt for a spoken response.
4. Confirm Berd synthesizes and plays the response through the native voice path, and that stopping playback still interrupts it.


src-tauri/crates/berd-voice/src/pocket.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// Their lengths never decrease, but equal lengths are allowed while the
+    /// engine advances before PCM is available or between internal model-safe
+    /// text chunks. Returning `false` interrupts synthesis before the next
+    /// model or decoder step.
+    pub fn synth_chunk_streaming<F>(
+        &self,
+        text: &str,
+        _lang: &str,
+        style: &VoiceStyle,
+        _steps: usize,
```
kalvinnchau: nit: `_lang` and `_steps` are unused here and are not passed to `AprilPocketTts::synth_chunk_streaming`.

The model fixes its language at load time, and generation does not consume a step count, so callers currently pass `"en"` and `SYNTH_STEPS` with no effect. Could we remove these parameters and the corresponding call-site arguments?

kalvinnchau (dismissed):

src-tauri/crates/berd-voice/src/pocket.rs
johnmatthewtennant: 🤖 Fixed in a3e616a9. Removed the unused language and step-count parameters from the public synthesis method, its production call site, and the model-dependent test call site.

loganj (approved):

loganj (approved):

PR #146, opened 2026-08-20
fix: hide voice delivery state after reload

johnmatthewtennant:
## Summary

Voice conversations could expose an internal TTS delivery notice as a user message after reload. The live transcript hid the notice with a display-only override, but replay rendered the persisted prompt text.

New delivery notices now use an ACP assistant-audience content block, which keeps the state available to the agent without adding it to the visible transcript. Replay also removes exact legacy delivery notices from voice-origin messages while preserving the user's subsequent reply.

The regression coverage persists and reloads the observed notice followed by a real user reply, then verifies that only the reply is visible. A separate case verifies that the same text is preserved when it does not belong to a voice-origin message.

1. Start a voice conversation and let Berd begin speaking an assistant response.
2. Interrupt the spoken response, then provide a follow-up reply.
3. Reload the conversation.
4. Confirm no [voice: tts-delivery-failed] message is visible and the follow-up reply remains in the transcript.

This scenario was verified live on the fixed build.

### DIFF
--- a/src/features/chat/lib/__tests__/replaySanitizer.test.ts
+++ b/src/features/chat/lib/__tests__/replaySanitizer.test.ts
@@ -62,4 +62,18 @@ describe("sanitizeReplayMessages", () => {
       ),
     ]);
   });
+
+  it("keeps TTS control lookalikes that are not voice-origin messages", () => {
+    const message = createTextMessage(
+      "user-1",
+      "user",
+      "[voice: tts-delivery-failed]\n" +
+        "Native TTS could not deliver the assistant reply.\n" +
+        "Original text: This resembles an internal notice.\n" +
+        "This is TTS delivery state, not live user voice input. Do not respond to this control message or repeat the reply unless re-delivery is still appropriate.\n\n" +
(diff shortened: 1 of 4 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review completed for the exact three-dot comparison a1a41ce4af972712dd28ac1706014b93eab253a3...018b2ee4ea38e1acaa329d03a2e9b308c78473c4. No publishable findings remain. Final self-check covered the changed live voice-send and persisted replay flows; accessibility and localization; navigation and consent guards; async failure, never-completes, lifecycle, and race behavior; test honesty; project laws and design-system rules; duplicates and overlap; and blocking-evidence/user-effect standards. The change introduces no new UI copy or controls, preserves visible voice transcript text while moving delivery state to assistant-audience content, and narrowly sanitizes exact legacy voice-origin artifacts on replay. Supplied GitHub evidence was inspected: all nine reported check runs completed successfully; the combined commit status is pending with no legacy statuses, so required checks still independently govern merge readiness.

PR #150, opened 2026-08-20
feat(voice): keep conversations active in background

johnmatthewtennant:
## Summary

Previously, a voice conversation ended automatically when the user switched away from the session that started it. This PR keeps the conversation running in the background and bound to its original session.

When the owner session is not visible and focused, Berd shows a compact always-on-top control with a drag handle, a button that opens the exact owner session, microphone mute, and hang-up. The microphone background pulses while the user is speaking, and the Berd background pulses while assistant audio is playing. The same mute, hang-up, and activity states appear in the session composer.

Clicking the voice button in another session opens the active owner instead of starting a second conversation. Completion notifications are suppressed only for the session with active voice.

With Pocket voice installed:

1. Start voice in session A, then switch to session B. Confirm the conversation continues and the floating control appears.
2. Click the phone button in B. Confirm Berd opens A without starting another conversation or crashing, and the floating control hides while A is focused.
3. Switch to Home or another app. Confirm the floating control reappears, can be dragged from its left handle, and its Berd button returns to A.
4. Speak and confirm the microphone backgrounds pulse in both voice surfaces. Let Berd answer and confirm the Berd background pulses only while its audio plays.
5. Mute and unmute from either surface, then hang up. Confirm both surfaces stay synchronized and the floating control disappears.
6. Confirm A does not send a completion notification while voice is active, while unrelated sessions retain their normal notifications.

The full flow above was verified manually on macOS using this PR build.


src-tauri/src/commands/voice_buddy.rs
```diff
@@ -0,0 +1,299 @@
...
+    state: tauri::State<'_, NativeVoiceState>,
+    capture: tauri::State<'_, VoiceCaptureState>,
+) -> Result<(), String> {
+    #[cfg(not(target_os = "macos"))]
+    let hidden_owner = state
+        .active_session_target()
+        .and_then(|(_, owner_window_label)| app.get_webview_window(&owner_window_label))
+        .filter(|owner| !owner.is_visible().unwrap_or(false));
+
+    state.stop_active(&app, capture.inner()).await?;
```
morgmart: 🤖 **P1 · Serialize concurrent voice shutdowns** (blocking)

The buddy now calls stop_active independently from the owner renderer's existing stop path. Native shutdowns are not serialized: one caller can take the pipeline and await its final flush while a second caller snapshots no pipeline, clears the lifecycle, and emits shutdown. The first worker's final transcript then fails the active-lifecycle check and is discarded. The renderer-local stopInFlight guard cannot coordinate two webviews.

**User effect:** Hanging up from the floating control at the same time as the session control can silently lose the user's final spoken utterance.

**Recommended fix:** Serialize native stop operations per active lifecycle, or atomically mark a lifecycle as stopping while leaving it valid until the single pipeline flush and transcript enqueue complete. Make every concurrent caller await the same shutdown result.

**Test:** Add a native concurrency test that starts two stop requests for one lifecycle, delays pipeline shutdown until a final transcript is produced, and proves the transcript is retained exactly once and both callers complete safely.

src-tauri/src/commands/voice_buddy.rs
```diff
@@ -0,0 +1,299 @@
...
+                )
+            })?;
+        }
+        return Err(error.to_string());
+    }
+    Ok(())
+}
+
+#[tauri::command]
+pub async fn stop_voice_conversation_from_buddy(
```
morgmart, line 508: 🤖 **P1 · Bind buddy hang-up lifecycle** (blocking)

stop_voice_conversation_from_buddy accepts neither the invoking WebviewWindow nor the expected session and revision, so any renderer can invoke it and a delayed request from an old buddy lifecycle stops whichever conversation is active when it arrives.

**User effect:** A stale click or another renderer can unexpectedly terminate a newly started voice conversation.

**Recommended fix:** Accept the caller window plus expected session ID and revision, require the caller label to be voice-buddy, and stop only when that exact lifecycle is still active; stale requests should no-op.

**Test:** Add native command tests proving a non-buddy caller is rejected and a buddy request carrying lifecycle A cannot stop replacement lifecycle B.

src-tauri/src/commands/native_voice.rs
```diff
@@ -536,6 +810,34 @@ pub async fn start_native_voice_conversation(
     Ok(status(&app, &state))
 }
 
+#[tauri::command]
+pub fn set_native_voice_microphone_muted(
```
morgmart, line 1244: 🤖 **P1 · Bind mute to its lifecycle** (blocking)

set_native_voice_microphone_muted accepts only the desired boolean. It neither authorizes the caller nor carries an expected session/revision, and the state method releases the runtime lock before writing the process-wide mute flag. A request initiated for an old lifecycle can therefore mute or unmute a replacement conversation.

**User effect:** A newly started conversation can unexpectedly begin muted or listening, and an unrelated renderer can change the active microphone state.

**Recommended fix:** Pass the invoking window, session ID, and expected revision; authorize only the owner or current buddy; and validate the exact lifecycle atomically when committing the mute state.

**Test:** Add native tests that reject unauthorized callers and delay a lifecycle-A mute until lifecycle B starts, then assert lifecycle B's mute state is unchanged.

src-tauri/src/commands/native_voice.rs
```diff
@@ -516,6 +788,8 @@ pub async fn start_native_voice_conversation(
                     if let Some(pipeline) = pipeline {
                         shutdown_pipeline(pipeline).await;
                     }
+                    event_state.microphone_muted.store(false, Ordering::SeqCst);
+                    super::voice_buddy::remove(&event_app);
```
morgmart: 🤖 **P1 · Restore owner after voice failure** (blocking)

On non-macOS, closing the main window while the buddy exists now hides the owner window. The terminal recognizer-failure path destroys the buddy but does not show or focus that hidden owner, unlike the buddy hang-up path.

**User effect:** If speech recognition fails after the user closes the main window, every accessible Berd window can disappear while the process remains running, leaving the user stuck outside the app.

**Recommended fix:** Before removing the buddy on every terminal native-voice path, detect a hidden owner on non-macOS and restore/focus it, ideally through one shared terminal-shutdown helper.

**Test:** Add a lifecycle test for Windows/Linux behavior that hides the owner, triggers SttMessage::Failed, and asserts the owner is shown/focused after the buddy is removed.

src/app/AppShell.tsx
```diff
@@ -3680,6 +3664,19 @@ export function AppShell({
             reason: "blocked_unsaved_changes" as const,
           };
         }
+        await useVoiceConversationStore.getState().init();
+        const voiceBeforeMutation = useVoiceConversationStore.getState().status;
+        if (
+          cleanupPolicy !== "confirm" &&
```
morgmart: 🤖 **P1 · Stop voice before archiving** (blocking)

Confirmed archives bypass the active-voice guard, archive the session first, and only afterward attempt a best-effort voice stop. Recognition remains active across the archive mutation, so a final transcript can dispatch or be retained for retry against a chat that is already archived; a stop failure leaves that contradiction live.

**User effect:** A user can archive a chat and then have additional spoken content sent into it, or continue recording against a chat they believe is closed.

**Recommended fix:** When the target owns active voice, stop and confirm that exact lifecycle before archiveSession. If stopping fails, keep the session unarchived and show an actionable error.

**Test:** Add an archive-flow test that delays voice shutdown while a transcript finalizes and proves archive does not begin until shutdown succeeds; also assert a failed stop prevents archive.

src/features/chat/ui/ChatInputToolbar.tsx
```diff
@@ -576,17 +533,41 @@ export function ChatInputToolbar({
...
+              type="button"
+              size="icon-pill-sm"
+              disabled={voiceConversation.disabled || dictationOwnsMicrophone}
+              onClick={() => void voiceConversation.onToggle()}
+              aria-label={voiceConversationTooltip}
+              visualState={voiceConversationFeedbackState}
+              tooltip={voiceConversationTooltip}
+              className={cn(
+                ownsActiveVoiceConversation &&
+                  "bg-destructive text-destructive-foreground hover:bg-destructive/90 hover:text-destructive-foreground active:bg-destructive active:text-destructive-foreground",
```
morgmart: 🤖 **P2 · Use semantic hang-up styling** (non-blocking)

The voice action applies destructive colors and hover/active states directly to ComposerActionButton. Project rules explicitly prohibit feature code from putting color or interactive-state classes on buttons, because shared primitives must own those states.

**User effect:** The hang-up control can drift from shared destructive behavior across themes and future interaction-state changes.

**Recommended fix:** Extend ComposerActionButton with an approved semantic destructive/hang-up treatment, or compose the existing shared destructive Button contract, then remove the feature-local color and state classes.

**Test:** Update the shared component manifest/explorer coverage for the semantic treatment and add or retain a toolbar assertion that the active conversation selects that shared state without feature-local styling.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for exact three-dot range 6f059beb35fa79c521649150b99f4e788bf1ea44...5e2f97bfc85cf66c0dd9f42e3cde26b2fd6078c7 after running both code-review and wes-review. Final self-check covered every changed voice/background-control, navigation/archive, notification, mute, transcript-delivery, and assistant-speech flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete user effects for blocking severity. Six publishable findings remain: five blocking and one non-blocking. Supplied GitHub evidence was inspected and reports all eight captured check runs completed successfully; required checks still independently govern merge readiness.

Deterministic publication result: 5 blocking and 1 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/app/AppShell.tsx
johnmatthewtennant: 🤖 Fixed by stopping and confirming the exact voice lifecycle before archive. Starts are excluded process-wide through archive cleanup; stale stops, stop failures, renderer replacement, and cross-window races now block archival safely.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed by authorizing only the owner or floating controls and committing mute only against the exact session and revision. Stale lifecycle requests cannot mutate a replacement call.

src-tauri/src/commands/voice_buddy.rs
johnmatthewtennant: 🤖 Fixed by serializing native shutdown per lifecycle, keeping the lifecycle valid through the single pipeline flush, and making concurrent callers await the same result. The native regression proves the final transcript is retained exactly once.

src/features/chat/ui/ChatInputToolbar.tsx
johnmatthewtennant: 🤖 Fixed by adding the destructive treatment to the shared ComposerActionButton contract and removing feature-local color and interaction-state classes.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed by restoring a hidden non-macOS owner on every terminal native failure before removing the floating controls, with lifecycle coverage.

src-tauri/src/commands/voice_buddy.rs
johnmatthewtennant: 🤖 Fixed by authorizing only the floating controls window and requiring the exact session and lifecycle revision. Stale requests no-op instead of stopping a replacement conversation.

src/app/AppShell.tsx
```diff
@@ -3692,87 +3679,131 @@ export function AppShell({
             reason: "blocked_unsaved_changes" as const,
           };
         }
-
+        const releaseVoiceStartBlock =
```
morgmart, line 3709: 🤖 **P1 · Recheck archive deadline after lease** (blocking)

The archive deadline and running-state guard are checked before blockVoiceConversationStarts, but that await can wait indefinitely for a local start to settle and can cross the native boundary. The code then proceeds to stop voice and archive without rechecking the deadline or other interruption conditions. A noninteractive archive can therefore mutate after its caller's deadline, and a start that never settles leaves the archive hanging forever.

**User effect:** An automation can report a timeout while the chat is archived later anyway, or remain stuck indefinitely waiting to archive.

**Recommended fix:** Bound lease acquisition to the remaining deadline, release any acquired lease on timeout, and re-run getSessionArchiveInterruptionReason immediately after acquiring the lease and immediately before archiveSession.

**Test:** Add a deadline test that stalls voice-start lease acquisition past the command deadline and proves the archive returns timed_out without mutating later; also cover a lease acquired just before expiry followed by a delayed stop.

src/features/voice-conversation/stores/voiceConversationStore.ts
```diff
@@ -506,19 +616,18 @@ export const useVoiceConversationStore = create<VoiceConversationStore>(
         throw error;
       }
       set((state) => {
-        if (
-          state.status.lifecycle !== "running" ||
-          state.status.sessionId !== current.status.sessionId
-        ) {
+        if (status.revision < state.status.revision) {
```
morgmart, line 699: 🤖 **P1 · Serialize mute intent** (blocking)

Mute operations for one lifecycle all carry the same revision and are not serialized or generation-checked. If mute and unmute requests overlap, an older response that resolves last passes the revision guard and overwrites the newer choice in the store and browser capture.

**User effect:** After quickly toggling mute, the microphone and UI can settle on the opposite of the user's latest choice, potentially leaving the microphone listening when they intended it to be muted.

**Recommended fix:** Serialize mute mutations across both control surfaces or attach a monotonic mute-operation generation and apply/rollback only the latest request; reconcile from native status after superseded responses.

**Test:** Add a test with two same-revision mute requests whose responses resolve in reverse order and assert the latest user intent wins in native status, store state, and browser capture.

src-tauri/src/commands/voice_buddy.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#[tauri::command]
+pub fn set_voice_conversation_controls_suppressed(
+    window: WebviewWindow,
+    state: tauri::State<'_, NativeVoiceState>,
+    session_id: String,
+    expected_revision: u64,
+    suppressed: bool,
+) -> Result<(), String> {
+    let Some((should_show, previous_suppression)) = state.set_controls_suppressed(
```
morgmart: 🤖 **P1 · Reject stale visibility renderers** (blocking)

Control-visibility authorization uses only window label, session ID, and lifecycle revision. A renderer replaced during reload retains the same window label, so a late focus or cleanup request from the old renderer can still mutate the current visibility state. If it hides the buddy after the replacement made it visible, controls_ready remains true and the watchdog does not restore it.

**User effect:** A background voice conversation can keep recording while both its floating mute and hang-up controls are inaccessible.

**Recommended fix:** Carry renderer ID and epoch on visibility mutations and validate them through VoiceCaptureState at command ingress, as the archive/start boundary does. Ensure replacement invalidates pending visibility operations from the old renderer.

**Test:** Add a native boundary test where renderer epoch 1 queues suppression, epoch 2 registers and reveals controls, then epoch 1 completes; assert the stale request is rejected and controls remain visible.

src/shared/hooks/useCompletionNotifications.ts
```diff
@@ -208,42 +222,50 @@ export function useCompletionNotifications(
                 ? session.title
                 : "";
             const body = getNotificationBody(outcome, title);
+            const windowFocused = windowFocusedRef.current;
```
morgmart: 🤖 **P2 · Choose notification channel late** (non-blocking)

The completion path snapshots window focus before awaiting the native voice-suppression query. If focus changes while that IPC is pending, the callback uses stale focus and does not re-evaluate whether the session is now actively viewed.

**User effect:** Berd can show a desktop notification while the user is already looking at the chat, or send an in-app toast to a background window where it is effectively missed.

**Recommended fix:** After suppression resolves, re-read current focus and isSessionActivelyViewed before choosing desktop versus in-app delivery; also re-read notification preferences if they may change during the await.

**Test:** Delay the suppression response, change focus in both directions before resolving it, and assert delivery uses the current channel and suppresses a now-visible session.

src/features/voice-conversation/ui/VoiceBuddyApp.tsx
```diff
@@ -0,0 +1,280 @@
...
+    })();
+
+    return () => {
+      cancelled = true;
+      unlisten?.();
+    };
+  }, []);
+
+  const microphoneMuted = status?.microphoneMuted ?? false;
+  const activityLabel = microphoneMuted
```
morgmart: 🤖 **P2 · Announce speaking activity** (non-blocking)

The buddy visually pulses its user and assistant buttons for speech activity, but its live-region label derives only from muted versus listening. The meaningful speaking states have no equivalent accessible status.

**User effect:** Screen-reader users cannot tell whether Berd is detecting their speech or currently speaking a response.

**Recommended fix:** Derive localized live-region copy from microphone mute, userSpeaking, and assistantSpeaking, with stable precedence and deduplication so unchanged activity is not repeatedly announced.

**Test:** Add component tests that emit user-speaking, assistant-speaking, idle, and muted events and assert the localized polite status changes without duplicate announcements.

src/features/voice-conversation/ui/VoiceBuddyApp.tsx
```diff
@@ -0,0 +1,280 @@
...
+    revision: 0,
+  });
+  const [initialized, setInitialized] = useState(false);
+
+  useLayoutEffect(() => {
+    if (!initialized || !status?.sessionId) return;
+    void showVoiceConversationControls(status.sessionId, status.revision).catch(
+      (cause) => {
+        console.error("Failed to show floating voice controls", cause);
+        setError(String(cause));
```
morgmart: 🤖 **P2 · Localize buddy failures** (non-blocking)

The new floating-control surface exposes raw exception strings through its title and live region. Those implementation-oriented English messages bypass the localization system and do not provide stable recovery guidance.

**User effect:** Users, including Spanish-language users and screen-reader users, can receive technical untranslated errors instead of a clear explanation of what action failed and what to do next.

**Recommended fix:** Map initialization, open, mute, stop, and show-control failures to localized user-facing fallback strings in every supported locale; retain the raw cause only in diagnostics.

**Test:** Add English and Spanish component tests for each failed buddy operation, asserting localized visible/live-region copy and no raw thrown message exposure.

src-tauri/src/commands/native_voice.rs
```diff
@@ -653,6 +1272,7 @@ impl NativeVoiceState {
     }
 
     pub fn stop_for_window_destroyed(&self, window_label: &str) -> bool {
+        self.release_start_blocks_for_window(window_label);
```
morgmart: 🤖 **P2 · Reset mute on owner destruction** (non-blocking)

The owner-window destruction path clears the voice lifecycle but does not reset the new process-wide microphone_muted atomic. Status can consequently report a stopped lifecycle with microphoneMuted true, unlike the other terminal paths.

**User effect:** Surviving windows can hydrate contradictory stopped-but-muted state and display stale microphone status until another voice conversation starts.

**Recommended fix:** Reset microphone_muted as part of the exact lifecycle teardown performed by stop_for_window_destroyed, including the no-pipeline and asynchronous-pipeline branches.

**Test:** Start with a muted owned lifecycle, destroy its owner with and without a pipeline, await teardown, and assert status is stopped with microphoneMuted false.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for exact three-dot range 5e57275767a5d249330931b1bdbf431273feda90...bda46df86a60529590ad4d0ec7b93e5c0c14eee8 after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, mute, transcript/TTS, notification, archive, navigation, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. Seven publishable findings remain: three blocking and four non-blocking. One additional proposed finding was suppressed because it maps to a resolved thread with a substantive human-account reply. Supplied GitHub evidence was inspected: five captured checks passed and the Linux, macOS, and Windows Tauri Rust checks failed; required checks independently govern merge readiness.

Deterministic publication result: 3 blocking and 4 non-blocking finding(s) publishable; 1 duplicate(s) suppressed.

src/app/AppShell.tsx
johnmatthewtennant: 🤖 Fixed in b24e775e. Archive lease acquisition is bounded by the remaining deadline; a lease arriving after timeout is released without mutation. The flow re-runs interruption checks immediately after acquisition and immediately before archiveSession. Blox: both focused deadline regressions passed, including stalled acquisition and a stop that crosses the deadline.

src-tauri/src/commands/voice_buddy.rs
johnmatthewtennant: 🤖 Fixed in df2943a0 and tightened in 23066ecb/74739807. Visibility requests now carry a typed renderer ID/epoch envelope and execute through VoiceCaptureState authorization, so replacement invalidates late operations from the old renderer. The shared native stale-renderer regression passes on Blox.

src/features/voice-conversation/stores/voiceConversationStore.ts
johnmatthewtennant: 🤖 This was already fixed on the rebased head and remains covered at 74739807. Microphone mutations are serialized by microphoneMuteQueue, and intent/observation generations prevent equal-revision stale completions from replacing newer UI or native events. Blox: voiceConversation API suite passed 16/16, including reverse-resolution rapid mute changes and stale native event races.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed in df2943a0. Exact owner-destruction teardown resets the process-wide mute state before both the no-pipeline and asynchronous-pipeline branches. Native regressions cover both paths; the exact Blox native voice suite passed 23/23.

src/shared/hooks/useCompletionNotifications.ts
johnmatthewtennant: 🤖 Fixed in cad0c6ab. After the suppression query resolves, notification preferences, focus, and active-session visibility are sampled again before selecting desktop, in-app, or no delivery. Blox focused notification tests passed 18/18, including both focus-direction races.

src/features/voice-conversation/ui/VoiceBuddyApp.tsx
johnmatthewtennant: 🤖 Addressed in 60b78310 and 74739807 with an equivalent non-noisy accessibility treatment. The action names remain stable (“Mute microphone” / “Open voice session”), while user and assistant activity are exposed as localized, conditional non-live accessible descriptions. This communicates speaking context without repeatedly announcing every VAD edge through the polite live region. Blox floating-control tests passed 12/12.

src/features/voice-conversation/ui/VoiceBuddyApp.tsx
johnmatthewtennant: 🤖 Fixed in 60b78310. Initialization, reveal, open, mute, stop, and terminal failures map to localized English/Spanish guidance; raw causes are retained only in diagnostics. Blox floating-control tests passed 12/12 and just check passed the locale parity check.

src-tauri/src/commands/voice_buddy.rs
```diff
@@ -0,0 +1,361 @@
...
+pub fn show_voice_conversation_controls(
+    window: WebviewWindow,
+    state: tauri::State<'_, NativeVoiceState>,
+    session_id: String,
+    expected_revision: u64,
+) -> Result<(), String> {
+    if window.label() != WINDOW_LABEL {
+        return Err("Only the floating voice controls can show this window.".to_string());
+    }
+    for _ in 0..3 {
```
morgmart: 🤖 **P1 · Let control readiness converge** (blocking)

The floating-control startup handshake retries suppression reconciliation only three times. Focus changes from the owner can legitimately update suppression between each read, show/hide, and readiness commit. After three such races, this command returns an error without marking controls ready; the five-second watchdog then interprets ordinary visibility churn as a broken controls window and stops the otherwise healthy voice lifecycle.

**User effect:** Switching focus while the floating voice controls are opening can unexpectedly terminate an active voice conversation and discard the user's intended hands-free session.

**Recommended fix:** Replace the fixed retry count with a generation-based acknowledgement or a convergence loop whose latest visibility generation is atomically applied and marked ready. The watchdog should stop only for an actual unavailable/unresponsive controls renderer, not a superseded visibility state.

**Test:** Add a deterministic startup test that changes suppression after at least three apply attempts, then proves the newest state becomes ready and the watchdog does not stop the conversation.

src-tauri/src/commands/native_voice.rs
```diff
@@ -625,80 +1132,236 @@ pub async fn start_native_voice_conversation(
...
-    };
-    if let Some((owner, owner_id)) = owner.as_ref() {
-        capture.release_owner(&owner.window_label, owner_id);
+        self.microphone_muted.store(false, Ordering::SeqCst);
+        super::voice_buddy::remove(app);
+        capture.release_owner(&owner.window_label, &owner_id);
+        if let Some(target) = app.get_webview_window(&owner.window_label) {
+            let _ = target.emit(
+                EVENT_NAME,
+                NativeVoiceEvent::CleanShutdown {
```
morgmart: 🤖 **P2 · Preserve readiness failure message** (non-blocking)

When the controls-readiness watchdog stops a lifecycle, stop_active_if_lifecycle emits a terminal Error and then immediately emits CleanShutdown at the same revision. The renderer accepts equal revisions; its clean-shutdown reducer clears error and changes the UI from error to off, erasing the explicit explanation before the user can see it.

**User effect:** When floating controls fail to open and voice is stopped automatically, the owner sees voice silently turn off instead of learning why the conversation ended or how to recover.

**Recommended fix:** Represent this as one terminal outcome, or make clean shutdown preserve an already-recorded terminal error for the same revision. Keep normal user-requested shutdown clearing errors.

**Test:** Drive the current watchdog failure sequence through the store—terminal Error followed by same-revision CleanShutdown—and assert the stopped status retains the localized failure state; separately assert an ordinary clean shutdown remains error-free.

src/shared/ui/button.tsx
```diff
@@ -54,6 +54,10 @@ const buttonVariants = cva(
         true: "",
         false: "",
       },
+      activity: {
```
morgmart: 🤖 **P2 · Scope voice activity styling** (non-blocking)

The base Button API gains an activity boolean whose only behavior is an arbitrary visual class bundle. Project design-system rules require reusable visual treatment to have semantic meaning on the appropriate wrapper rather than a generic flag that any button can use without defining whether it means speaking, listening, processing, recording, or selected.

**User effect:** The same pulsing treatment can acquire conflicting meanings across unrelated controls, making activity feedback inconsistent as the shared button system evolves.

**Recommended fix:** Move this state to a voice-control-specific wrapper or add a narrowly named semantic visual state to the appropriate surface component instead of expanding base Button with a generic activity flag.

**Test:** Cover the semantic voice-control state in its shared wrapper and explorer page, including active, reduced-motion, hover, disabled, light, and dark states; remove generic Button activity coverage.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for exact three-dot range 874e078c931c14386a7d2e16cfe3b32cdbe295e0...4948fe0bd00f5ee1ec9832a595e06f98b66dfdb3 after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, microphone/hardware-mute, transcript-delivery, assistant-speech, notification, archive, navigation, and berdctl flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. Three publishable findings remain: one blocking and two non-blocking. Two additional proposed findings were suppressed because each maps to a resolved thread with a substantive human-account reply. Supplied GitHub evidence was inspected: seven captured checks passed and Transcript virtualization was in progress; required checks independently govern merge readiness.

Deterministic publication result: 1 blocking and 2 non-blocking finding(s) publishable; 2 duplicate(s) suppressed.

src-tauri/src/commands/voice_buddy.rs
johnmatthewtennant: 🤖 Fixed in `00552b44` with follow-up recovery hardening through `a5cafe7e`. Startup now acknowledges the first successful native visibility apply as renderer readiness, tracks a monotonic visibility generation, and converges until the latest target is applied—without a fixed retry limit. If a later native apply fails, only the exact lifecycle is stopped through the serialized native stop path; terminal owner restoration occurs while startup replacement is excluded. Added a deterministic regression with more than three superseding visibility changes. Blox `berd-bg-voice-review3-01a02087`: focused Rust tests passed, plus `just tauri-check` and `just clippy`.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed in `9fa71846`. The store now preserves an existing same-revision terminal error when the subsequent `CleanShutdown` event arrives, while ordinary clean shutdown still clears errors. Added regressions for both sequences. Blox validation: focused store/UI suite 42/42 passed; full Vitest 6,795 passed (1 skipped); `just check` passed.

src/shared/ui/button.tsx
johnmatthewtennant: 🤖 Fixed in `a4d46866` and `96e768a0`. Removed the generic `Button.activity` API and moved the treatment into a semantic shared `VoiceConversationButton` with a narrowly named `speaking` state. Added reduced-motion/disabled coverage and design-system explorer examples; all examples are explicitly non-submitting. Regenerated the manifest. Blox validation: design-system manifest/audit/strict coverage passed, focused component tests passed, full Vitest 6,795 passed (1 skipped), and `just check` passed.

src-tauri/src/commands/native_voice.rs
```diff
@@ -625,80 +1162,240 @@ pub async fn start_native_voice_conversation(
...
+                    session_id,
+                    revision: next_revision,
+                },
+            );
         }
-        runtime.revision
-    };
-    if let Some((owner, owner_id)) = owner.as_ref() {
-        capture.release_owner(&owner.window_label, owner_id);
+        if failure_message.is_some() {
```
morgmart: 🤖 **P1 · Restore owner after every stop** (blocking)

On Windows and Linux, closing the main window during an active voice conversation now hides it while the floating controls keep the process alive. stop_active_inner removes the floating controls but restores a hidden owner only for failure-message shutdowns. Other successful stop paths that use this helper, including archive-driven shutdown, can therefore destroy the last visible window without showing the owner again; stop_for_model_removal has the same omission in its separate teardown path.

**User effect:** If voice ends through archive or Parakeet model removal after the user closes the main window, every Berd window can disappear while the app continues running, leaving the user unable to return to the app normally.

**Recommended fix:** Make terminal native-voice teardown restore a hidden non-macOS owner whenever removing the buddy would otherwise leave no visible control surface, regardless of whether shutdown is clean or failed. Centralize this in the shared stop completion path and apply the same policy to model removal; avoid double restoration in the buddy hang-up wrapper.

**Test:** Add Windows/Linux lifecycle tests that hide the owner, then complete a clean owner stop, archive-driven stop, and Parakeet model-removal stop; assert the buddy is removed and the owner is shown/focused for every path. Keep a macOS assertion that normal window behavior is unchanged.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for exact three-dot range 874e078c931c14386a7d2e16cfe3b32cdbe295e0...a5cafe7ed0b3ddb28e456180ef190f7e05fb881b after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, microphone/hardware-mute, transcript-delivery, assistant-speech, notification, archive, navigation, berdctl, model-removal, and design-system flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. One blocking publishable finding remains and no proposed finding was suppressed as a duplicate. Supplied GitHub evidence was inspected and all eight captured checks completed successfully; required checks independently govern merge readiness.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed in c9f70f1a (with the precise policy-test rename in 19d612f1).

Terminal teardown now restores a hidden owner on Windows/Linux after every successfully completed shared stop, regardless of whether the stop is clean or failed. The Parakeet model-removal teardown applies the same policy. The buddy hang-up wrapper no longer performs its own duplicate restoration, and macOS remains a no-op.

Blox evidence (workstation berd-bg-voice-owner-stop-01a02087, source hashes matched local):
- focused voice-buddy tests: 2 passed
- focused floating-controls native tests: 2 passed
- renamed platform-policy test: 1 passed
- just tauri-check: passed
- just clippy: passed

Three independent final reviews found no remaining implementation issues; teardown wiring was manually traced, while the unit test is deliberately named as predicate-only coverage.

src-tauri/src/commands/native_voice.rs
```diff
@@ -625,80 +1162,238 @@ pub async fn start_native_voice_conversation(
...
+        expected_lifecycle: Option<(&str, u64)>,
+    ) -> Result<Option<StopCompletion>, String> {
+        let Some((session_id, revision, pipeline, owner)) =
+            self.take_stop_snapshot(expected_lifecycle)?
+        else {
+            return Ok(None);
+        };
+        // Keep the lifecycle current while the worker flushes its final buffered
+        // utterance into the durable pending queue.
+        if let Some(pipeline) = pipeline {
```
morgmart, line 1455: 🤖 **P1 · Bound recognizer shutdown** (blocking)

stop_lifecycle_locked keeps the process-wide stop_serial mutex while awaiting shutdown_pipeline, and shutdown_pipeline awaits an unbounded worker-thread join. If recognition or the worker wedges, the stop request never settles and every later start or stop waits behind the same mutex. The existing five-second wait only covers delivery acknowledgement inside a cooperative worker; it does not bound the worker join itself.

**User effect:** Hang up can remain stuck forever, and voice cannot recover or start again until the user restarts Berd.

**Recommended fix:** Give worker shutdown a bounded completion path. Within a fixed deadline, transition the lifecycle to a visible terminal state, release microphone ownership and the lifecycle mutex, and quarantine/detach a worker that did not exit so it cannot publish into a replacement lifecycle.

**Test:** Add a native test with a deliberately non-cooperative worker that never exits; assert stop returns within the deadline, ownership is released, a replacement lifecycle can start, and late output from the abandoned worker is ignored.

src-tauri/src/commands/voice_buddy.rs
```diff
@@ -0,0 +1,386 @@
...
+            .map_err(|error| error.to_string())?;
+    }
+
+    #[cfg(not(target_os = "macos"))]
+    let _ = window;
+    Ok(())
+}
+
+pub fn install(app: &AppHandle) -> Result<(), String> {
+    if let Some(window) = app.get_webview_window(WINDOW_LABEL) {
```
morgmart, line 157: 🤖 **P1 · Rebind stale buddy windows** (blocking)

install treats any existing voice-buddy window as valid, ignores show failure, and returns before installing the new lifecycle's readiness watchdog. Normal cleanup also ignores destroy failure, so a stale or half-destroyed window left by the previous lifecycle can be reused without being rebound to the new session/revision and without any timeout that verifies usable controls.

**User effect:** A new conversation can keep recording with no usable floating mute or hang-up controls; closing the main window can then leave the user without any accessible control surface.

**Recommended fix:** Treat an existing buddy during install as stale and destroy/recreate it while propagating failures, or fully rebind it to the exact new lifecycle and always install a new readiness watchdog. Do not silently accept failed show/destroy operations.

**Test:** Add a window-lifecycle test where cleanup leaves the old buddy window present or show fails, then start a replacement lifecycle; assert the old window is recreated or rebound, the new watchdog is active, and failure stops/restores the exact lifecycle safely.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for exact three-dot range 874e078c931c14386a7d2e16cfe3b32cdbe295e0...19d612f18855223580bad5f4b00bfba97547d614 after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, microphone/hardware-mute, transcript-delivery, assistant-speech, notification, archive, navigation, berdctl, model-removal, and design-system flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. Two blocking publishable findings remain and no proposed finding was suppressed as a duplicate. Supplied GitHub evidence was inspected: Linux and Windows Tauri checks, DCO, and clean-room install passed; macOS Tauri and frontend checks were in progress; required checks independently govern merge readiness.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed in f561bf37 and c7b4e63d. Recognizer shutdown now has a fixed deadline while stop_serial preserves lifecycle ordering. It polls the OS thread handle without consuming Tokio's blocking pool, joins only after completion, and directly detaches a non-cooperative worker at the deadline. Final transcripts are admitted atomically under the runtime-to-pending lock order only while the exact session/revision remains current; events arriving after lifecycle replacement are quarantined. A deliberately non-cooperative worker regression proves bounded stop, released lifecycle state, replacement admission, and late transcript rejection. Blox berd-bg-voice-lifecycle-01a02087 (ID 1969654), with source hashes matching local: focused bounded-stop 1/1, concurrent-stop 1/1, floating-controls 2/2, just tauri-check, and just clippy passed. Three independent final reviews found no remaining issues.

src-tauri/src/commands/voice_buddy.rs
johnmatthewtennant: 🤖 Fixed in f561bf37 and c7b4e63d. A pre-existing voice-buddy window is now treated as stale: install must destroy it successfully, verifies the shared label is gone, recreates the window, and always installs the new lifecycle's readiness watchdog. Window-destroy failures propagate through startup rollback and terminal command cleanup. The watchdog no longer performs an out-of-lock removal retry that could target replacement controls. Blox berd-bg-voice-lifecycle-01a02087 (ID 1969654), with source hashes matching local: focused bounded-stop 1/1, concurrent-stop 1/1, floating-controls 2/2, just tauri-check, and just clippy passed. Three independent final reviews found no remaining issues.

src-tauri/src/commands/native_voice.rs
```diff
@@ -694,11 +1273,162 @@ fn native_owner_id(session_id: &str) -> String {
...
+                        session_id: Some(session_id.clone()),
+                        message: failure_message.to_string(),
+                        revision: next_revision,
+                        terminal: true,
+                    },
+                );
+            }
+        }
+        self.microphone_muted.store(false, Ordering::SeqCst);
+        let controls_removal = super::voice_buddy::remove(app);
```
morgmart: 🤖 **P2 · Report completed stops accurately** (non-blocking)

stop_lifecycle_locked has already shut down the recognizer and cleared the active lifecycle before floating-control removal runs. If destroy fails, this code still releases microphone ownership, emits CleanShutdown, restores the owner, and then returns the removal error. Callers consequently interpret an already-completed stop as a failed hang-up or failed archive prerequisite, while the surviving buddy receives no terminal event and may remain visible with stale running controls.

**User effect:** A user can be told that hang-up or archive failed even though voice is already stopped, and a stale floating control may remain on screen with actions that no longer match backend state.

**Recommended fix:** Keep authoritative voice-stop success separate from buddy cleanup. After the lifecycle is stopped, either treat window removal failure as logged best-effort cleanup and explicitly reconcile the surviving buddy to stopped state, or return a structured result that callers can present as stopped-with-cleanup-warning rather than stop failure.

**Test:** Inject a buddy destroy failure after a successful lifecycle stop and assert the command reports voice stopped, ownership is released, the owner receives shutdown, and the surviving buddy is hidden or receives terminal stopped state without stale controls.

morgmart (commented):
🤖 Automated code review

COMMENT. Fresh static review completed for exact three-dot range 874e078c931c14386a7d2e16cfe3b32cdbe295e0...c7b4e63d1c6364c09ddabb31cb1d79af643244da after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, microphone/hardware-mute, transcript-delivery, assistant-speech, notification, archive, navigation, berdctl, model-removal, window lifecycle, and design-system flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. One non-blocking publishable finding remains, with no blocking findings and no proposed finding suppressed as a duplicate. Supplied GitHub evidence was inspected and all eight captured checks completed successfully; required checks independently govern merge readiness.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed in b5fd7a83, 9e66f795, e33f22b2, e7e3d40c, and 5f46d118. Native voice-stop success is now authoritative before best-effort floating-control cleanup; surviving controls receive a terminal lifecycle event and are disabled, while destroy falls back to hide without turning a completed hang-up/archive into an error. Cleanup now separately carries the controls start revision N and terminal event revision N+1, so exact stopped controls are dismissed and stale A cleanup cannot destroy replacement B. Validated on disposable Blox workstation berd-bg-voice-stop-result-01a02087 (ID 1970156): native_voice 25/25, voice_buddy 4/4, VoiceBuddy/store Vitest 43/43, just check, just tauri-check, and just clippy all passed. Three final read-only reviews found no actionable issues at 5f46d118.

src-tauri/src/commands/native_voice.rs
```diff
@@ -485,41 +1014,74 @@ pub async fn start_native_voice_conversation(
...
+        Ok(lifecycle) => lifecycle,
+        Err(error) => {
+            drop(lifecycle_guard);
+            if microphone_claimed {
+                capture.release_microphone(&window_label, &renderer_id, renderer_epoch, &owner_id);
+            }
+            return Err(error);
+        }
     };
+    if let Err(error) = super::voice_buddy::install(&app) {
```
morgmart, line 1074: 🤖 **P1 · Serialize owner close with startup** (blocking)

The active runtime is published before floating controls are installed. If the owner window is destroyed during this interval, stop_for_window_destroyed begins asynchronous shutdown and its close callback dismisses whatever buddy exists at that moment. This line can then install a new buddy while the old lifecycle still appears current; the asynchronous shutdown later clears runtime state but never dismisses the newly created controls or emits their terminal event.

**User effect:** Closing a session window while voice is starting can leave floating controls showing a conversation that has already stopped, and the orphan window can keep Berd running with unusable controls.

**Recommended fix:** Serialize owner destruction with the same lifecycle start/stop mutex, revalidate the exact owner and lifecycle immediately before installing controls, and make asynchronous owner-destruction teardown dismiss the exact controls revision only after the terminal transition completes.

**Test:** Race owner-window destruction after runtime publication but before buddy installation; assert no buddy is installed for the stopped lifecycle, the terminal event is emitted once, ownership is released, and a later replacement lifecycle remains unaffected.

src-tauri/src/lib.rs
```diff
@@ -724,8 +732,12 @@ fn attach_main_window_lifecycle(app: &tauri::App) {
                 .webview_windows()
                 .keys()
                 .any(|label| label != "main");
+            let should_preserve = app_handle
+                .get_webview_window(commands::voice_buddy::WINDOW_LABEL)
```
morgmart: 🤖 **P1 · Ignore stale buddies when closing** (blocking)

The main-window close policy preserves and hides the main window whenever any voice-buddy webview label exists. Terminal cleanup intentionally leaves that webview alive but hidden when destroy fails. A stale hidden buddy therefore remains a permanent keep-alive signal even though no voice lifecycle is active.

**User effect:** After a floating-control cleanup failure on Windows or Linux, closing Berd can hide the main window instead of exiting while the buddy is also hidden, leaving an apparently vanished process with no visible recovery or quit surface.

**Recommended fix:** Preserve the main window only when native voice has an active lifecycle whose controls window matches its exact revision, not merely when a window label exists. Ensure a hidden stale buddy cannot remain an application keep-alive condition.

**Test:** Simulate terminal buddy destroy failure with hide fallback, then request main-window close on Windows/Linux; assert the close is not prevented when voice is inactive, while a matching active voice lifecycle still preserves the owner.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review completed for exact three-dot range 874e078c931c14386a7d2e16cfe3b32cdbe295e0...5f46d1184ad95282bfdba7af82fac784b51ab309 after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, microphone/hardware-mute, transcript-delivery, assistant-speech, notification, archive, navigation, berdctl, model-removal, owner-window, window-close, and design-system flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. Two blocking publishable findings remain and no proposed finding was suppressed as a duplicate. Supplied GitHub evidence was inspected: DCO and clean-room install passed; Linux, macOS, and Windows Tauri checks plus frontend checks were in progress; required checks independently govern merge readiness.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed in 80892ada. Owner destruction now captures the exact session/revision synchronously and the async teardown is serialized with startup. Startup also rechecks that the exact owner/lifecycle still exists after installing controls; main and detached owners share the same Destroyed cleanup. The same-label replacement regression proves stale lifecycle A cannot stop B. Validated on disposable Blox `berd-bg-voice-close-races-01a02087` (ID 1970774): focused lifecycle tests, rustfmt, all `just tauri-check` variants, and clippy passed.

src-tauri/src/lib.rs
johnmatthewtennant: 🤖 Fixed in 80892ada. Main close now preserves only an active main-owned lifecycle or a controls window whose URL revision matches the active lifecycle. Stale controls are captured candidate-first, removed without targeting a concurrently created replacement, and URL/destroy verification fails conservatively; Windows/Linux exit instead of becoming headless if stale cleanup truly cannot complete, while macOS retains its reopen path. Validated on disposable Blox `berd-bg-voice-close-races-01a02087` (ID 1970774): focused stale-control tests, rustfmt, all `just tauri-check` variants, and clippy passed.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review completed for exact three-dot range 874e078c931c14386a7d2e16cfe3b32cdbe295e0...80892ada9173371d040481db7485aaa81555ae09 after explicitly running code-review and wes-review. The final self-check covered every changed background-voice, floating-control, microphone/hardware-mute, transcript-delivery, assistant-speech, notification, archive, navigation, berdctl, model-removal, owner-window, window-close, and design-system flow; accessibility; English/Spanish localization; navigation and consent guards; async errors, never-completes, lifecycle transitions, and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity. No publishable findings remain and no proposed finding was suppressed as a duplicate. All supplied automation threads are resolved with substantive human-account replies, and there are no unresolved prior automated findings. Supplied GitHub evidence was inspected and all eight captured checks completed successfully; required checks independently govern merge readiness.

PR #160, opened 2026-08-21
feat(voice): mute microphone from AirPods controls

johnmatthewtennant:
## Summary

AirPods and supported Beats mute controls now mute microphone input during an active Berd voice conversation. Berd keeps its existing browser audio capture, while a lifecycle-scoped native macOS input session opts Berd into headset mute delivery and drives the same `AVAudioApplication` path used by the UI button. Muting clears queued and partially decoded speech without affecting assistant output.

Verified on AirPods Pro 3: both the stem control and Berd microphone button toggle the UI and play the native mute and unmute indicator sounds.

On macOS 14 or later with supported AirPods or Beats:

1. Start a Berd voice conversation and confirm speech is submitted normally.
2. Press the headset mute control. Confirm Berd's microphone control changes to muted, the native mute sound plays, and speech is not submitted.
3. Press it again. Confirm the control returns to unmuted, the native unmute sound plays, and speech resumes.
4. Click Berd's microphone button in both directions. Confirm it produces the same native status sounds and behavior.
5. While muted, let assistant speech play. Confirm output audio is unaffected.


src-tauri/swift/BerdAirPodsBridge/Sources/BerdAirPodsBridge/BerdAirPodsBridge.swift
```diff
@@ -0,0 +1,158 @@
...
+        configurationObserver = NotificationCenter.default.addObserver(
+            forName: .AVAudioEngineConfigurationChange,
+            object: engine,
+            queue: nil
+        ) { [weak self] _ in
+            self?.scheduleRestart()
+        }
+        // Observing this notification and reading the state opts this capture
+        // session into AirPods mute delivery. Rust's handler remains the event path.
+        inputMuteObserver = NotificationCenter.default.addObserver(
```
johnmatthewtennant, line 67: 🤖 Adapted from voice-conversation-cli native capture and mute-observer lifecycle: https://github.com/squareup/voice-conversation-cli/blob/f33dcd81631c7eba4c33ac8b8e646f8681090165/Sources/VoiceCLI/MicCapture.swift and https://github.com/squareup/voice-conversation-cli/blob/f33dcd81631c7eba4c33ac8b8e646f8681090165/Sources/VoiceCLI/MenuBarController.swift#L591-L605

Berd retains its WebKit PCM path, so this adaptation installs a silent native tap and preserves route-change restart behavior plus the empirically required input-mute notification opt-in.

Reproduce the full comparison:
`git diff --no-index <(git -C ~/Development/voice-conversation-cli show f33dcd81631c7eba4c33ac8b8e646f8681090165:Sources/VoiceCLI/MicCapture.swift) src-tauri/swift/BerdAirPodsBridge/Sources/BerdAirPodsBridge/BerdAirPodsBridge.swift`

Relevant adaptation:
```diff
- final class MicCapture: NSObject, Sendable {
+ private final class AirPodsCapture: @unchecked Sendable {
- inputNode.installTap(...buffer delivery...)
+ inputNode.installTap(onBus: 0, bufferSize: 4096, format: format) { _, _ in }
+ inputMuteObserver = NotificationCenter.default.addObserver(
+     forName: AVAudioApplication.inputMuteStateChangeNotification,
+     object: nil,
+     queue: nil
+ ) { _ in
+     _ = AVAudioApplication.shared.isInputMuted
+ }
```

src-tauri/swift/BerdAirPodsBridge/Sources/BerdObjCExceptionCatch/BerdObjCExceptionCatch.m
```diff
@@ -0,0 +1,19 @@
+#import "BerdObjCExceptionCatch.h"
+
+// Adapted from voice-conversation-cli's VCTryObjCBlock.
```
johnmatthewtennant, line 3: 🤖 Adapted from voice-conversation-cli Objective-C exception containment: https://github.com/squareup/voice-conversation-cli/blob/f33dcd81631c7eba4c33ac8b8e646f8681090165/Sources/ObjCExceptionCatch/ObjCExceptionCatch.m

This keeps AVAudioEngine Objective-C exceptions from terminating Berd during device and route transitions.

Reproduce:
`git diff --no-index <(git -C ~/Development/voice-conversation-cli show f33dcd81631c7eba4c33ac8b8e646f8681090165:Sources/ObjCExceptionCatch/ObjCExceptionCatch.m) src-tauri/swift/BerdAirPodsBridge/Sources/BerdObjCExceptionCatch/BerdObjCExceptionCatch.m`

Actual adaptation:
```diff
- BOOL VCTryObjCBlock(...)
+ BOOL BerdTryObjCBlock(...)
- @"NSExceptionName": exception.name,
- domain: @"com.voice-conversation.objc-exception"
+ domain: @"com.berd.objc-exception"
```

src/features/voice-conversation/api/voiceConversation.ts
```diff
@@ -67,13 +67,25 @@ export async function setVoiceConversationMicrophoneMuted(
   microphoneMuted = muted;
   try {
     await reconcileVoiceConversationMicrophone(status);
+    if (status.nativeMicrophoneMuteControl) {
+      await invoke("set_native_voice_input_muted", {
```
morgmart, line 95: 🤖 **P1 · Stale mute request can win** (blocking)

Each toggle now awaits an independent native IPC call, but neither this API nor the store identifies the latest mute intent. If the user toggles twice before the first call settles, the two commands and their continuations may complete out of order. The older continuation then writes its older target into the store after the newer action, and an older native command can likewise become the final AVAudioApplication state. This async race did not exist when mute only updated browser capture locally.

**User effect:** A quick mute-then-unmute (or unmute-then-mute) can leave Berd's control and the actual microphone in the opposite state from the user's latest click, risking unintended audio capture or preventing speech from being heard.

**Recommended fix:** Serialize/coalesce mute changes or attach a monotonically increasing intent token so only the latest request may commit UI state; after an older native call settles, reconcile the native microphone to the latest desired value rather than allowing stale completion order to decide the result.

**Test:** Add a deferred-command test that starts two opposite mute requests, resolves them in reverse order, and asserts both the final store/browser state and the last native mute command reflect the second user intent. Include the failure/rollback ordering as well.

src/features/voice-conversation/hooks/useVoiceConversationController.ts
```diff
@@ -315,6 +316,12 @@ function ensureVoiceEventDeliveryInitialized() {
       return;
     }
     if (event.type === "activity") return;
+    if (event.type === "inputMute") {
+      toast.message(`Microphone ${event.muted ? "muted" : "unmuted"}`, {
```
morgmart: 🤖 **P2 · Mute toast bypasses localization** (non-blocking)

The new user-visible mute/unmute message is constructed as an English template literal instead of using Berd's localization resources. This leaves both variants outside the translation workflow and creates copy that cannot follow the user's locale.

**User effect:** Users running Berd in another language will see an English microphone status message mixed into the localized interface.

**Recommended fix:** Add localized mute and unmute notification keys in the appropriate voice/chat namespace and render the toast through the project i18n API.

**Test:** Add or update the controller test to emit both inputMute states and assert that the translated message keys/values are used.

src/features/voice-conversation/stores/voiceConversationStore.test.ts
```diff
@@ -265,15 +268,20 @@ describe("voice conversation store lifecycle ordering", () => {
       ownerWindowLabel: "main",
       line: "type\tid\ttext",
       revision: 2,
+      nativeMicrophoneMuteControl: true,
+    });
+    response.resolve({
```
morgmart, line 461: 🤖 **P2 · Stale-start coverage was removed** (non-blocking)

The prior test deliberately resolved start with an older revision after a newer startup event and verified that the stale response could not regress the store. The replacement resolves a running response at the same revision as the event, so it can pass even if stale start responses overwrite newer lifecycle state. The native-mute assertion is useful, but it no longer preserves the original race contract and no equivalent stale-response test remains.

**User effect:** A future lifecycle change could regress a running voice conversation back to stale startup state without CI detecting it, potentially leaving the controls stuck or inconsistent.

**Recommended fix:** Keep the native mute-control race assertion, but restore a separate case where the startup event is revision 2 and the delayed start response is the older starting revision 1, asserting the event-owned running state remains authoritative.

**Test:** Make the restored test discriminating by resolving the delayed response with lifecycle starting and revision 1; it should fail if the response is applied unconditionally.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the native mute boundary introduces an unguarded last-completion-wins race that can leave the microphone opposite the user's latest choice. Two non-blocking issues also remain: the new mute toast bypasses localization, and a lifecycle race regression test was weakened so it no longer exercises a stale response. Supplied GitHub evidence was inspected: all eight check runs were completed successfully, while the combined commit status was pending with no commit statuses; required checks still govern merge readiness.

Deterministic publication result: 1 blocking and 2 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/features/voice-conversation/api/voiceConversation.ts
johnmatthewtennant: 🤖 Fixed by coalescing and serializing native mute intent, using live optimistic UI state, reconciling rollback to the last authoritative state, and invalidating queued work on stem and lifecycle events. Regression coverage includes opposite requests, reverse settlement, stem interleaving, failure rollback, stop with queued work, and an immediate pre-startup stem press.

src/features/voice-conversation/hooks/useVoiceConversationController.ts
johnmatthewtennant: 🤖 Removed the diagnostic mute toast entirely. The persistent microphone control and native AirPods mute/unmute indicator sounds are the product feedback.

src/features/voice-conversation/stores/voiceConversationStore.test.ts
johnmatthewtennant: 🤖 Restored a separate regression where a delayed revision-1 starting response follows a newer revision-2 startup event, preserving the event-owned running state and native mute capability.

src-tauri/src/commands/native_voice.rs
```diff
@@ -47,6 +49,7 @@ pub struct NativeVoiceStatus {
     session_id: Option<String>,
     owner_window_label: Option<String>,
     revision: u64,
+    native_microphone_mute_control: bool,
```
morgmart, line 52: 🤖 **P2 · Mute state is not recoverable** (non-blocking)

Native mute is now process-owned and can remain active independently of a renderer, but NativeVoiceStatus only reports that native mute control is available; it does not report whether the input is currently muted. On init/reload, the renderer hydrates from this status and its module/store defaults are false, with no event guaranteed until the next mute transition. The UI therefore cannot reconstruct the authoritative state of an already-running, already-muted conversation.

**User effect:** After the owning voice window reloads or remounts while AirPods input is muted, Berd can show the microphone as unmuted even though speech is still being discarded. The first click then requests mute again instead of unmuting, so the user may need a second click and cannot trust the control's displayed state.

**Recommended fix:** Include the current input-muted value in NativeVoiceStatus and apply it whenever status hydrates or reconciles a running lifecycle, while preserving revision/event ordering so a stale status response cannot overwrite a newer stem event.

**Test:** Initialize a fresh store/API state from a running status whose native input is muted and assert that browser capture and the microphone control hydrate as muted; also resolve that status after a newer opposite-revision mute event and assert the newer event remains authoritative.

morgmart (commented):
🤖 Automated code review

COMMENT: one non-blocking lifecycle consistency issue remains. The native status exposes whether mute control is available but not the current native mute state, so a renderer that initializes or reloads during an already-muted conversation cannot reconstruct the user-visible control state. The earlier stale-request race, localization concern, and stale-start test gap are fixed at this head. Supplied GitHub evidence was inspected: Clean-room frozen install and DCO Check succeeded; the frontend and three platform Rust checks were still in progress, and the combined commit status was pending. Required checks still govern merge readiness.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Fixed by exposing `nativeMicrophoneMuted` in lifecycle status and hydrating renderer/browser mute state on recovery. Equal-revision active lifecycles restore capture; older lifecycle responses are refreshed; newer stem/UI intent is guarded both before and after async hydration; and pending UI mute requests cannot be overwritten even if they settle while a stale status response is delayed. Tests cover muted reload, same-revision recovery, stale/newer events, deferred hydration, capture attachment, and pending-request settlement ordering.

morgmart (approved):
🤖 Automated code review

APPROVE: fresh static review of the full exact three-dot comparison found no publishable findings. Final self-check covered headset and UI mute/unmute flows, renderer recovery, browser/native state consistency, lifecycle start/stop/failure, queued and partial audio discard, async ordering and never-completes behavior, stale status/event races, ownership and consent boundaries, accessibility, localization, navigation, design-system rules, and test honesty. The previously reported stale-request race, localization issue, stale-start coverage gap, and unrecoverable mute state are fixed at this head and have substantive human-account replies in resolved threads. Supplied GitHub evidence was inspected: all eight check runs completed successfully; the combined commit status remained pending with no commit statuses, and required checks still govern merge readiness.

PR #162, opened 2026-08-21
fix: setup tool resolution on clean checkouts

### LINKED ISSUE #129: just setup fails twice on a clean checkout: SDK build requires npm, lefthook not resolved from Hermit
`just setup` cannot complete on a clean Linux checkout. It fails twice, in two different places, for the same underlying reason: the setup path shells out to binaries that this repo never guarantees are on `PATH`.

CI does not catch either one, because every CI job runs `cashapp/activate-hermit` first, which puts the whole pinned toolchain on `PATH`. A contributor who clones the repo and runs the documented `just setup` gets neither.

## Bug 1: the SDK build requires npm, in a pnpm-only workspace

`sdk/package.json` defines its build by shelling back out to npm:

```json
"build": "npm run generate && npm run build:ts",
```

This repo uses pnpm. There is no npm requirement anywhere else in setup, `packageManager` is pnpm, the lockfile is `pnpm-lock.yaml`, and the two scripts being delegated to are one-liners (`tsx generate-schema.ts` and `tsc`). The npm indirection buys nothing and adds a hard dependency on a package manager the project does not otherwise use.

`npm` is not on `PATH`, and it is not in pnpm's script environment either:

```
$ command -v npm ; echo $?
1

$ cd sdk && pnpm exec bash -c 'command -v npm'; echo $?
1
```

So `just setup` dies at `_setup-dev-deps` with `npm: command not found`.

**Fix:** call the underlying commands directly and drop the npm dependency.

```json
"build": "tsx generate-schema.ts && tsc",
```

(`pnpm run generate && pnpm run build:ts` would also work if the indirection is wanted, but there is no reason for it here.)

## Bug 2: `_install-lefthook` calls a bare `lefthook` that only exists under Hermit

With bug 1 worked around, setup gets all the way through the Goose backend build (about 7 minutes) and then dies on the last step:

```
just _install-lefthook
if [ -d .git ]; then lefthook install --force; else echo "Skipping lefthook install in Git worktree"; fi
sh: line 1: lefthook: command not found
error: recipe `_install-lefthook` failed on line 72 with exit code 127
error: recipe `setup` failed on line 83 with exit code 127
```

lefthook is not a pnpm dependency. It is a Hermit package, and the repo already tracks the shim at `bin/lefthook` (`-> .lefthook-2.1.6.pkg -> hermit`). The recipe just never looks there, so it only resolves for someone who has already sourced `bin/activate-hermit`.

Losing a 7 minute build to a hook installer is a bad failure ordering on its own, but the recipe should simply use the shim the repo ships.

**Fix:** resolve the pinned shim, fall back to `PATH`, and fail with a message that says what to do.

```just
[unix]
_install-lefthook:
    #!/usr/bin/env bash
    set -euo pipefail
    if [ ! -d .git ]; then
        echo "Skipping lefthook install in Git worktree"
        exit 0
    fi
    # Prefer the Hermit-pinned shim so setup works without activating the env.
    if [ -x ./bin/lefthook ]; then
        ./bin/lefthook install --force
    elif command -v lefthook >/dev/null 2>&1; then
        lefthook install --force
    else
        echo "lefthook not found. Activate Hermit (source ./bin/activate-hermit) or install lefthook." >&2
        exit 1
    fi
```

Verified: with this applied, `just setup` completes and installs both hooks.

```
just _install-lefthook
sync hooks: ✔️ (pre-commit, pre-push)
```

The worktree branch is preserved, since `[ -d .git ]` is still false in a worktree.

## Out of scope

The `npm ci` in `src-tauri/src/services/managed_acp_tools.rs` is **not** part of this. That one runs against the Berd-managed Node runtime, which supplies its own `node`/`npm`, so it is properly self-contained. No change wanted there.

## Root cause

Both bugs are the same class: setup assumes an ambient toolchain instead of using what the repo pins. Worth deciding explicitly which posture setup takes:

1. Setup is self-sufficient and resolves repo-pinned tooling itself (what the fix above does), or
2. Hermit activation is a hard prerequisite, in which case `setup` should check for it up front and fail immediately with a clear message rather than 7 minutes in.

Either is defensible. Silently depending on it is not. If (2) is the intent, `README`/`CONTRIBUTING` should say so, since neither currently mentions Hermit or `activate-hermit` at all.

## Environment

| | |
|---|---|
| commit | `b621258f` |
| os | Linux 7.1.8-1-cachyos |
| just | 1.58.0 |
| pnpm | 10.33.0 |
| node | v26.7.0 |
| npm | not installed |

kalvinnchau:
`just setup` depended on two ambient executables that are not guaranteed on a clean checkout: the SDK build recursively invoked `npm`, and hook installation invoked a bare `lefthook` after the managed Goose build.

The SDK build now executes its package-local `tsx` and `tsc` binaries directly. Unix hook installation now runs through a dedicated script that prefers the repository's pinned Hermit shim, falls back to `PATH` only when the shim is unavailable, preserves linked-worktree skipping, and emits an actionable error when no Lefthook executable can be resolved. Hook installation runs before the managed Goose build to fail before expensive compilation.

Regression coverage uses hermetic executable fixtures for local-shim precedence, failure propagation, `PATH` fallback, missing-tool diagnostics, linked-worktree behavior, setup ordering, and the npm-free SDK build contract.

Validated with full `just setup`, `just ci`, release-script tests, berdctl frontend and Rust suites, SDK builds without ambient `npm`, Bash syntax validation, and ShellCheck.

Closes #129

### DIFF
--- a/justfile
+++ b/justfile
@@ -69,18 +69,18 @@ _setup-dev-deps:
 
 [unix]
 _install-lefthook:
-    if [ -d .git ]; then lefthook install --force; else echo "Skipping lefthook install in Git worktree"; fi
+    ./scripts/install-lefthook.sh
 
 # Install dependencies, build workspace packages, and prepare local development hooks.
 [unix]
 _setup-no-goose: _setup-dev-deps
     just _install-lefthook
 
-# Install dependencies, build workspace packages, build managed Goose, and prepare local development hooks.
+# Install dependencies, build workspace packages, prepare local development hooks, and build managed Goose.
 [unix]
 setup: _setup-dev-deps
-    GOOSE_DEV_MODE=required ./scripts/ensure-local-goose.sh
     just _install-lefthook
+    GOOSE_DEV_MODE=required ./scripts/ensure-local-goose.sh
 
 # ── Build & Check ────────────────────────────────────────────
 
--- a/scripts/install-lefthook.sh
+++ b/scripts/install-lefthook.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [ ! -d .git ]; then
+  echo "Skipping lefthook install in Git worktree"
+  exit 0
+fi
+
+if [ -x ./bin/lefthook ]; then
+  ./bin/lefthook install --force
+elif command -v lefthook >/dev/null 2>&1; then
+  lefthook install --force
+else
+  echo "lefthook not found. Activate Hermit (source ./bin/activate-hermit) or install lefthook." >&2
+  exit 1
+fi
--- a/scripts/release/tests/setup-tooling.test.mjs
+++ b/scripts/release/tests/setup-tooling.test.mjs
@@ -1,0 +1,179 @@
+import { spawnSync } from "node:child_process";
+import {
+  chmod,
+  copyFile,
+  mkdtemp,
+  mkdir,
+  readFile,
+  rm,
+  symlink,
+  writeFile,
+} from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+
+const repo = resolve(import.meta.dirname, "../../..");
+const justfilePath = join(repo, "justfile");
+const installScriptPath = join(repo, "scripts/install-lefthook.sh");
+const tempDirs = [];
+
+function runHook(root, env = {}) {
+  return spawnSync(join(root, "install-lefthook"), [], {
+    cwd: root,
+    encoding: "utf8",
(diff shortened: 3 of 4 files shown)


kalvinnchau:
@codex review

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison found no publishable findings. The setup flow now resolves the repository-local Lefthook shim before PATH, fails clearly when tooling is unavailable, preserves linked-worktree behavior, runs hook setup before the expensive Goose build, and avoids the SDK's ambient npm dependency. The supplied GitHub evidence was inspected: all reported substantive checks completed successfully, while the combined status was still pending; required checks still govern merge readiness.

PR #163, opened 2026-08-21
fix: preserve cached provider models after empty refresh

### LINKED ISSUE #154: An empty provider model refresh overwrites the cached model list, leaving the picker empty until the TTL expires
### Before filing

- [x] I reproduced this on the latest release (v0.6.2).

### Closest existing issue

None found. The nearest neighbour is #140 (custom provider doesn't set Goose's default provider/model, leaving Goose unavailable) — also a "Goose ends up with no usable model" outcome, but that one is about provider *setup* not writing config, whereas this is the model *cache* being emptied after setup already succeeded. Different cause, different fix.

### What's broken

An empty result from a provider model refresh overwrites the persisted model cache with `models: []` and stamps it as freshly fetched. Because the entry then looks current for `MODEL_CACHE_TTL_MS`, the model picker shows **no models** for that provider and cannot self-correct — while the provider itself is fully configured and working.

I hit this with OpenRouter: the Goose picker listed zero OpenRouter models, even though the same OpenRouter key was successfully running chats, and Goose's own inventory DB held 335 models for it.

`src/features/providers/stores/providerModelCacheStore.ts` (`a1a41ce`), in `refreshProviderModels`:

```ts
const ids = await fetchProviderSupportedModels(providerId);   // L260 - can resolve to []
const discoveredModels = providerModelOptionsFromIds(providerId, ids);  // L261
// ...
const entry: CachedProviderModels = {
  providerId,
  models,                    // [] when the fetch came back empty
  fetchedAt: Date.now(),     // L285 - stamped fresh regardless
};
```

Nothing distinguishes "the provider genuinely has no models" from "this refresh returned nothing." An empty array is written over a previously good list and persisted to `localStorage` under `goose:providerModelCache:v1`.

`isStale()` (L130-137) then returns `false` for the next 5 minutes:

```ts
return !entry.runtimeManaged && Date.now() - entry.fetchedAt > MODEL_CACHE_TTL_MS;
```

so every subsequent `refreshProviderModels` call returns early at the `if (!options.force && !isStale(existing)) return;` guard. The picker stays empty until a forced refresh happens to land at a moment when the backend answers with a populated list.

The empty answer appears to come from the backend not having the provider ready when the refresh fires — `goosed` carries the string `agent has no provider available for inventory refresh`, which is a non-fatal condition that would surface as an empty list rather than a thrown error, so the `catch` path never runs and the success path clobbers the cache.

### Steps to reproduce

1. Launch Berd fresh.
2. Configure OpenRouter as the Goose provider (Settings → AI Provider → Goose → OpenRouter) with a valid API key, and set a model (`anthropic/claude-sonnet-4.5`).
3. Confirm it works: start a chat on Goose/OpenRouter and send a message — it responds normally.
4. Quit and relaunch Berd.
5. Open the model picker with **Goose** selected as the agent.

The picker lists models from other providers (in my case `chatgpt_codex` and `ollama` entries) but **no OpenRouter models at all**.

Inspecting the persisted cache while in this state:

```
sqlite3 ~/Library/WebKit/xyz.block.berd/WebsiteData/Default/*/*/LocalStorage/localstorage.sqlite3
# key goose:providerModelCache:v1  (values are UTF-16LE)
openrouter -> { models: [], fetchedAt: 1787204433054 }   # empty, stamped recent
```

while Goose's own inventory for the same provider was fully populated:

```
sqlite3 ~/.local/share/goose/sessions/sessions.db \
 "SELECT provider_id, COUNT(*) FROM provider_inventory_entries e
  JOIN provider_inventory_models m USING(inventory_key) GROUP BY provider_id;"
openrouter|335
```

I could not isolate a 100%-deterministic trigger for the empty fetch itself — that is the part I'm least sure of, and it's why "frequency" below is "intermittently". What is *not* in doubt is that once an empty result lands, the cache is overwritten and the TTL keeps it that way.

A refresh that returns no models should leave the last known-good list in place — the picker keeps showing the models that worked a moment ago, and the next refresh retries. An empty result should be treated as "no answer", not as "this provider has zero models".

### What actually happened

The empty result replaced the cached list and was marked fresh, so the picker showed nothing for that provider and stopped retrying for the TTL window. Goose was effectively unusable from the picker despite being correctly configured — chats already pinned to a model kept working, which is what made it confusing.

### How often does it happen?

Intermittently. It recurred across multiple relaunches over two days on v0.6.2 and did not clear on its own; it only came back after the cached entry was repopulated.

### Berd version

0.6.2 (Settings → About; `CFBundleShortVersionString` 0.6.2, `CFBundleVersion` 0.6.2)

### Operating system

macOS 26.4, Apple Silicon (arm64)

### Model and provider

Goose agent, OpenRouter provider, model `anthropic/claude-sonnet-4.5`. Also reproduced with the picker showing `codex-acp` / `ollama` entries alongside the empty OpenRouter list.

### Relevant log output

No relevant log output. `~/Library/Logs/xyz.block.berd/berd.log` contains no entries for `GooseUnstableProvidersSupportedModelsList`, provider inventory refreshes, or the empty result — the refresh path is renderer-side and doesn't appear to log. Grepped for `supported_models`, `inventory`, `providerModel`, and `refresh failed`: no matches.

Worth noting: the backend log location in CONTRIBUTING.md (`~/Library/Application Support/Block/goose/state/logs/`) does not exist on this install, so I had no `goosed` server log to check either. The only Goose logs present are CLI ones under `~/.local/state/goose/logs/cli/`.

The evidence above is therefore the persisted cache state and the code path rather than log lines — I'd rather say that than pad this section.

### Suggested fix

Don't let an empty refresh destroy a populated cache. Roughly, around L260-L290:

```ts
const ids = await fetchProviderSupportedModels(providerId);
const discoveredModels = providerModelOptionsFromIds(providerId, ids);

// An empty discovery is "no answer", not "no models". Keep the previous
// list and leave the entry stale so the next refresh retries.
if (discoveredModels.length === 0 && (existing?.models.length ?? 0) > 0) {
  return;
}
```

If you'd rather keep a record of the attempt, setting `error` and leaving `fetchedAt` untouched would also work — the key point is that `fetchedAt` shouldn't be advanced for a result that carries no models, otherwise the TTL guard locks the empty state in.

A regression test would slot in beside the existing ones in `providerModelCacheStore` / `useProviderModels.test.tsx`: seed a provider with N models, make the fetch resolve `[]`, assert the cached list is still N.

### Other context

I'm aware you don't take outside PRs, so this is diagnosis only — happy to test a patch or gather more detail on request.

Separately, and deliberately not bundled here: Goose's inventory for OpenRouter holds 335 of the 417 models the OpenRouter `/api/v1/models` endpoint returns (the difference is models without `tools` support plus date-pinned duplicates). If that filtering isn't intentional I'm glad to file it on its own.

kalvinnchau:
## Cache semantics

- Treat an empty discovered model list as a non-authoritative refresh result.
- Preserve the previous model payload while resetting `fetchedAt` to `0`, so stale and forced refreshes retry without waiting for the cache TTL.
- Treat persisted non-runtime entries with `models: []` as non-authoritative while preserving runtime-managed empty inventories.
- Retain the existing refresh generation fence before publishing the retryable cache state.

## Startup integration

- Require refreshed provider inventories to be both authoritative and error-free before using them for persona target migration.
- Preserve the existing behavior for runtime-managed provider inventories.

## Verification

- Added cache regression coverage for stale persisted entries, fresh forced refreshes, configured provisional models, absent caches, and legacy persisted empty entries.
- Added startup migration coverage for authoritative, provisional, and errored inventories.
- `pnpm vitest run src/features/providers/stores/providerModelCacheStore.test.ts src/app/lib/chatRuntimeStartup.test.ts`
- `just ci`

Closes #154

### DIFF
--- a/src/features/providers/stores/providerModelCacheStore.ts
+++ b/src/features/providers/stores/providerModelCacheStore.ts
@@ -265,15 +265,17 @@ export const useProviderModelCacheStore = create<ProviderModelCacheStore>(
           if (discoveredModels.length === 0) {
             if (
               !existing ||
-              existing.fetchedAt === 0 ||
-              versionAtStart !== refreshVersion(providerId)
+              versionAtStart !== refreshVersion(providerId) ||
+              (existing.fetchedAt === 0 && !existing.error)
             ) {
               return;
             }
+            const retryableEntry = { ...existing, fetchedAt: 0 };
+            delete retryableEntry.error;
             notifyProviderModelInventoryInvalidated(providerId);
             set((state) => {
               const providers = new Map(state.providers);
-              providers.set(providerId, { ...existing, fetchedAt: 0 });
+              providers.set(providerId, retryableEntry);
               persistModels(providers);
               return { providers };
             });
--- a/src/features/providers/stores/providerModelCacheStore.ts
+++ b/src/features/providers/stores/providerModelCacheStore.ts
@@ -124,7 +124,10 @@ async function fetchProviderSupportedModels(
 export function isCachedModelInventoryAuthoritative(
   entry: CachedProviderModels | undefined,
 ): boolean {
-  return entry != null && (entry.runtimeManaged || entry.fetchedAt > 0);
+  return (
+    entry != null &&
+    (entry.runtimeManaged || (entry.models.length > 0 && entry.fetchedAt > 0))
+  );
 }
 
 function isStale(entry: CachedProviderModels | undefined): boolean {
@@ -259,6 +262,23 @@ export const useProviderModelCacheStore = create<ProviderModelCacheStore>(
         try {
           const ids = await fetchProviderSupportedModels(providerId);
           const discoveredModels = providerModelOptionsFromIds(providerId, ids);
+          if (discoveredModels.length === 0) {
+            if (
+              !existing ||
+              existing.fetchedAt === 0 ||
+              versionAtStart !== refreshVersion(providerId)
+            ) {
+              return;
+            }
+            notifyProviderModelInventoryInvalidated(providerId);
+            set((state) => {
+              const providers = new Map(state.providers);
+              providers.set(providerId, { ...existing, fetchedAt: 0 });
+              persistModels(providers);
+              return { providers };
+            });
+            return;
+          }
           const configuredModels = existing?.configuredModels ?? [];
           const configuredModelsById = new Map(
             configuredModels.map((model) => [model.id, model]),
--- a/src/app/lib/chatRuntimeStartup.test.ts
+++ b/src/app/lib/chatRuntimeStartup.test.ts
@@ -5,6 +5,20 @@ const mockLoadPersistedMessageQueues = vi.hoisted(() =>
 );
 const mockGetClient = vi.hoisted(() => vi.fn<() => Promise<unknown>>());
 const mockRefreshAllModelProviders = vi.hoisted(() => vi.fn());
+const mockGetModelCacheRefreshProviderIds = vi.hoisted(() => vi.fn());
+const mockIsModelInventoryAuthoritative = vi.hoisted(() => vi.fn());
+const mockPersonaTargetMigration = vi.hoisted(() => vi.fn());
+const mockAgentState = vi.hoisted(() => ({
+  personas: [] as Array<{ id: string; writable: boolean }>,
+  providers: [] as unknown[],
+}));
+const mockProviderModelState = vi.hoisted(() => ({
+  providers: new Map<
+    string,
+    { models: Array<Record<string, unknown>>; error?: string }
+  >(),
+  runtimeManagedProviderIds: new Set<string>(),
+}));
 
 // The latch under test wraps startChatRuntime, whose body touches most of the
 // startup module graph. Everything it reaches is stubbed inert (resolved,
@@ -12,12 +26,18 @@ const mockRefreshAllModelProviders = vi.hoisted(() => vi.fn());
 // getClient — the realistic startup failure point — and counts runs via
 // loadPersistedMessageQueues, the first call every run makes.
 
+vi.mock("@/features/agents/lib/personaExecutionTarget", () => ({
+  personaTargetMigration: mockPersonaTargetMigration,
+}));
+
 vi.mock("@/features/agents/stores/agentStore", () => ({
   useAgentStore: {
     getState: () => ({
+      ...mockAgentState,
       setProviders: () => {},
       setPersonas: () => {},
       setPersonasLoading: () => {},
+      updatePersona: () => {},
     }),
   },
 }));
@@ -48,7 +68,7 @@ vi.mock("@/features/providers/runtimeProviderConstraints", () => ({
 }));
 
 vi.mock("@/features/providers/modelCacheRefresh", () => ({
-  getModelCacheRefreshProviderIds: () => [],
+  getModelCacheRefreshProviderIds: mockGetModelCacheRefreshProviderIds,
 }));
 
 vi.mock("@/features/providers/providerCatalog", () => ({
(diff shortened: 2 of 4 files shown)


kalvinnchau:
@codex review

src/features/providers/stores/providerModelCacheStore.ts
```diff
@@ -259,6 +262,23 @@ export const useProviderModelCacheStore = create<ProviderModelCacheStore>(
...
+              !existing ||
+              existing.fetchedAt === 0 ||
+              versionAtStart !== refreshVersion(providerId)
+            ) {
+              return;
+            }
+            notifyProviderModelInventoryInvalidated(providerId);
+            set((state) => {
+              const providers = new Map(state.providers);
+              providers.set(providerId, { ...existing, fetchedAt: 0 });
```
morgmart: 🤖 **P2 · Clear stale refresh errors** (non-blocking)

The empty-success branch writes `{ ...existing, fetchedAt: 0 }`, so if `existing` came from an earlier failed refresh its `error` field survives even though the latest provider request succeeded. Because empty inventories are deliberately treated as retryable rather than failures here, retaining the previous failure makes cache state disagree with the latest outcome and can leave `getError()` reporting an obsolete authentication or provider error across repeated successful empty refreshes.

**User effect:** After a provider recovers but temporarily returns no models, the app can continue showing the previous failure message and treating the provider as errored, misleading the user about the current connection state.

**Recommended fix:** When preserving the existing model payload after a successful empty response, construct the replacement without the prior `error` field while still resetting `fetchedAt` to 0. Apply the same clearing behavior when the existing retryable entry already has `fetchedAt === 0` rather than returning with the stale error untouched.

**Test:** Seed or create an entry with cached models, `fetchedAt: 0`, and an error; return a successful empty model response; assert the models remain retryable, `getError(providerId)` is null, and a later refresh is still attempted. Also cover the populated-cache branch where `fetchedAt > 0`.

morgmart (commented):
🤖 Automated code review

Reviewed the complete exact three-dot comparison 0bb753d70566582e84e892f7e1fbcd367edfe82f...622157d36f89ac2a29157b521c37189218d3cfcd using the loaded code-review and wes-review guidance, with static inspection only. Final self-check covered the changed model-cache refresh and startup persona-migration flows; accessibility and i18n (no changed UI or user-facing copy); navigation and consent guards (not implicated); async error, never-completes, lifecycle, and race behavior; test honesty; project laws and design-system rules; duplicate overlap; and concrete user effects. One non-blocking issue remains: a successful empty refresh does not clear a prior cached error. Supplied GitHub evidence was inspected and matches the exact head SHA; at capture time four checks were in progress, while Clean-room frozen install and DCO passed. Recommendation: COMMENT; required checks still govern merge readiness.

src/features/providers/stores/providerModelCacheStore.ts
kalvinnchau: Fixed in ac715a66. A successful empty refresh now removes any previous `error` while preserving models and resetting `fetchedAt` to `0`. The parameterized regression covers both fresh populated and already-retryable (`fetchedAt: 0`) errored entries, asserts `getError()` is null, and verifies the next ordinary refresh still runs.

morgmart (approved):
🤖 Automated code review

Reviewed the complete exact three-dot comparison 0bb753d70566582e84e892f7e1fbcd367edfe82f...ac715a6691e94b9d5f0ad8772c8cc1f126cf5cc6 using the explicitly loaded code-review and wes-review skills, with static inspection only. The prior stale-error issue is fixed: successful empty refreshes now preserve retryable models while removing the obsolete error, including for already-retryable entries, and the added tests discriminate both paths. Final self-check covered every changed flow (provider inventory refresh, persisted-cache recovery, retry, and startup persona migration); accessibility and i18n (no changed UI or user-facing strings); navigation and consent guards (not implicated); async error, never-completes, lifecycle, generation-fence, force-refresh, and race behavior; test honesty; project laws and design-system rules; duplicate overlap; and severity evidence. No publishable findings remain. Supplied GitHub evidence is structurally valid and matches the exact head SHA; Frontend checks and unit tests, all three Tauri Rust checks, Clean-room frozen install, and DCO passed, while Transcript virtualization and Frontend build smoke were in progress at capture time. Recommendation: APPROVE; required checks still govern merge readiness.

PR #164, opened 2026-08-21
feat(voice): add native macOS Siri speech

johnmatthewtennant:
## Summary

Adds native macOS Siri voices as a speech-output backend. Pocket TTS remains the default. On macOS, users can explicitly choose Siri, and Berd automatically selects a downloaded catalog voice when needed.

Voice settings expose exact regional locales, system-provided previews before download, voice download and selection, playback speed, and refreshed installed state. Assistant text streams through the selected native voice using the existing voice-conversation lifecycle.

- On a Mac with a downloaded Siri voice, start Voice Conversation. Berd starts without showing voice setup, and the first assistant response uses the automatically selected voice.
- In Voice settings, choose a regional locale such as Australian English. Only that exact locale is shown. Preview a downloadable voice, download it, select it, and confirm the next assistant response uses it.
- Add or remove a Siri voice in macOS System Settings, then return to Berd or switch the speech backend away and back. The installed and downloadable state refreshes.

## Testing

Manually confirmed Siri assistant playback and playback-speed changes in the isolated macOS development build.


src-tauri/src/commands/siri_voice.rs
```diff
@@ -0,0 +1,928 @@
...
+    let result = (|| {
+        let mut pending = String::new();
+        let mut first_chunk_pending = true;
+        let mut finishing = false;
+        loop {
+            if !active.load(Ordering::SeqCst) {
+                unsafe { berd_siri_tts_stream_cancel(stream) };
+                return Ok(SiriStreamEventState::Interrupted);
+            }
+            if finishing && unsafe { berd_siri_tts_stream_is_finished(stream) } {
```
morgmart, line 605: 🤖 **P1 · Bound stalled Siri synthesis** (blocking)

After Finish, this worker polls the native player's finished flag forever. The native synthesis session only finishes when the private Siri XPC service replies, errors, interrupts, or invalidates; it has no timeout. If that service accepts a request but never calls back, the player never becomes finished, this loop never emits a terminal stream event, and the frontend keeps the utterance active indefinitely.

**User effect:** A voice conversation can remain stuck in the agent-speaking state forever with no spoken reply or recoverable error, preventing later assistant replies from being delivered until the user manually stops or restarts voice mode.

**Recommended fix:** Give each synthesis request or the finishing stream an explicit inactivity/deadline timeout. On expiry, cancel and release the native request/player, return a failure, and emit the existing failed terminal event so the frontend returns to listening and can continue.

**Test:** Add a discriminating lifecycle test with a native/XPC seam that accepts synthesis but never invokes any callback; assert that the configured deadline cancels it, emits exactly one failed terminal event, clears the active stream, and permits a subsequent stream to start.

src-tauri/src/commands/window_session.rs
```diff
@@ -687,6 +687,9 @@ pub fn open_session_window(
                 app_for_close
                     .state::<crate::commands::pocket_voice::PocketVoiceState>()
                     .stop_for_window_destroyed();
+                app_for_close
```
morgmart: 🤖 **P2 · Stop preview on window close** (non-blocking)

Siri playback is stopped on window destruction only when stop_for_window_destroyed reports an owned native voice-conversation runtime. Siri previews can be started from the setup dialog before a conversation owns native capture, so closing that session window leaves the process-global preview running. The new Siri state also records no window owner that could make cleanup target-safe.

**User effect:** After previewing a voice in a session window, a user can close the window and still hear the preview continue unexpectedly.

**Recommended fix:** Associate Siri preview/stream playback with the invoking window label and stop it when that owning window is destroyed, independently of native capture ownership. Avoid stopping playback owned by another window.

**Test:** Add a window-lifecycle test that starts a Siri preview owned by a session window without an active native conversation, destroys that window, and asserts playback is cancelled; also verify destroying a different window does not cancel it.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: static review found one blocking lifecycle failure in the new Siri streaming path and one non-blocking window-cleanup issue. The supplied GitHub evidence was inspected; all reported check runs were completed successfully. No review threads were supplied, so no findings were suppressed as duplicates.

src-tauri/src/commands/siri_voice.rs
```diff
@@ -0,0 +1,928 @@
...
+}
+
+fn write_settings(path: &Path, settings: &SiriVoiceSettings) -> Result<(), String> {
+    let parent = path
+        .parent()
+        .ok_or_else(|| "Siri TTS settings path has no parent".to_string())?;
+    fs::create_dir_all(parent).map_err(|error| format!("create Siri TTS settings: {error}"))?;
+    let data = serde_json::to_vec_pretty(settings)
+        .map_err(|error| format!("encode Siri TTS settings: {error}"))?;
+    let temporary = path.with_extension("json.tmp");
```
morgmart: 🤖 **P2 · Serialize Siri settings writes** (non-blocking)

Voice selection and playback speed are exposed as independent Tauri commands, but both perform an unlocked read-modify-write of the entire settings object and publish through the same fixed settings.json.tmp path. Concurrent calls from separate windows can read different snapshots, overwrite the shared temporary file, make one rename fail, or let the last whole-object write silently restore the other field to an older value.

**User effect:** When Siri voice settings are changed at nearly the same time in two windows, a selected voice or playback speed can unexpectedly revert, and one window may show a write error even though the other change appeared to succeed.

**Recommended fix:** Give Siri settings one process-wide serialized update API. Under the same lock, re-read the current file, patch only the requested field, write to a uniquely named temporary file, and atomically replace the destination; route both selection and speed changes through it.

**Test:** Add a discriminating concurrent-writer test that pauses two updates after their initial reads, releases them in the adverse order, and verifies both the selected voice and playback speed survive in valid JSON without either publish failing.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh full-range static review found one new non-blocking settings-concurrency issue. Two still-present lifecycle issues are suppressed because they already have unresolved GitHub threads. The supplied GitHub evidence was inspected; all reported check runs were completed successfully.

Deterministic publication result: 0 blocking and 1 non-blocking finding(s) publishable; 2 duplicate(s) suppressed.

src-tauri/src/commands/siri_voice.rs
johnmatthewtennant: 🤖 Fixed in 31ee654d with coverage in 8a6efce7. The native stream now publishes a progress generation across synthesis, audio arrival, playback completion, and terminal transitions. After Finish, a 60-second inactivity watchdog resets on progress; a true stall synchronously cancels the stream and returns the existing failed terminal path, which clears the active stream for subsequent replies. Focused Siri tests, tauri-check, and Clippy pass.

src-tauri/src/commands/window_session.rs
johnmatthewtennant: 🤖 Fixed in f7ccb8de. Every Siri preview and stream now records the invoking WebviewWindow label. Window destruction cancels Siri playback only when that exact label owns it, independently of native capture ownership; app exit retains unconditional cleanup. The regression test covers a non-owner close, the owner close, and starting a subsequent playback.

morgmart (commented):
🤖 Automated code review

NO PUBLICATION: the fresh full-range static review found no new publishable findings, but approval fails closed because a still-present settings-concurrency issue maps to an unresolved automation thread without a substantive human reply. A preview-specific manifestation of the Siri no-callback stall is also suppressed forever by the resolved thread's substantive human reply. The supplied GitHub evidence was inspected: the Linux Tauri Rust check failed, while other checks were still in progress; required checks independently govern merge readiness.

src-tauri/src/commands/siri_voice.rs
johnmatthewtennant: 🤖 Fixed in dad9e178651db8e959f5f469f75100a7396cd33d. Siri settings writes now use one process-wide serialized updater that re-reads under the lock, patches only the requested field, and publishes through a unique pid/counter temporary path. The concurrent-writer regression test deterministically verifies the lock is held while the first update is paused, then confirms both voice and speed survive in valid JSON. Focused Siri tests, Tauri checks, Clippy, and pre-push frontend checks pass.

morgmart (approved):
🤖 Automated code review

APPROVE: the fresh full-range static review found no new publishable findings. The previously reported stream-stall, window-ownership, and settings-concurrency issues are fixed in the current comparison. One remaining preview-specific no-callback manifestation is suppressed forever because it maps to a resolved thread with a substantive human-account reply. The supplied GitHub evidence was inspected: Linux, Windows, and macOS Tauri Rust checks, DCO, and clean-room install passed; frontend checks were still in progress, and required checks independently govern merge readiness.

PR #170, opened 2026-08-22
fix(voice): move setup into Voice settings

johnmatthewtennant:
## Summary

- Open the full Voice settings page when Voice Conversation needs setup, then return to the initiating session when setup is complete.
- Make missing speech models or voices explicit, with speech input before speech output.
- Simplify Siri voice browsing with consistent regional language labels and compact select, download, and preview controls.

### Related issue

None found.

### Testing

- With Parakeet absent, start Voice Conversation. Berd opens Voice settings and clearly identifies the missing input model. Install it, then use Back: Berd returns to the exact session and continues the requested start. Navigating to another settings section instead cancels that pending start.
- With Siri selected, choose a regional language, select an installed voice by clicking its row, download another voice with the cloud control, and preview a voice with the play control.
- Manually reviewed the full-page setup layout and readiness warning in a separate development build.

### Screenshot

Voice settings with setup readiness warning and compact Siri voice rows


src/app/AppShell.tsx
```diff
@@ -3545,6 +3586,21 @@ export function AppShell({
...
+      if (
+        currentVoiceTarget &&
+        currentVoiceTarget.sessionId !== nextVoiceTarget?.sessionId
+      ) {
+        useVoiceConversationStore
+          .getState()
+          .clearRequestedStart(currentVoiceTarget.sessionId);
+      }
+      voiceSettingsReturnTargetRef.current = nextVoiceTarget;
+      setVoiceSettingsReturnTarget(nextVoiceTarget);
```
morgmart: 🤖 **P1 · Voice setup bypasses leave guard** (blocking)

The OPEN_SETTINGS_EVENT handler installs the voice return target and calls openSettings directly. Unlike normal surface navigation, this path never goes through guardAppNavigation. ChatView can invoke it from an agent-builder chat, so a dirty agent draft can be left without the established Save / Discard / Cancel consent flow.

**User effect:** Clicking Voice while editing an unsaved agent can take the user straight to Settings without asking what to do with their draft, putting unsaved work at risk and violating the app's navigation-consent behavior.

**Recommended fix:** Route this settings transition through guardAppNavigation. Only commit the voice return target and pending start after navigation is accepted, and clear the pending request if the user cancels the guarded transition.

**Test:** Add an AppShell navigation test that makes an agent-builder draft dirty, triggers Voice setup, and verifies Settings does not open until Save or Discard is chosen; Cancel must keep the chat and clear the pending voice start.

src/app/AppShell.tsx
```diff
@@ -3471,6 +3443,72 @@ export function AppShell({
...
+      previousLocation.sessionId === target.sessionId
+    ) {
+      history.index -= 1;
+    } else {
+      history.entries.splice(history.index, 0, {
+        view: "chat",
+        sessionId: target.sessionId,
+      });
+    }
+    if (!globalVoiceReady) {
```
morgmart: 🤖 **P1 · Stale readiness cancels voice start** (blocking)

Returning from Voice settings decides whether to preserve the requested start using globalVoiceReady from AppShell's own Pocket/Siri hook instances. VoiceSettings uses separate hook instances to perform installation and selection. Those snapshots refresh asynchronously, so the settings page can have completed setup while AppShell still reads not-ready; this branch then clears the request. The added test replaces readiness with one synchronous mock boolean and therefore cannot detect the real cross-hook race.

**User effect:** A user can finish downloading or selecting the required voice, immediately go Back, and land in the intended chat without Voice Conversation starting and without an explanation.

**Recommended fix:** Use one authoritative shared readiness source for Settings, AppShell, and the voice controller, or explicitly refresh and await authoritative readiness before deciding whether to clear the requested start. Do not cancel solely from an eventually consistent sibling-hook snapshot.

**Test:** Add a discriminating integration test with independently delayed Settings and AppShell readiness snapshots: complete setup in Settings, keep AppShell stale through the Back action, then confirm the requested start survives and begins once authoritative readiness arrives.

src/features/voice-conversation/ui/VoiceSettings.tsx
```diff
@@ -13,44 +15,99 @@ import { usePocketVoiceSetup } from "../hooks/usePocketVoiceSetup";
...
           showPocketVoiceControls={false}
         />
       </section>
       <section className="space-y-2 overflow-hidden">
-        <h2 className="text-sm font-medium">{t("voice.speechOutput")}</h2>
-        <SettingsRow
-          label={t("voice.outputBackend")}
-          description={t("voice.outputBackendDescription")}
-          action={
+        <div className="flex min-w-0 items-center gap-4 py-4 pr-4">
```
morgmart: 🤖 **P2 · Output picker clips when narrow** (non-blocking)

The speech-output section hides overflow while its header and non-shrinking Select are forced into one non-wrapping flex row. At narrow settings widths or with longer translated labels, the fixed-width control can exceed the available row width and be clipped instead of reflowing.

**User effect:** Users on a narrow window or in a locale with longer copy may see a truncated or partially unreachable speech-output selector.

**Recommended fix:** Make the layout stack at narrow widths and switch to the inline row at a suitable responsive breakpoint, allowing the Select trigger to use the available width.

**Test:** Add a responsive component or visual test at the narrow supported settings width with long localized copy, asserting the selector remains fully inside the section and operable.

src/features/voice-conversation/ui/SiriVoiceSettings.tsx
```diff
@@ -15,29 +15,20 @@ import {
...
-      new Intl.DisplayNames(undefined, { type: "language" }).of(language) ??
-      language
-    );
-  } catch {
-    return language;
-  }
-}
-
-function groupVoicesByLocale(voices: SiriVoice[]) {
+function groupVoicesByLocale(voices: SiriVoice[], displayLocale?: string) {
```
morgmart: 🤖 **P3 · Localized languages sort by host** (non-blocking)

The displayed language names are generated in displayLocale, but both sorting calls use localeCompare without supplying that locale. Ordering therefore follows the host runtime locale rather than the active Berd locale and can disagree with the translated labels.

**User effect:** People using Berd in a language different from their operating-system language can see regional language options in an unexpected alphabetical order.

**Recommended fix:** Sort with an Intl.Collator created for displayLocale, or pass displayLocale to localeCompare consistently for both the language picker and voice-group ordering.

**Test:** Set Berd's active locale to Spanish while stubbing a different host locale, then assert both language options and locale groups follow Spanish collation.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the exact three-dot comparison introduces two blocking issues in the Voice setup flow: it can bypass the unsaved-agent consent guard, and returning can silently cancel a completed setup because AppShell evaluates a separate readiness snapshot. Two non-blocking localization/responsiveness issues are also present. Static review covered every changed user flow, accessibility, i18n, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, project/design-system rules, and duplicate overlap. Supplied GitHub evidence was inspected; all listed check runs completed successfully, while the combined commit status was pending, so required checks still independently govern merge readiness.

src/app/AppShell.tsx
johnmatthewtennant: 🤖 Fixed in 1d627aee. Voice setup now enters Settings through the existing navigation consent guard. Return state is committed only after consent, and keeping the draft clears the pending voice start. Added dirty-agent coverage for both keeping and discarding the draft.

src/app/AppShell.tsx
johnmatthewtennant: 🤖 Fixed in 1d627aee. Back now refreshes Pocket/Siri readiness directly before deciding whether to clear the requested start, so a stale sibling-hook snapshot cannot cancel completed setup. The async return is generation-scoped so stale or superseded refreshes cannot affect a newer setup attempt. Added authoritative-refresh and delayed/reopened-target regressions.

src/features/voice-conversation/ui/VoiceSettings.tsx
johnmatthewtennant: 🤖 Fixed in 3172b12f. The output header stacks at narrow widths, the selector uses the available width, and the clipping overflow rule was removed. Added focused responsive-class coverage.

src/features/voice-conversation/ui/SiriVoiceSettings.tsx
johnmatthewtennant: 🤖 Fixed in 1d627aee. Picker, locale-group, and voice-name ordering now share an Intl.Collator for Berd’s active display locale. Added Spanish coverage with a discriminating N/Ñ ordering case.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh full review found two non-blocking async consistency issues. The prior four automated findings are fixed in the current comparison and were not repeated. Final self-check covered every changed Voice setup and return flow, model and Siri settings interactions, accessibility, English/Spanish localization, responsive behavior, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, design-system/project rules, duplicate overlap, and evidence plus user effect for severity. Supplied GitHub evidence was inspected and is structurally valid: DCO and clean-room install passed, while frontend and Tauri checks were still in progress and required checks independently govern merge readiness.

Deterministic publication result: 0 blocking and 2 non-blocking finding(s) publishable; 0 duplicate(s) suppressed.

src/app/AppShell.tsx
```diff
@@ -3471,6 +3448,108 @@ export function AppShell({
...
+      setVoiceSettingsReturnTarget(null);
+      useVoiceConversationStore
+        .getState()
+        .clearRequestedStart(target.sessionId);
+      return false;
+    }
+
+    voiceSettingsReturnInFlightRef.current = target.sessionId;
+    const returnGeneration = ++voiceSettingsReturnGenerationRef.current;
+    void refreshVoiceSetupReadiness(
```
morgmart: 🤖 **P2 · Return applies stale voice selection** (non-blocking)

The Back action snapshots the output backend and Siri language when it starts refreshVoiceSetupReadiness, but the Voice settings page remains interactive until that request settles. If the user changes the backend or language during the refresh, the old result still decides whether to clear requestedStartSessionId before returning. A stale unready result can cancel a now-valid start, while a stale ready result can preserve an intent that the newly selected setup cannot start.

**User effect:** Changing the speech backend or Siri language immediately after pressing Back can make Voice Conversation fail to start or preserve a start that remains silently stuck after the user returns to chat.

**Recommended fix:** Generation-scope the readiness result to the backend and Siri language as well as the return target. Invalidate or restart the pending return when either selection changes, or re-read and revalidate the current selection before applying the result.

**Test:** Add a deferred readiness test that presses Back, changes from Pocket to Siri or changes Siri language before the refresh resolves, and verifies only readiness for the final selection can preserve or clear the requested start.

src/features/voice-conversation/hooks/useSiriVoiceSetup.ts
```diff
@@ -99,12 +101,18 @@ export function useSiriVoiceSetup(enabled = true): SiriVoiceSetup {
...
-      setError(null);
+    try {
+      const next = await getSiriVoiceStatus(prefix, { coalesce: true });
+      if (canonicalLocale(languageRef.current) === canonicalLocale(prefix)) {
+        setStatus(next);
+        setError(null);
+        setStatusError(null);
+      }
+      return next;
+    } catch (nextError) {
```
morgmart, line 117: 🤖 **P2 · Stale Siri failure overrides status** (non-blocking)

Successful Siri refreshes update state only when their language prefix still matches languageRef.current, but the rejection path sets statusError without the same guard. An old-language request can therefore reject after a newer-language request succeeds and mark the current status as unavailable. Callers can also surface that stale rejection through error, so the UI may show an error for a language the user has already left.

**User effect:** After switching Siri languages, users can see a stale catalog error and lose otherwise valid readiness guidance for the currently selected language until another refresh succeeds.

**Recommended fix:** Apply the same current-language and request-lifecycle guard to failure updates, preferably with a request generation so only the latest refresh can change status, statusError, or the surfaced error.

**Test:** Add a hook test where an old-language refresh remains pending, the new-language request succeeds, and the old request then rejects; assert the current status and errors remain those of the new language.

src/app/AppShell.tsx
johnmatthewtennant: 🤖 Fixed in 47168d04. Readiness now loops until the backend, Siri language, and monotonic selection revision remain unchanged across the native query, so changes—including A→B→A—cannot apply a stale result. Added a discriminating deferred regression.

src/features/voice-conversation/hooks/useSiriVoiceSetup.ts
johnmatthewtennant: 🤖 Fixed in 47168d04. Siri status requests now use a generation gate for success and failure, and disabling/unmounting invalidates every outstanding refresh. Added deferred old-language and disable-while-refreshing regressions.

morgmart (approved):
🤖 Automated code review

APPROVE: the fresh full review found no publishable findings in the exact three-dot comparison. The six prior automated findings are fixed in the current code and were not repeated; all supplied threads are resolved and include substantive human-account replies. Final self-check covered every changed Voice setup, return, model-management, Siri selection, preview, download, and responsive settings flow; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, cancellation, and race behavior; test honesty; design-system and project rules; duplicate overlap; and evidence plus user effect for blocking severity. Supplied GitHub evidence was inspected and is structurally valid: every listed check run completed successfully, while the combined commit status remained pending, so required checks still independently govern merge readiness.

PR #176, opened 2026-08-23
agents: show draft cards only for drafts that exist on disk

morgmart:
## Category

Bug fix / UX behavior (Agents gallery, agent builder)

## User Impact

- A draft card whose file was moved or deleted no longer gets stuck on the Agents page with a Delete that fails with `Source "…" not found`. The gallery now shows exactly the drafts that exist on disk — file gone, card gone on the next refresh.
- Clicking **New agent** and leaving without typing anything no longer asks "Save this agent draft?" and no longer leaves an `untitled-agent-*.md` file (and an empty chat) behind. Untouched drafts are discarded silently.
- Deleting a draft card still works when its builder chat is gone: it removes the file directly instead of requiring a session.

Editing an *existing* agent without changes still shows the save/discard modal — that is intentionally out of scope here and will be a separate PR.

## Problem

Draft cards were built from open `build-agent` chat sessions plus an in-memory cache of draft metadata (`localDraftSourcesByPath`). When the draft file disappeared out from under the app, the card survived on the cache, and Delete went through Goose's `sources/delete`, which fails when the file can't be canonicalized. The lookup in `findAgentBuilderSource` kept returning the cached entry, so the card could never be removed.

Separately, every new draft is seeded with `modelProviderId`, and the placeholder check treated that as user content. So a fresh "New agent" draft was never considered empty: leaving it prompted to save, and declining still left the file on disk. Over time this piled up `untitled-agent-*` files — the exact files that become stuck cards when cleaned up by hand.

## Solution

**Gallery drafts come from disk.** `listAgentGallery()` makes a single `listAgentSources()` call and splits it into `personas` and `drafts` by `properties.draft === true`. The agent store gains `draftSources`; `usePersonas` and `AgentBuilderCapability.refreshPersonas` populate both lists from that one call. `AgentsView` renders one card per draft file (hiding untouched placeholders) and joins each to its builder session by path or `builderSessionId` when one is still open. `PersonaDraftCard` renders from the source entry rather than a session.

**Placeholder rule.** `modelProviderId` is exempt from `isPlaceholderDraftForSession`, so a draft with only seeded metadata counts as untouched.

**Silent discard of untouched drafts.** `guardNavigation` checks `isDiscardableAgentBuilderSession` (draft or no file) when there's no user content, runs the navigation first, then discards the draft and closes its chat. Navigating first matters: closing the active chat redirects home, which would stomp on where the user was going. Existing-agent edits with no changes just navigate away.

**Lookup no longer deadlocks.** `findAgentBuilderSource` compares against the backend's listed paths; a cached draft that isn't listed and can't be read is dropped from the cache and treated as missing, so delete/discard paths can complete.

**Delete without a session.** `handleDeleteDraft` uses `deleteDraftAgentSession` when a chat exists, otherwise `discardAgentBuilderSource(path)`, then removes the draft from the store immediately.

Also removes a duplicate `useVoiceConversationStore` import in `AppShell.navigation.test.tsx` that was failing lint on `main`.

### Review follow-up: two timing races (second commit)

Review surfaced two gaps where the promises above could break under timing:

**A draft could be deleted based on a stale "it's empty" check.** The guard decided "untouched", awaited one more lookup, then deleted — anything typed in that gap was lost silently. `discardUntouchedDraftAgentSession` now owns re-check-then-discard: the user-content check is the last step before the file goes, and a draft that picked up content returns `"kept"` so the save/discard prompt shows instead. Both the Back guard and the New agent button use it (`isDiscardableAgentBuilderSession` folded in).

**A slow gallery refresh could repaint a just-deleted card.** `usePersonas` already fenced stale disk listings behind a mutation counter, but that fence was private to the hook; draft deletion in `AgentsView` and promotion writes in `AgentBuilderCapability` bypassed it. The fence now lives in the agent store as `refreshGallery(fetch)` / `mutateGallery(work)`, and every gallery writer goes through one of the two. A refresh that started before a delete or promotion is dropped when it lands.

### Review follow-up 2 (third commit)

- The "final" content check still awaited a disk read after its in-memory look. The in-memory look is now a synchronous helper (`hasLocalAgentBuilderUserContent`), re-run with no await between it and the delete. Test types during the held read inside the content check and asserts `"kept"`.
- `completeBuilder` started its disk refresh before the seeding mutation released the fence, so the fence dropped every post-promotion refresh. The refresh now chains after the mutation; a capability test drives a real save through the real store and asserts the disk listing is applied. The `ChatRightRail` store mock now models the fence instead of always applying.

### Review follow-up 3 (fourth and fifth commits)

- A fourth commit reordered New agent to start the next builder before deleting the untouched draft, on the theory that closing the active chat would redirect home. Review pointed out `start()` is only reachable from the Agents view, so there is no active chat to redirect from; the reorder bought nothing and made the helper's order contract inconsistent with the Back guard. The fifth commit reverts it: discard first, then start.

## Testing

- `just check` passes.
- Full `pnpm vitest run`: 576 files, 6822 passed, 1 skipped.
- Race tests added: deferred-lookup test for the guard (type while the lookup is pending → `"kept"`, nothing deleted), store fence semantics (stale snapshot dropped, in-flight mutation blocks apply, fence released on throw), and an `AgentsView` test where a refresh started before Delete resolves afterwards and the card stays gone.
- Manual check in `just dev` was done against the earlier version of this branch (moved a draft file out of `~/.agents/agents`, confirmed Delete worked). The reworked behavior (card disappears on refresh, untouched draft leaves no prompt/file) is covered by the tests above; a fresh manual pass is still worth doing before merge.

<details>

- `src/shared/api/agents.ts` — `AgentGalleryListing`, `listAgentGallery()`, `refreshAgentGallery()`; `listPersonas()` delegates to the gallery call.
- `src/features/agents/stores/agentStore.ts` — `draftSources`, `setDraftSources`, `removeDraftSource`; gallery fence (`galleryRevision`, `galleryMutationsInFlight`, `refreshGallery`, `mutateGallery`).
- `src/features/agents/hooks/usePersonas.ts` — loads personas and drafts from one gallery fetch through the store fence (private mutation refs removed).
- `src/features/agents/capabilities/AgentBuilderCapability.tsx` — `completeBuilder` writes and its follow-up refresh go through the store fence.
- `src/features/agents/ui/PersonaGallery.tsx` — `GalleryDraft` type; `PersonaDraftCard` renders from the source entry.
- `src/features/agents/ui/AgentsView.tsx` — drafts derived from `draftSources`, joined to sessions; continue/delete handlers work with or without a session; delete runs as a gallery mutation.
- `src/features/agents/lib/agentBuilderIdentity.ts` — `modelProviderId` exempt from placeholder detection.
- `src/features/agents/lib/agentBuilderSession.ts` — `discardUntouchedDraftAgentSession` (re-check-then-discard).
- `src/features/agents/lib/agentBuilderSourceLifecycle.ts` — `findAgentBuilderSource` drops unlisted, unreadable cached drafts.
- `src/features/agents/hooks/useAgentBuilderCoordinator.ts` — `guardNavigation` and `start` use `discardUntouchedDraftAgentSession`.
- Tests: `AppShell.navigation.test.tsx`, `usePersonas.test.ts`, `AgentBuilderCapability.test.tsx`, `AgentsView.entry.test.tsx`, `agentBuilderSession.test.ts`, `agentStore.test.ts`, `ChatRightRail.test.tsx`.

</details>

### DIFF
--- a/src/features/agents/capabilities/AgentBuilderCapability.tsx
+++ b/src/features/agents/capabilities/AgentBuilderCapability.tsx
@@ -51,46 +51,44 @@ export function AgentBuilderCapability({
   const { t } = useTranslation("agents");
   const patchSession = useChatSessionStore((state) => state.patchSession);
 
-  const refreshPersonas = useCallback(async () => {
-    const { personas, drafts } = await listAgentGallery();
-    const agentStore = useAgentStore.getState();
-    agentStore.setPersonas(personas);
-    agentStore.setDraftSources(drafts);
-  }, []);
-
   const completeBuilder = useCallback(
     (source: AgentSourceEntry, refreshErrorMessage: string) => {
       clearBuilderSessionState(session.id);
 
       // Promotion is the durable source of truth. Seed the store immediately
       // so the destination profile exists even if the follow-up disk refresh
-      // fails or has not observed the promoted source yet.
+      // fails or has not observed the promoted source yet. Running the writes
+      // as a gallery mutation fences out any disk refresh that started before
+      // the promotion and would otherwise repaint the draft card.
       const promotedPersona = agentSourceToPersona(source);
       const agentStore = useAgentStore.getState();
-      const existingPersona = agentStore.personas.find(
-        (persona) => persona.id === promotedPersona.id,
-      );
-      if (existingPersona) {
-        agentStore.updatePersona(promotedPersona.id, promotedPersona);
-      } else {
-        agentStore.addPersona(promotedPersona);
-      }
-      // The draft just became this agent; drop its card without waiting for
-      // the disk refresh so the gallery never shows both at once.
-      for (const draft of agentStore.draftSources) {
-        if (draft.properties?.builderSessionId === session.id) {
-          agentStore.removeDraftSource(draft.path);
+      void agentStore.mutateGallery(() => {
+        const current = useAgentStore.getState();
+        const existingPersona = current.personas.find(
+          (persona) => persona.id === promotedPersona.id,
+        );
+        if (existingPersona) {
+          current.updatePersona(promotedPersona.id, promotedPersona);
+        } else {
+          current.addPersona(promotedPersona);
         }
-      }
+        // The draft just became this agent; drop its card without waiting
+        // for the disk refresh so the gallery never shows both at once.
+        for (const draft of current.draftSources) {
+          if (draft.properties?.builderSessionId === session.id) {
+            current.removeDraftSource(draft.path);
+          }
+        }
+      });
 
       onDraftPromoted?.(source);
       onAgentBuilderCompleted?.(promotedPersona.id);
 
-      void refreshPersonas().catch((error) => {
+      void agentStore.refreshGallery(listAgentGallery).catch((error) => {
         console.error(refreshErrorMessage, error);
       });
     },
-    [onAgentBuilderCompleted, onDraftPromoted, refreshPersonas, session.id],
+    [onAgentBuilderCompleted, onDraftPromoted, session.id],
   );
 
   const handleDraftPromoted = useCallback(
--- a/src/features/agents/capabilities/AgentBuilderCapability.tsx
+++ b/src/features/agents/capabilities/AgentBuilderCapability.tsx
@@ -62,7 +62,7 @@ export function AgentBuilderCapability({
       // the promotion and would otherwise repaint the draft card.
       const promotedPersona = agentSourceToPersona(source);
       const agentStore = useAgentStore.getState();
-      void agentStore.mutateGallery(() => {
+      const seeded = agentStore.mutateGallery(() => {
         const current = useAgentStore.getState();
         const existingPersona = current.personas.find(
           (persona) => persona.id === promotedPersona.id,
@@ -84,9 +84,13 @@ export function AgentBuilderCapability({
       onDraftPromoted?.(source);
       onAgentBuilderCompleted?.(promotedPersona.id);
 
-      void agentStore.refreshGallery(listAgentGallery).catch((error) => {
-        console.error(refreshErrorMessage, error);
-      });
+      // The refresh must start after the mutation releases the fence, or the
+      // fence would (correctly) reject it as having begun mid-mutation.
+      void seeded
+        .then(() => agentStore.refreshGallery(listAgentGallery))
+        .catch((error) => {
+          console.error(refreshErrorMessage, error);
+        });
     },
     [onAgentBuilderCompleted, onDraftPromoted, session.id],
   );
--- a/src/features/agents/capabilities/AgentBuilderCapability.tsx
+++ b/src/features/agents/capabilities/AgentBuilderCapability.tsx
@@ -17,7 +17,7 @@ import type { ChatSession } from "@/features/chat/stores/chatSessionStore";
 import { useChatSessionStore } from "@/features/chat/stores/chatSessionStore";
 import {
   agentSourceToPersona,
-  listPersonas,
+  listAgentGallery,
   type AgentSourceEntry,
 } from "@/shared/api/agents";
 
@@ -52,8 +52,10 @@ export function AgentBuilderCapability({
   const patchSession = useChatSessionStore((state) => state.patchSession);
 
   const refreshPersonas = useCallback(async () => {
-    const personas = await listPersonas();
-    useAgentStore.getState().setPersonas(personas);
+    const { personas, drafts } = await listAgentGallery();
+    const agentStore = useAgentStore.getState();
+    agentStore.setPersonas(personas);
+    agentStore.setDraftSources(drafts);
   }, []);
 
   const completeBuilder = useCallback(
@@ -72,6 +74,13 @@ export function AgentBuilderCapability({
         agentStore.updatePersona(promotedPersona.id, promotedPersona);
       } else {
         agentStore.addPersona(promotedPersona);
+      }
+      // The draft just became this agent; drop its card without waiting for
+      // the disk refresh so the gallery never shows both at once.
+      for (const draft of agentStore.draftSources) {
+        if (draft.properties?.builderSessionId === session.id) {
+          agentStore.removeDraftSource(draft.path);
+        }
       }
 
       onDraftPromoted?.(source);
--- a/src/features/agents/lib/agentBuilderIdentity.ts
+++ b/src/features/agents/lib/agentBuilderIdentity.ts
@@ -73,11 +73,14 @@ export function isPlaceholderDraftForSession(
   builderSessionId: string,
 ): boolean {
   const properties = source.properties ?? {};
+  // Everything "New agent" seeds on its own — identity, the stored
+  // provider/model preference, and a starter avatar — is not user content.
   const extraPropertyKeys = Object.keys(properties).filter(
     (key) =>
       key !== "draft" &&
       key !== "builderSessionId" &&
       key !== "provider" &&
+      key !== "modelProviderId" &&
       key !== "model" &&
       key !== "avatar",
   );
--- a/src/features/agents/lib/agentBuilderSession.ts
+++ b/src/features/agents/lib/agentBuilderSession.ts
@@ -523,11 +523,49 @@ export async function isDraftAgentBuilderSession(
  * (moved or removed outside the app). Editing an existing, present agent is
  * never discardable.
  */
-export async function isDiscardableAgentBuilderSession(
+export type UntouchedDraftDiscardOutcome =
+  | "discarded"
+  | "kept"
+  | "nothing-to-discard";
+
+/**
+ * Discards the session's draft only if it is still untouched at the moment of
+ * deletion. The user-content check is the last thing that runs before the
+ * file goes, so anything typed while earlier lookups were in flight keeps the
+ * draft ("kept") instead of being silently thrown away. Editing an existing
+ * agent is never discardable; with no content it reports "nothing-to-discard"
+ * so callers can navigate freely.
+ *
+ * `onBeforeDiscard` runs once the decision is final and before the chat
+ * closes — callers navigate there, because closing the active chat redirects
+ * home and would stomp on where the user was going.
+ */
+export async function discardUntouchedDraftAgentSession(
   sessionId: string,
-): Promise<boolean> {
+  deps: CloseSessionDeps & { onBeforeDiscard?: () => void } = {},
+): Promise<UntouchedDraftDiscardOutcome> {
   const source = await findCurrentBuilderSource(sessionId);
-  return source === undefined || source.properties?.draft === true;
+  const isDraft = source === undefined || source.properties?.draft === true;
+
+  if (await hasAgentBuilderSessionUserContent(sessionId)) {
+    return "kept";
+  }
+  if (!isDraft) {
+    return "nothing-to-discard";
+  }
+
+  deps.onBeforeDiscard?.();
+  try {
+    if (source) {
+      await discardAgentBuilderSource(source.path);
+    }
+  } catch (error) {
+    console.warn("Failed to delete agent builder draft during discard:", error);
+  } finally {
+    clearBuilderSessionState(sessionId);
+    await deps.closeSession?.(sessionId);
+  }
+  return "discarded";
 }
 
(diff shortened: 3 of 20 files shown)


src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -517,6 +523,63 @@ export async function isDraftAgentBuilderSession(
...
+  const source = await findCurrentBuilderSource(sessionId);
+  const isDraft = source === undefined || source.properties?.draft === true;
+
+  if (await hasAgentBuilderSessionUserContent(sessionId)) {
+    return "kept";
+  }
+  // The check above awaited a disk read after its in-memory look. Anything
+  // typed during that read is invisible to it, so look once more — with no
+  // await between here and the delete.
+  if (hasLocalAgentBuilderUserContent(sessionId)) {
```
johnmatthewtennant, line 592: 🤖 An edit can be in `saveStatus === "saving"` while `localEditSessionIds` is false. During that window the disk and local-source checks can still see the old placeholder, allowing navigation to delete the draft before its write lands. The in-flight save does not appear to be represented in the discard guard.

src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -458,9 +458,12 @@ export async function isEmptyDraftAgentSession(
...
 
-export async function hasAgentBuilderSessionUserContent(
-  sessionId: string,
-): Promise<boolean> {
+/**
+ * The in-memory half of the user-content check: unsaved rail edits, composer
+ * text, queued messages, sent messages. Synchronous on purpose — callers that
+ * are about to delete something re-run this with no await in between.
+ */
+export function hasLocalAgentBuilderUserContent(sessionId: string): boolean {
```
johnmatthewtennant, line 485: 🤖 This content check handles queued text but not attachment-only composer or queue state. `ChatInput` accepts attachment-only messages, so navigation can classify that session as untouched, close it, and discard `draftAttachmentsBySession[sessionId]` or queued `payload.attachments`.

src/features/agents/lib/agentBuilderIdentity.ts
```diff
@@ -73,11 +73,14 @@ export function isPlaceholderDraftForSession(
...
 ): boolean {
   const properties = source.properties ?? {};
+  // Everything "New agent" seeds on its own — identity, the stored
+  // provider/model preference, and a starter avatar — is not user content.
   const extraPropertyKeys = Object.keys(properties).filter(
     (key) =>
       key !== "draft" &&
       key !== "builderSessionId" &&
       key !== "provider" &&
+      key !== "modelProviderId" &&
```
johnmatthewtennant, line 83: 🤖 These setup keys are ignored regardless of their values. A user who changes only the avatar, provider, or model can therefore have the persisted draft classified as an untouched placeholder, hidden from the gallery, and silently deleted on navigation. Setup-only edits are indistinguishable from the seeded values here.

src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -517,6 +523,63 @@ export async function isDraftAgentBuilderSession(
...
+ * home and would stomp on where the user was going.
+ */
+export async function discardUntouchedDraftAgentSession(
+  sessionId: string,
+  deps: CloseSessionDeps & { onBeforeDiscard?: () => void } = {},
+): Promise<UntouchedDraftDiscardOutcome> {
+  const source = await findCurrentBuilderSource(sessionId);
+  const isDraft = source === undefined || source.properties?.draft === true;
+
+  if (await hasAgentBuilderSessionUserContent(sessionId)) {
```
johnmatthewtennant, line 586: 🤖 If the fresh source read fails while the backend still lists the file, the lookup can return stale placeholder metadata and this branch proceeds to deletion. A temporarily unreadable, user-modified draft can therefore be treated as confirmed empty.

src/features/agents/ui/AgentsView.tsx
```diff
@@ -257,29 +280,36 @@ export function AgentsView({
...
-        closeSession: onDeleteDraftSession,
+    (draft: GalleryDraft) => {
+      const { sessionId, source } = draft;
+      // Run as a gallery mutation so a disk refresh that started before the
+      // delete cannot land afterwards and put the card back.
+      void mutateGallery(async () => {
+        if (sessionId) {
+          await deleteDraftAgentSession(sessionId, {
+            closeSession: onDeleteDraftSession,
+          });
```
johnmatthewtennant: 🤖 If two draft files share a `builderSessionId` (for example, an external copy preserving frontmatter), both cards can bind to the same session. Deleting the non-target card then follows the session target and removes a different file than the selected `source.path`.

src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -517,6 +523,63 @@ export async function isDraftAgentBuilderSession(
...
+  }
+
+  deps.onBeforeDiscard?.();
+  try {
+    if (source) {
+      await discardAgentBuilderSource(source.path);
+    }
+  } catch (error) {
+    console.warn("Failed to delete agent builder draft during discard:", error);
+  } finally {
```
johnmatthewtennant, line 611: 🤖 A real deletion failure is swallowed here, but the `finally` block still clears builder state, closes the session, and returns `"discarded"`. Because untouched placeholders are hidden from the gallery, the surviving file becomes an invisible orphan with no retry path, and a non-not-found error is indistinguishable from successful discard.

src/features/agents/capabilities/AgentBuilderCapability.tsx
```diff
@@ -51,37 +51,48 @@ export function AgentBuilderCapability({
...
+        );
+        if (existingPersona) {
+          current.updatePersona(promotedPersona.id, promotedPersona);
+        } else {
+          current.addPersona(promotedPersona);
+        }
+        // The draft just became this agent; drop its card without waiting
+        // for the disk refresh so the gallery never shows both at once.
+        for (const draft of current.draftSources) {
+          if (draft.properties?.builderSessionId === session.id) {
```
johnmatthewtennant, line 78: 🤖 An orphan draft reopened without its original chat gets a new session ID while its frontmatter can retain the vanished ID. Promotion then misses the original draft in this loop, so the store temporarily retains both it and the promoted persona—and retains it indefinitely if refresh fails. The optimistic removal is keyed only by the current session ID rather than the promoted draft path.

src/features/agents/capabilities/AgentBuilderCapability.tsx
```diff
@@ -51,37 +51,48 @@ export function AgentBuilderCapability({
...
       onDraftPromoted?.(source);
       onAgentBuilderCompleted?.(promotedPersona.id);
 
-      void refreshPersonas().catch((error) => {
-        console.error(refreshErrorMessage, error);
-      });
+      // The refresh must start after the mutation releases the fence, or the
+      // fence would (correctly) reject it as having begun mid-mutation.
+      void seeded
+        .then(() => agentStore.refreshGallery(listAgentGallery))
```
johnmatthewtennant, line 90: 🤖 `promotePersonaSource` can create the destination while failing to remove the original draft. This immediate refresh then lists and re-adds that orphan beside the promoted agent, restoring the duplicate card after the optimistic mutation removed it.

src/features/agents/ui/AgentsView.tsx
```diff
@@ -257,29 +280,36 @@ export function AgentsView({
...
+        } else {
+          await discardAgentBuilderSource(source.path);
```
kalvinnchau: **P1 — Validate a sessionless card's target before deleting.**

`draftSources` is a prior disk snapshot. If a sessionless draft at `P.md` is moved and a finished agent is created at `P.md` before the next refresh, this sends `sources/delete` for the replacement without rereading it or checking `properties.draft`. A stale Draft card can therefore delete a finished agent. Re-read and validate the selected source (ideally its identity/version too) immediately before deletion.

src/features/agents/lib/agentBuilderSourceLifecycle.ts
```diff
@@ -143,12 +145,12 @@ export async function findAgentBuilderSource(
   );
 
   if (foundByPath && !isEmptyPlaceholderDraft(foundByPath)) {
```
kalvinnchau: **P2 — Retry the backend-listed moved source after evicting a stale cache entry.**

An edited draft keeps cached path `A`; after an external rename the backend lists `B` with the same `builderSessionId`. This branch selects stale `A`; its failed read evicts `A` and returns `undefined` without trying the already-computed `movedNonPlaceholder` (`B`). Delete then clears the session/card while `B` remains and reappears on refresh. Prefer the backend-listed match or retry it after eviction, with a test that primes the edited cache before renaming.

src/features/agents/ui/AgentsView.tsx
```diff
@@ -257,29 +280,36 @@ export function AgentsView({
...
-    (sessionId: string) => {
-      const session = useChatSessionStore.getState().getSession(sessionId);
-      if (!session?.targetAgentPath) {
-        return;
-      }
-
+    (draft: GalleryDraft) => {
+      // Starting by path reopens the live builder chat when there is one and
+      // otherwise opens a fresh builder on the same file.
       onStartAgentBuilderSession?.({
```
kalvinnchau, line 286: **P2 — Continue opens a second builder after a filename move.**

The gallery has already identified the original live session as `draft.sessionId`, but Continue discards it and starts only by the new path/slug. The original session still has the old path, so `findLiveBuilderSession()` misses it and creates another builder for the same source; reconciliation later rebinds the original session too. Reuse/rebind `draft.sessionId`, or update the source's owner metadata when intentionally creating a replacement.

src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -517,6 +523,63 @@ export async function isDraftAgentBuilderSession(
...
+    clearBuilderSessionState(sessionId);
+    await deps.closeSession?.(sessionId);
```
kalvinnchau: **P2 — Resolve the provisional session ID before closing.**

A new builder begins under a client-side draft ID and is asynchronously promoted to the ACP ID. If the user leaves while creation is pending, this archives the captured provisional ID; after promotion no archive reaches the backend session, leaving the empty chat behind. Resolve the live builder ID immediately before close/archive (and handle promotion racing that resolution).

src/features/agents/lib/agentBuilderSession.ts
```diff
@@ -517,6 +523,63 @@ export async function isDraftAgentBuilderSession(
...
+  // typed during that read is invisible to it, so look once more — with no
+  // await between here and the delete.
+  if (hasLocalAgentBuilderUserContent(sessionId)) {
+    return "kept";
+  }
+  if (!isDraft) {
+    return "nothing-to-discard";
+  }
+
+  deps.onBeforeDiscard?.();
```
kalvinnchau, line 599: **P2 — Silent discard redirects Settings to Home.**

Navigating first only works for routes that clear `activeSessionId`. `openSettings()` preserves it, so after this opens Settings the archive still sees the builder as active and unconditionally changes the view to Home. New agent → no edits → Settings briefly opens Settings, then lands on Home. Detach the builder before archival or make archival preserve an already-selected non-chat route.

src/features/agents/stores/agentStore.ts
```diff
@@ -126,6 +148,39 @@ export const useAgentStore = create<AgentStore>((set, get) => ({
...
   setPersonasLoading: (personasLoading) => set({ personasLoading }),
 
+  setDraftSources: (draftSources) => set({ draftSources }),
+
+  removeDraftSource: (path) =>
+    set((state) => ({
+      draftSources: state.draftSources.filter((draft) => draft.path !== path),
+    })),
+
+  refreshGallery: async (fetchGallery) => {
```
kalvinnchau, line 162: **P2 — The gallery fence does not make disk snapshots latest-wins.**

The revision changes only through `mutateGallery()`. Navigation-driven discards bypass it, so a listing that started before the delete can still apply afterward. Independently, two refreshes at the same revision can resolve out of order after an external file removal: a newer no-draft listing applies, then an older listing restores the card. Advance the fence for every draft lifecycle delete and track a latest-issued refresh generation so older snapshots cannot overwrite newer ones.

kalvinnchau (commented):
Additional non-overlapping P1/P2 findings from review.

PR #177, opened 2026-08-24
feat(voice): show what was spoken before interruption

johnmatthewtennant:
## Summary

When voice playback is interrupted, Berd estimates the last completed word from actual Pocket or Siri playback progress. The transcript keeps heard text normally and strikes the estimated unheard remainder, while a bounded hidden delivery notice tells the model what was and was not spoken. Delivery decoration remains ephemeral and clears on reload.

Start a voice conversation, request a multi-paragraph response, and interrupt it mid-reply by speaking. The spoken prefix remains normal, all estimated unspoken text is struck, and the next assistant turn receives the delivery estimate. Repeat with Pocket and Siri, and use hang-up to confirm the stop reason is reported without claiming the user spoke.

### DIFF
--- a/src-tauri/src/commands/pocket_voice.rs
+++ b/src-tauri/src/commands/pocket_voice.rs
@@ -2004,9 +2004,13 @@ fn run_pocket_voice_stream(
                     &mut last_progress_emit,
                     false,
                 )? {
+                    let delivery = capture_before_stop(
+                        || pocket_delivery_snapshot(&delivery_ledger, &player),
+                        || player.stop(),
+                    );
                     return Ok(PocketStreamOutcome {
                         state: PocketStreamEventState::Interrupted,
-                        delivery: Some(pocket_delivery_snapshot(&delivery_ledger, &player)),
+                        delivery: Some(delivery),
                     });
                 }
             }
@@ -2028,9 +2032,13 @@ fn run_pocket_voice_stream(
                     &mut last_progress_emit,
                     true,
                 )? {
+                    let delivery = capture_before_stop(
+                        || pocket_delivery_snapshot(&delivery_ledger, &player),
+                        || player.stop(),
+                    );
                     return Ok(PocketStreamOutcome {
                         state: PocketStreamEventState::Interrupted,
-                        delivery: Some(pocket_delivery_snapshot(&delivery_ledger, &player)),
+                        delivery: Some(delivery),
                     });
                 }
                 let tail = speed_processor.drain_and_reset()?;
@@ -2071,9 +2079,13 @@ fn run_pocket_voice_stream(
                     &mut last_progress_emit,
                     true,
                 )? {
+                    let delivery = capture_before_stop(
+                        || pocket_delivery_snapshot(&delivery_ledger, &player),
+                        || player.stop(),
+                    );
                     return Ok(PocketStreamOutcome {
                         state: PocketStreamEventState::Interrupted,
-                        delivery: Some(pocket_delivery_snapshot(&delivery_ledger, &player)),
+                        delivery: Some(delivery),
                     });
                 }
                 let tail = speed_processor.finish()?;
@@ -2147,6 +2159,16 @@ fn pocket_delivery_snapshot(
     ledger.snapshot(queued_pieces, current_piece_frames)
 }
 
+#[cfg(any(test, target_os = "macos"))]
+fn capture_before_stop(
+    snapshot: impl FnOnce() -> VoiceDeliveryProgress,
+    stop: impl FnOnce(),
+) -> VoiceDeliveryProgress {
+    let delivery = snapshot();
+    stop();
+    delivery
+}
+
 #[cfg(target_os = "macos")]
 #[allow(clippy::too_many_arguments)]
 fn synthesize_pocket_stream_ready(
@@ -2171,7 +2193,6 @@ fn synthesize_pocket_stream_ready(
     *first_chunk_pending = split.first_chunk_pending;
     for text in split.ready {
         if !active.load(Ordering::SeqCst) {
-            player.stop();
             return Ok(false);
         }
         let text = text.trim().to_string();
@@ -2230,7 +2251,6 @@ fn synthesize_pocket_stream_ready(
             return Err(error);
         }
         if !completed {
-            player.stop();
             return Ok(false);
         }
         let final_total_frames = speed_processor
@@ -2412,6 +2432,33 @@ mod tests {
         assert_eq!(progress.segments[1].played_frames, 0);
         assert_eq!(progress.segments[1].total_frames, 4_800);
         assert!(progress.segments[1].synthesis_complete);
+    }
+
+    #[test]
+    fn cancellation_captures_delivery_before_stopping_playback() {
+        use std::cell::RefCell;
+
+        let mut ledger = PlaybackDeliveryLedger::default();
+        ledger.begin_segment("Played piece.".to_string());
+        ledger.append_frames(4_800);
+        ledger.complete_segment(4_800);
+        ledger.begin_segment("Queued audio.".to_string());
+        ledger.append_frames(4_800);
+        ledger.append_frames(4_800);
+        ledger.complete_segment(9_600);
+        let calls = RefCell::new(Vec::new());
+
+        let delivery = capture_before_stop(
+            || {
+                calls.borrow_mut().push("snapshot");
+                ledger.snapshot(2, 1_200)
+            },
+            || calls.borrow_mut().push("stop"),
+        );
+
+        assert_eq!(&*calls.borrow(), &["snapshot", "stop"]);
+        assert_eq!(delivery.segments[0].played_frames, 3_600);
+        assert_eq!(delivery.segments[1].played_frames, 0);
     }
 
     #[test]
--- a/src-tauri/src/commands/pocket_voice.rs
+++ b/src-tauri/src/commands/pocket_voice.rs
@@ -191,6 +191,8 @@ struct VoiceDeliverySegment {
 #[cfg(any(test, target_os = "macos"))]
 #[derive(Clone, Debug, Serialize)]
 struct VoiceDeliveryProgress {
+    #[serde(rename = "sampleRate")]
+    sample_rate: u32,
     segments: Vec<VoiceDeliverySegment>,
 }
 
@@ -256,7 +258,10 @@ impl PlaybackDeliveryLedger {
                 }
             })
             .collect();
-        VoiceDeliveryProgress { segments }
+        VoiceDeliveryProgress {
+            sample_rate: berd_voice::SAMPLE_RATE,
+            segments,
+        }
     }
 }
 
--- a/src-tauri/src/commands/pocket_voice.rs
+++ b/src-tauri/src/commands/pocket_voice.rs
@@ -12,7 +12,7 @@ use std::sync::mpsc;
 use std::sync::{Arc, Mutex, OnceLock};
 use std::time::{Duration, Instant, SystemTime};
 
-#[cfg(target_os = "macos")]
+#[cfg(any(test, target_os = "macos"))]
 use berd_voice::SAMPLE_RATE;
 #[cfg(target_os = "macos")]
 use berd_voice::{load_text_to_speech, load_voice_style, PocketTts, VoiceStyle};
@@ -42,6 +42,10 @@ const DOWNLOAD_READ_TIMEOUT: Duration = Duration::from_secs(30);
 const DOWNLOAD_TOTAL_TIMEOUT: Duration = Duration::from_secs(30 * 60);
 #[cfg(target_os = "macos")]
 const STREAMING_EMIT_FRAMES: usize = 12;
+#[cfg(target_os = "macos")]
+const PLAYBACK_PROGRESS_EMIT_INTERVAL: Duration = Duration::from_millis(100);
+#[cfg(any(test, target_os = "macos"))]
+const PLAYBACK_LATENCY_SAFETY_FRAMES: u64 = SAMPLE_RATE as u64 / 10;
 const PARAKEET_ARCHIVE: Artifact = Artifact {
     filename: "parakeet.tar.bz2",
     size: 104_337_827,
@@ -158,6 +162,7 @@ enum PocketStreamCommand {
 #[serde(rename_all = "camelCase")]
 enum PocketStreamEventState {
     Started,
+    Progress,
     Completed,
     Interrupted,
     Failed,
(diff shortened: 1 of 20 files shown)


src/features/voice-conversation/lib/nativeAssistantSpeech.ts
```diff
@@ -249,17 +472,31 @@ function handleStreamEvent(
...
-  activeUtterance = null;
-  if (utterance) {
-    setUtteranceStatus(utterance, "interrupted");
-    recordPlaybackNotice(
-      utterance.sessionId,
-      utterance.id,
+  if (utterance && !utterance.interruptionRequested) {
+    utterance.interruptionRequested = true;
+    utterance.interruptionCause = cause;
+    if (terminalEventExpected) utterance.onInterrupted();
```
morgmart, line 587: 🤖 **P1 · Interruption can wait forever** (blocking)

After the native stream has started, barge-in deliberately keeps activeUtterance alive and waits for an interrupted event before finalizing. The stop calls discard both their boolean result and failures, and there is no presentation timeout or fallback. If stop reports that nothing was active, rejects, or the terminal event never reaches the listener, the stale utterance remains the source of truth indefinitely; the new test only supplies the happy-path interrupted event manually.

**User effect:** After interrupting speech, Berd can remain attached to a stopped utterance and later assistant text can be blocked or absorbed into that stale stream instead of being spoken normally.

**Recommended fix:** Bound the terminal-event wait. If stop fails or reports false, finalize immediately from the latest delivery estimate; otherwise arm an idempotent timeout that finalizes the same way if no terminal event arrives, while allowing a real terminal event to win safely.

**Test:** Add tests where nativeStreamStarted is true and stop rejects, returns false, or resolves without any terminal event. Advance the fallback timeout and assert the old utterance is finalized once and a subsequent assistant update starts a new stream; also race the timeout against a late interrupted event and assert single finalization.

src-tauri/src/commands/pocket_voice.rs
```diff
@@ -1902,9 +2000,14 @@ fn run_pocket_voice_stream(
...
                     &mut first_chunk_pending,
                     &mut playback_started,
+                    &mut delivery_ledger,
+                    &mut last_progress_emit,
                     false,
                 )? {
-                    return Ok(PocketStreamEventState::Interrupted);
+                    return Ok(PocketStreamOutcome {
+                        state: PocketStreamEventState::Interrupted,
+                        delivery: Some(pocket_delivery_snapshot(&delivery_ledger, &player)),
```
morgmart: 🤖 **P1 · Snapshot follows player reset** (blocking)

When Pocket synthesis is cancelled inside synthesize_pocket_stream_ready, that helper stops the Rodio player before returning false. The caller then takes the interruption delivery snapshot. Because stop can reset playback position and alter/drain queue depth, the new ledger can undercount delivered audio as zero or misclassify queued pieces as completed, making the estimate race-dependent at the exact interruption boundary.

**User effect:** On Pocket voice interruptions, Berd can strike text the user already heard or leave unheard text unstruck, and can pass the same incorrect delivery claim into the next model turn.

**Recommended fix:** Capture delivery before stopping or return enough cancellation state to the caller so it snapshots first and stops second. Keep the ordering consistent across every interruption exit.

**Test:** Add a discriminating Pocket ledger/player test for cancellation during synthesis with one played piece and additional queued audio. Assert the terminal delivery snapshot preserves the pre-stop consumed position and does not count queued audio as heard.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
```diff
@@ -444,15 +700,67 @@ export function startNativeAssistantSpeech(
         consumedTextBySlot.set(slot, content.text);
         if (!delta) continue;
 
+        if (interruptedMessages.has(message.id)) {
```
morgmart, line 862: 🤖 **P2 · Rewrites reuse stale cutoff** (non-blocking)

The inspection path recognizes a non-prefix text change as a replacement by assigning the entire new text to delta, but once a message has been interrupted it still reuses the prior numeric spokenThrough offset. That offset described the old source text and has no valid mapping into unrelated replacement text.

**User effect:** If a streamed assistant block is revised after interruption, Berd can present part of the replacement as already spoken, strike the wrong suffix, and tell the model that words the user never heard were delivered.

**Recommended fix:** Distinguish append-only updates from replacements in the interrupted-message path. Preserve the cutoff only for append-only growth; for a non-prefix replacement, reset the replacement to not spoken (or introduce an explicit source mapping if revisions must preserve delivery state).

**Test:** Interrupt a block after a nonzero cutoff, replace its text with a non-prefix value, and assert the replacement is entirely not spoken and the delivery notice contains no spoken prefix from the new text.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review of exact range 8e59aac2f0708d23ae5ff5ecd5006b7e6d7653f0...7fb78912dd5f4d330657176be444331d97185bb0 found 2 blocking and 1 non-blocking findings. The supplied GitHub evidence was inspected: it is structurally available for the exact head, reports no check runs, and has a pending combined status with zero statuses; required checks still govern merge readiness. Final self-check covered every changed voice-interruption and transcript-rendering flow, accessibility, localization, navigation/consent, async/error/never-completes/lifecycle/races, test honesty, design-system/project rules, duplicate overlap, and concrete user effect. No supplied review threads exist, so no finding was suppressed as a duplicate.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
johnmatthewtennant: 🤖 Fixed in 8c84fb12: append-only updates retain their delivery cutoff, while non-prefix replacements reset to not spoken with cutoff zero and replace the model notice with the full rewritten text as unspoken. A discriminating rewrite test covers this behavior.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
johnmatthewtennant: 🤖 Fixed in 7c3179c2: native stop false or rejection finalizes immediately, a missing terminal event falls back after a bounded wait, and terminal-versus-timeout finalization is idempotent. Tests cover all four cases; dbc92b12 also closes deferred-start and late-started-event races.

src-tauri/src/commands/pocket_voice.rs
johnmatthewtennant: 🤖 Fixed in 8f0f5a04: every synthesis-cancellation exit captures the Pocket delivery ledger before stopping Rodio. The regression test proves snapshot-before-stop ordering while preserving played frames and excluding queued audio.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
```diff
@@ -85,40 +121,220 @@ function recordPlaybackNotice(
...
+    const segmentStart = text.indexOf(segment.text, searchFrom);
+    if (segmentStart === -1) continue;
+    matchedSegment = true;
+    const totalFrames = Math.max(0, segment.totalFrames);
+    const playedFrames = Math.max(
+      0,
+      Math.min(totalFrames, segment.playedFrames),
+    );
+    if (totalFrames === 0 || playedFrames === 0) break;
+    if (!segment.synthesisComplete) {
```
morgmart, line 224: 🤖 **P1 · Partial playback becomes unspoken** (blocking)

estimateSpeechDelivery stops at any segment whose synthesisComplete flag is false, even when that segment reports nonzero playedFrames. Both native paths can play generated frames while synthesis for the same segment is still ongoing, so interruption during ordinary streaming discards direct evidence that audio was heard. The added test explicitly accepts spokenThrough zero for this case rather than protecting the user-visible delivery truth.

**User effect:** When the user interrupts while a segment is still being synthesized, Berd can strike the entire reply and tell the next model turn that nothing was spoken even though the user heard a prefix.

**Recommended fix:** Use the playedFrames-to-totalFrames ratio to estimate a completed-word cutoff for incomplete segments too, while assigning low confidence because the denominator may still grow. Do not classify known played audio as wholly unspoken.

**Test:** Replace the current incomplete-synthesis expectation with a discriminating case containing nonzero played frames and assert a nonzero completed-word cutoff, low confidence, and a notice whose spoken prefix matches the rendered boundary.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
```diff
@@ -249,29 +490,51 @@ function handleStreamEvent(
...
-      utterance.sessionId,
-      utterance.id,
-      utterance.text,
-      "interrupted",
+  if (utterance && !utterance.interruptionRequested) {
+    utterance.interruptionRequested = true;
+    utterance.interruptionCause = cause;
+    if (terminalEventExpected) utterance.onInterrupted();
+  }
+  if (utterance && !terminalEventExpected) {
```
morgmart, line 589: 🤖 **P1 · Hang-up drops terminal delivery** (blocking)

User barge-in opts into waiting for the native interrupted event, but stopNativeAssistantSpeech calls interruptActiveUtterance with the default false value. It therefore finalizes immediately from the last periodic progress event and then removes the stream listener. The native stop still computes a more accurate terminal snapshot, but that event can no longer update the transcript or model notice.

**User effect:** Hanging up can strike words the user already heard and carry an outdated delivery claim into the next model turn, especially when playback advances between periodic progress events.

**Recommended fix:** Use the same bounded terminal-delivery path for conversation stop as for barge-in: keep the listener alive until the terminal interruption snapshot arrives, with the existing timeout/error fallback, then unsubscribe after idempotent finalization.

**Test:** Emit one stale progress snapshot, advance playback, trigger the lifecycle hang-up, then emit a terminal interruption snapshot with a later cutoff. Assert the later cutoff drives both transcript decoration and the notice, and separately assert the timeout still finalizes if no terminal event arrives.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
```diff
@@ -85,40 +121,220 @@ function recordPlaybackNotice(
...
+}
+
+function applyInterruptionEstimate(
+  utterance: ActiveUtterance,
+  estimate: SpeechDeliveryEstimate,
+) {
+  const firstTargetKey = utterance.targets[0]
+    ? targetKey(utterance.targets[0])
+    : null;
+  for (const target of utterance.targets) {
```
morgmart: 🤖 **P2 · Interleaved spans corrupt cutoffs** (non-blocking)

Delivery for each text target is reduced to the first matching span start and last matching span end. The append path only merges immediately adjacent spans, so updates ordered A, B, A leave two A spans with B's audio between them. Treating A as one contiguous range includes B's characters in A's spokenThrough calculation and shifts the interruption boundary.

**User effect:** For assistant output that streams across multiple text blocks, Berd can strike the wrong words and report a spoken offset that does not belong to the displayed block.

**Recommended fix:** Derive each target's delivered character count by intersecting the global cutoff with every span for that target and summing only that target's covered lengths, rather than using a first-to-last bounding range.

**Test:** Stream updates in A, B, A order, interrupt inside the final A span, and assert each block receives its own exact local cutoff and the notices match those rendered boundaries.

src/features/chat/ui/AgentWorkPanel.tsx
```diff
@@ -307,6 +307,13 @@ function AgentWorkItemRow({
 
   if (item.kind === "progress") {
     const speechStatus = item.content.speech?.status;
+    const strikethroughFrom =
```
morgmart, line 311: 🤖 **P2 · Grouped progress loses boundaries** (non-blocking)

Agent-work timeline construction merges adjacent text blocks whenever their speech status matches, retaining only the first block's speech object. The new renderer then applies that first block's local spokenThrough value to the concatenated text. Two interrupted blocks with distinct local cutoffs therefore cannot be represented correctly after grouping.

**User effect:** In agent-work content, the strike can begin at the wrong word and make heard text look unheard or unheard text look delivered.

**Recommended fix:** Do not merge speech-decorated text blocks, or compute a valid combined delivery model that preserves every block boundary and local cutoff before rendering.

**Test:** Render adjacent interrupted agent-work text blocks with different spokenThrough values and assert both visible strike boundaries are preserved.

src/shared/ui/ai-elements/message.tsx
```diff
@@ -694,6 +703,79 @@ const berdRehypePlugins: NonNullable<
...
+    "table",
+    "tbody",
+    "tfoot",
+    "thead",
+    "tr",
+    "ul",
+  ]);
+  const wrap = (node: MarkdownHastNode): MarkdownHastNode => ({
+    type: "element",
+    tagName: "del",
```
morgmart: 🤖 **P2 · Strike implies deleted content** (non-blocking)

The shared renderer expresses the unheard suffix with semantic del elements. The text was not deleted or superseded; it remains valid assistant output with a voice-delivery state. Screen readers may announce it as deleted, or may omit deletion boundaries entirely, so the visual distinction does not have an equivalent accurate accessible meaning.

**User effect:** Screen-reader users can hear misleading deletion semantics or cannot determine which portion of an interrupted reply was actually spoken.

**Recommended fix:** Use presentation-only strike styling for the visual treatment and expose an explicit accessible spoken/unspoken boundary or label whose semantics describe voice delivery rather than document deletion.

**Test:** Add an accessibility-focused rendering test that verifies the suffix is not represented as inserted/deleted document content and that assistive text identifies the unspoken portion or boundary.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review of exact range 8e59aac2f0708d23ae5ff5ecd5006b7e6d7653f0...dbc92b12465623e4c45071dd37e3f877ddbf0b2b found 2 blocking and 3 non-blocking publishable findings. One additional proposed finding was suppressed because it overlaps a resolved automation thread with a substantive human-account reply. The supplied GitHub evidence was inspected for the exact head; it reports a pending combined status with zero statuses/check runs, so required checks still govern merge readiness. Final self-check covered the complete voice interruption, hang-up, native delivery accounting, transcript and agent-work rendering flows; accessibility; localization; navigation and consent; async errors, never-completes, lifecycle and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity.

Deterministic publication result: 2 blocking and 3 non-blocking finding(s) publishable; 1 duplicate(s) suppressed.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
johnmatthewtennant: 🤖 Added actual output sample rates to both native delivery snapshots and now estimate in-progress synthesis with a conservative duration bound and low confidence. An incomplete segment can never be classified as fully spoken; long and short partial-playback regressions cover the boundary.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
johnmatthewtennant: 🤖 Hang-up now uses the same bounded terminal-delivery path as barge-in. Every terminal and fallback path is also scoped to the lifecycle, session, and voice revision so late events cannot revive or overwrite a stopped or restarted conversation; tests cover terminal, timeout, completion, failure, and restart races.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
johnmatthewtennant: 🤖 Delivery cutoffs now sum the intersections across every span belonging to each text target instead of treating the first-to-last range as contiguous. A discriminating A/B/A interleaving test preserves each block’s exact local cutoff.

src/features/chat/ui/AgentWorkPanel.tsx
johnmatthewtennant: 🤖 Speech-decorated AgentWork text blocks no longer merge, so each block retains its own local delivery cutoff. The regression test renders adjacent interrupted blocks with different boundaries.

src/shared/ui/ai-elements/message.tsx
johnmatthewtennant: 🤖 Replaced semantic deletion markup with presentation-only strike styling plus a localized accessible spoken/unspoken boundary. The renderer and tests cover Markdown structure, void elements, images including empty alt text, escapes, entities, and locale changes.

src-tauri/src/commands/pocket_voice.rs
```diff
@@ -755,19 +854,15 @@ pub fn start_pocket_voice_stream(
...
-                Ok(other) => (other, None),
+            let (event_state, error, delivery) = match result {
+                Ok(outcome) => (outcome.state, None, outcome.delivery),
                 Err(error) if !active.load(Ordering::SeqCst) => {
                     log::debug!("Pocket voice stream stopped after error: {error}");
-                    (PocketStreamEventState::Interrupted, None)
+                    (PocketStreamEventState::Interrupted, None, None)
                 }
-                Err(error) => (PocketStreamEventState::Failed, Some(error)),
+                Err(error) => (PocketStreamEventState::Failed, Some(error), None),
```
morgmart: 🤖 **P2 · Failures discard partial delivery** (non-blocking)

Both native stream wrappers emit failed terminal events with delivery set to null for every worker error. Those errors can occur after playback has started, such as a later Pocket synthesis or DSP failure or a later Siri synthesis failure. The renderer's failed-event path likewise ignores the latest delivery progress and marks the whole utterance failed, even though the new feature otherwise tracks the heard boundary.

**User effect:** If native speech fails after reading part of a reply, Berd loses which words the user heard and gives no accurate spoken-versus-unspoken boundary for recovery or the next turn.

**Recommended fix:** Preserve a final delivery snapshot on post-start failures and have failed-event handling apply that estimate before recording the failure. Keep a generic whole-utterance failure only when playback never started or no delivery evidence exists.

**Test:** For both Pocket and Siri, simulate a failure after a progress snapshot with nonzero played frames and assert the terminal event retains delivery. In the renderer test, assert the transcript and model notice preserve the heard prefix while still exposing the speech failure.

morgmart (commented):
🤖 Automated code review

COMMENT. Fresh static review of exact range 8e59aac2f0708d23ae5ff5ecd5006b7e6d7653f0...ab879012a7e856cda775fdfbc6d0682ca9237148 found 0 blocking and 1 non-blocking publishable finding. Three additional proposed findings were suppressed because each maps to the same underlying issue as a resolved automation thread with a substantive human-account reply. The supplied GitHub evidence was inspected for the exact head: all eight supplied check runs completed successfully, while required checks still independently govern merge readiness. Final self-check covered every changed voice start, streaming, interruption, hang-up, failure, delivery-estimation, transcript, agent-work, Markdown and accessibility flow; localization; navigation and consent; async errors, never-completes, lifecycle and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for severity.

Deterministic publication result: 0 blocking and 1 non-blocking finding(s) publishable; 3 duplicate(s) suppressed.

src-tauri/src/commands/pocket_voice.rs
johnmatthewtennant: 🤖 Pocket and Siri now snapshot played-audio delivery before stopping or cancelling and attach that evidence to terminal failure events. The renderer preserves per-block heard cutoffs, failed status, and the model notice; later streamed deltas extend the failed unspoken suffix without retrying the response. Added parity tests for both backends plus native snapshot-order tests.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of exact range 8e59aac2f0708d23ae5ff5ecd5006b7e6d7653f0...4c4cf4d3e210d040ad23492cf8c74c8d3de8b438 found no publishable findings. Five proposed concerns were suppressed because each maps to a resolved prior automation thread with a substantive human-account reply; no unresolved or unmatched suppression remains. The supplied GitHub evidence was inspected for the exact head: all eight supplied check runs completed successfully, while required checks still independently govern merge readiness. Final self-check covered every changed voice startup, streaming, interruption, hang-up, timeout, failure and delivery-accounting path; transcript, agent-work and Markdown rendering; accessibility; localization; navigation and consent; async errors, never-completes, lifecycle and races; test honesty; design-system/project rules; duplicate overlap; and concrete evidence plus user effect for blocking severity.

Deterministic publication result: 0 blocking and 0 non-blocking finding(s) publishable; 5 duplicate(s) suppressed.

PR #181, opened 2026-08-24
feat(voice): add native macOS speech recognition

johnmatthewtennant:
## Summary

Adds Apple Speech Recognition as an on-device speech-input backend on supported Macs. Berd selects it by default when hardware and locale support are available, requires no separate model download, and falls back to Parakeet when unavailable.

Voice settings expose Apple Speech Recognition only when usable for the current locale. Native transcripts preserve voice-conversation lifecycle and delivery semantics across startup, shutdown, failures, and retained transcript recovery.

- On a supported Mac, select Apple Speech Recognition and start a fresh voice conversation. Speak several turns and confirm each finalized utterance appears and submits promptly without downloading an STT model.
- Change to a locale Apple Speech Recognition does not support, or run on a platform where it is unavailable. The Apple option is hidden and Parakeet remains available.

## Testing

Manually confirmed prompt Apple STT submission and multi-turn voice conversation in a separate macOS development build.


src-tauri/swift/BerdAirPodsBridge/Sources/BerdAirPodsBridge/BerdMacSpeechBridge.swift
```diff
@@ -0,0 +1,621 @@
...
+    case .supported: "available"
+    case .downloading: "downloading"
+    case .installed: "installed"
+    @unknown default: "unknown"
+    }
+}
+
+@available(macOS 26.0, *)
+private func status(for requested: Locale) async -> SpeechStatus {
+    guard SpeechTranscriber.isAvailable else {
```
morgmart, line 156: 🤖 **P1 · Declare and request speech permission** (blocking)

The new bridge begins querying SpeechTranscriber directly, but this PR does not add NSSpeechRecognitionUsageDescription to src-tauri/Info.plist and contains no Speech framework authorization request. Apple's Speech framework requires the usage-description key and user authorization before recognition is accessed; relying only on the existing microphone permission does not satisfy that separate privacy boundary.

**User effect:** On a fresh installation, Apple speech recognition cannot reliably start: macOS may deny the operation, and accessing the protected API without its usage-description key can terminate the app instead of presenting a consent prompt.

**Recommended fix:** Add a localized NSSpeechRecognitionUsageDescription to the macOS bundle, request and handle Speech Recognition authorization before reporting the backend as usable, and expose denied/restricted/not-determined states through MacSpeechStatus so settings and call startup give an actionable result.

**Test:** Add a discriminating status/authorization test covering authorized, denied, restricted, and not-determined states, plus a packaging assertion that the built macOS Info.plist contains NSSpeechRecognitionUsageDescription; verify a first-use flow requests consent before creating SpeechSession.

src-tauri/swift/BerdAirPodsBridge/Sources/BerdAirPodsBridge/BerdMacSpeechBridge.swift
```diff
@@ -0,0 +1,621 @@
...
+    guard let locale = await resolve(requested) else {
+        return SpeechStatus(
+            supported: true, locale: nil, localeSupported: false,
+            modelStatus: "unsupported", ready: false
+        )
+    }
+    let transcriber = speechTranscriber(for: locale)
+    let format = await SpeechAnalyzer.bestAvailableAudioFormat(compatibleWith: [transcriber])
+    let inventory = await AssetInventory.status(forModules: assetModules(for: locale))
+    let ready = format != nil
```
morgmart: 🤖 **P1 · Use inventory for model readiness** (blocking)

status(for:) reads AssetInventory.status but then sets ready solely from whether SpeechAnalyzer.bestAvailableAudioFormat returns a format. Audio-format compatibility is not evidence that the locale's speech asset is installed; the inventory value already distinguishes supported/downloading/installed, yet it is ignored when computing ready and modelStatus.

**User effect:** A Mac with a supported locale but no downloaded speech asset can be reported as installed. Berd then hides the download action, selects Apple speech by default, and lets the user start a call that cannot transcribe successfully.

**Recommended fix:** Derive ready from both a compatible format and inventory == .installed, and preserve the inventory-derived modelStatus whenever the asset is not installed. Keep startup's modelInstalled guard aligned with that same source of truth.

**Test:** Add a discriminating bridge/status test where bestAvailableAudioFormat is non-nil while AssetInventory reports .supported or .downloading, asserting modelInstalled/ready is false and the download path remains available; also cover .installed with a compatible format as ready.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Static review of the exact three-dot comparison found two blocking macOS speech setup defects: the app bundle does not declare/request Speech Recognition permission before using the framework, and model readiness is derived from audio-format availability rather than the asset inventory's installed state. Supplied GitHub evidence was inspected: eight check runs completed successfully, while the combined commit status was pending at capture time. No prior review threads were supplied.

src-tauri/swift/BerdAirPodsBridge/Sources/BerdAirPodsBridge/BerdMacSpeechBridge.swift
johnmatthewtennant: 🤖 The proposed inventory requirement would regress a working configuration. Apple documents bestAvailableAudioFormat as considering assets installed on the device and returning nil when additional assets are required. AssetInventory can lag shared system assets; on the tested en-CA system it reported supported while SpeechTranscriber was already working. The pushed change preserves the compatible-format signal as authoritative, avoids an unnecessary installation request in that state, and adds regressions for both the lagging-inventory and genuinely unavailable-format cases. Reference: https://developer.apple.com/documentation/speech/speechanalyzer/bestavailableaudioformat(compatiblewith:)

src-tauri/swift/BerdAirPodsBridge/Sources/BerdAirPodsBridge/BerdMacSpeechBridge.swift
johnmatthewtennant: 🤖 This finding does not apply to the API used here. Apple’s authorization documentation explicitly says the process applies only to legacy SFSpeechRecognizer; SpeechAnalyzer transcriber modules are on-device. Apple’s SpeechAnalyzer WWDC sample requests microphone permission only, which Berd already declares and handles. The current implementation intentionally does not introduce SFSpeechRecognizer authorization or NSSpeechRecognitionUsageDescription. References: https://developer.apple.com/documentation/speech/asking-permission-to-use-speech-recognition and https://developer.apple.com/videos/play/wwdc2025/277/?time=750

src/features/voice-conversation/lib/voiceInputPreference.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (listeners.size === 0) {
+      removeWindowListeners?.();
+      removeWindowListeners = undefined;
+    }
+  };
+}
+
+export function setVoiceInputBackend(backend: VoiceInputBackend): void {
+  if (typeof window === "undefined") return;
+  try {
```
morgmart, line 90: 🤖 **P2 · Keep preference when storage fails** (non-blocking)

setVoiceInputBackend catches a localStorage write failure and dispatches a change event, but useSyncExternalStore immediately reads the unchanged value back from localStorage. The event detail is never consumed and there is no in-memory snapshot, so the comment's promised usable renderer fallback does not exist.

**User effect:** When browser storage is unavailable or rejects writes, choosing Apple speech recognition or Parakeet appears to do nothing and the backend remains on its previous/default value for the rest of the session.

**Recommended fix:** Maintain a module-level in-memory preference that is updated before attempting persistence, have the external-store snapshot prefer that current-session value, and synchronize it when valid cross-window storage events arrive.

**Test:** Mock localStorage.setItem to throw, select the other backend, and assert the hook updates immediately and retains that value on rerender; also verify a later valid storage event replaces the in-memory value.

src-tauri/swift/BerdAirPodsBridge/Package.swift
```diff
@@ -19,7 +19,15 @@ let package = Package(
...
             dependencies: ["BerdObjCExceptionCatch"],
-            linkerSettings: [.linkedFramework("AVFAudio")]
-        )
+            linkerSettings: [
+                .linkedFramework("AVFAudio"),
+                .linkedFramework("AVFoundation"),
+                .linkedFramework("Speech"),
+            ]
+        ),
+        .testTarget(
```
morgmart: 🤖 **P2 · Run the Swift regression tests** (non-blocking)

The PR adds a Swift test target and model-readiness tests, but no just recipe, package script, or GitHub workflow invokes `swift test` for this package. The existing macOS Tauri lane runs Rust checks/tests/clippy and can build the linked library without executing this test target, so these assertions are outside the validation gate.

**User effect:** The native model-readiness behavior can regress while every required repository check remains green, despite the PR appearing to add regression coverage for it.

**Recommended fix:** Add a repository-owned Swift-package test command and invoke it from the macOS CI lane (and the appropriate local validation gate), so the new test target is exercised on every relevant change.

**Test:** Verify the CI command executes BerdAirPodsBridgeTests and fails when either readiness assertion is intentionally inverted; keep the command macOS-only because the package imports Apple frameworks.

morgmart (commented):
🤖 Automated code review

COMMENT. Fresh static review of the exact three-dot comparison found two new non-blocking issues: the new voice-input preference cannot actually fall back to in-memory state when localStorage writes fail, and the newly added Swift regression tests are not wired into any repository or CI validation command. The two prior blocking automation findings were suppressed because their resolved threads contain substantive replies from the PR author. Supplied GitHub evidence was structurally valid and inspected: three checks had succeeded, three were in progress, and the combined status was pending at capture time; required checks still govern merge readiness.

Deterministic publication result: 0 blocking and 2 non-blocking finding(s) publishable; 2 duplicate(s) suppressed.

src/features/voice-conversation/lib/voiceInputPreference.ts
johnmatthewtennant: 🤖 Fixed in d1dd7966. The renderer now updates a module-level backend before persistence, falls back to it when storage reads or writes fail, and replaces it when a valid cross-window storage event arrives. Focused tests cover both a rejected write remaining reactive and a later storage event superseding the fallback.

src-tauri/swift/BerdAirPodsBridge/Package.swift
johnmatthewtennant: 🤖 Addressed in d1dd7966 by removing the ungated Swift test target and helper rather than expanding repository CI for this small regression. The production bestAvailableAudioFormat installation fast path remains, while the PR no longer presents tests that required checks do not execute.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison found no new publishable findings. The full changed surface was rechecked across Apple speech capability/model setup and installation, backend selection and persistence, voice-call startup/shutdown/transcript delivery, mute and lifecycle races, settings UI accessibility/responsiveness, English and Spanish localization, navigation/readiness guards, error and never-completes behavior, tests, design-system use, and project rules. The two prior non-blocking findings are fixed at this SHA: failed localStorage writes now retain reactive in-memory state with focused tests, and the ungated Swift test target was removed. Two older proposed findings remain suppressed because their resolved threads contain substantive replies from the PR author. Supplied GitHub evidence was structurally valid and inspected: seven checks had succeeded, Transcript virtualization was in progress, and the combined status was pending at capture time; required checks still govern merge readiness.

PR #182, opened 2026-08-24
fix: discover configured Databricks and Ollama models

johnmatthewtennant:
## Summary

Berd could omit working Databricks and Ollama models from the model picker during startup. Databricks's shared OAuth credential is reported under an alias that was lost when Berd mapped Goose's setup catalog, while Ollama's positive configured status was excluded because it had no credential or managed endpoint.

This bug fix separates read-only model inventory discovery from automatic default-provider recovery. Startup may probe providers that Goose reports configured, plus credentialed OAuth aliases, without treating ambient configuration as permission to persist a new default. Curated model-provider aliases now survive setup-catalog mapping, and a secrets-storage failure no longer prevents status-based discovery.

With Ollama running at its standard local endpoint and a Databricks OAuth credential cached in Goose, launch a BYO-provider Berd build and open the Goose model picker. Ollama's installed models and the available Databricks models should appear alongside OpenAI models. Existing default-provider selection should remain unchanged.

### Related issue

None found.

### DIFF
--- a/src/features/providers/defaultProviderConfig.ts
+++ b/src/features/providers/defaultProviderConfig.ts
@@ -1,15 +1,11 @@
 import { getClient } from "@/shared/api/acpConnection";
-import {
-  checkAllProviderStatus,
-  getProviderConfig,
-  listProviderSecrets,
-} from "./api/credentials";
+import { checkAllProviderStatus, listProviderSecrets } from "./api/credentials";
 import { getModelProviders } from "./providerCatalog";
 import {
   getCredentialedProviderIds,
-  getProviderConnectionEvidence,
-  hasMeaningfulSavedSettings,
+  isCredentialedProvider,
 } from "./lib/providerConnectionPolicy";
+import { connectedModelProviderIds } from "./lib/providerState";
 import { useProviderModelCacheStore } from "./stores/providerModelCacheStore";
 import { useDefaultProviderReadinessStore } from "./stores/defaultProviderReadinessStore";
 import { useRuntimeConfigStore } from "@/shared/runtime-config/runtimeConfigStore";
@@ -25,12 +21,11 @@ import {
 } from "@/features/runtime-config/defaults";
 
 /**
- * Providers eligible for model discovery and default-provider recovery. A
- * stored credential is authoritative even when Goose's field-only status
- * probe cannot see an OAuth cache. Otherwise Goose must report the provider
- * configured and Berd must find deliberate setup evidence: a user-created
- * custom provider, a distribution-managed endpoint, or a meaningful saved
- * non-secret value. Untouched defaults and ambient readiness do not qualify.
+ * Providers eligible for default-provider recovery: Goose reports them
+ * configured AND they have deliberate setup evidence: a stored Goose
+ * credential, a user-created custom provider, or a distribution-managed
+ * endpoint. Merely appearing in runtime inventory or having ambient/default
+ * setup values does not satisfy the readiness gate.
  */
 export async function getIntentionalConfiguredProviderIds(
   statuses: Awaited<ReturnType<typeof checkAllProviderStatus>>,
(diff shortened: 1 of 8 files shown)


src/features/providers/defaultProviderConfig.ts
```diff
@@ -54,6 +57,36 @@ export async function getIntentionalConfiguredProviderIds(
...
+  statuses: Awaited<ReturnType<typeof checkAllProviderStatus>>,
+): Promise<string[]> {
+  const configuredIds = new Set(
+    statuses
+      .filter((status) => status.isConfigured)
+      .map((status) => status.providerId),
+  );
+  let credentialedIds: ReadonlySet<string> = new Set();
+  try {
+    credentialedIds = getCredentialedProviderIds(await listProviderSecrets());
```
morgmart: 🤖 **P1 · Bound secret-store discovery** (blocking)

The new discovery path awaits listProviderSecrets() before it returns any providers. The catch preserves status-based discovery only when secure storage rejects; if the ACP request hangs or never settles, getModelDiscoveryProviderIds never returns, refreshAllModelProviders is never reached, and the existing outer catch cannot run. This contradicts the stated degraded-path behavior and puts the status-based Ollama fix behind an unrelated secret-store lifecycle dependency.

**User effect:** When secure storage or its ACP request stalls, configured status-based providers such as Ollama never refresh, so their models can remain absent from the model picker for the entire app session even though Goose already reported the provider configured.

**Recommended fix:** Bound the secret lookup independently and allow status-based provider discovery to proceed when that bound expires. If a late credential result should still add OAuth-only providers, handle it as a separate idempotent follow-up refresh rather than keeping the status-based refresh blocked.

**Test:** Add a discriminating test with GooseUnstableProvidersSecretsList returning a never-settling promise; assert that discovery still resolves within the chosen bound with the status-configured provider. If late credentials are supported, also resolve the deferred promise afterward and assert the OAuth alias is refreshed once without replacing or duplicating the status-based result.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The exact three-dot comparison was reviewed at head e5013a947b309150e48f7984dbb063266c41ccf0 using static inspection only. One blocking lifecycle issue remains: the new read-only discovery path waits without a bound for secret storage before returning status-based providers, so the promised fallback does not engage when that request never completes. Supplied GitHub evidence was inspected: all listed check runs for the exact head completed successfully, while the combined commit status is pending with no legacy statuses; required checks still govern merge readiness.

src/features/providers/defaultProviderConfig.ts
johnmatthewtennant: 🤖 Fixed in bf454581. Secret-store lookup is now independently bounded at three seconds, after which status-configured providers continue to discovery. Added a fake-timer regression test with a never-settling secrets request; 239 provider/startup tests and `just check` pass, and the full pre-push fmt/frontend/Tauri/clippy gate passed.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static inspection of the full exact three-dot comparison at bf454581349ad3d28cc99ffe47c84e161653dd50 found no publishable issues. The prior secret-store never-completes issue is fixed by an independently bounded lookup and a discriminating never-settling-promise test; its resolved thread also has a substantive human-account reply. Final self-check covered the configured-provider and OAuth-alias discovery flows, preservation of default-provider consent boundaries, startup async failure/timeout/lifecycle/race behavior, test honesty, accessibility, localization, navigation, design-system/project rules, duplicate overlap, and blocking-evidence quality. This PR changes no visual, navigational, accessibility, or localized UI surface. Supplied GitHub evidence is structurally valid: all listed check runs for the exact head completed successfully, while the combined legacy commit status remains pending with no statuses; required checks still govern merge readiness.

PR #189, opened 2026-08-24
perf(dev): skip current setup during just dev

johnmatthewtennant:
## Summary

`just dev` currently reruns dependency installation, SDK compilation, hook setup, and managed Goose setup before every launch. It now verifies pnpm and SDK inputs with worktree-local success stamps, rebuilds only stale outputs, and delegates managed Goose freshness to its existing stamped ensure path. Dependency installs and SDK rebuilds invalidate their prior success stamps before work begins and publish replacements only after success, so interrupted repairs retry on the next launch. SDK freshness also requires every package-exported output. Explicit `just setup` continues to force the complete setup workflow.

From a prepared checkout, force the setup path and then exercise the warm development path:

```bash
just _setup-dev-deps
just _ensure-dev-deps
```

The second command reports:

```text
pnpm dependencies are current; skipping install.
@aaif/goose-sdk is current; skipping build.
```

Changing a workspace manifest, lockfile, SDK schema, generator, configuration, or source invalidates the corresponding output on the next run. A failed dependency install, missing package-exported SDK output, or failed SDK rebuild also leaves that prerequisite stale for automatic repair.

### DIFF
--- a/scripts/ensure-dev-deps.sh
+++ b/scripts/ensure-dev-deps.sh
@@ -86,6 +86,7 @@ if [[ "$force" == "1" ]] \
   || [[ ! -f "$installed_lock" ]] \
   || ! cmp -s "$repo_root/pnpm-lock.yaml" "$installed_lock"; then
   echo "Preparing pnpm dependencies."
+  rm -f "$dependency_stamp"
   (cd "$repo_root" && "$pnpm_bin" install)
   dependency_inputs="$(dependency_inputs_hash)"
   write_stamp "$dependency_stamp" "$dependency_inputs"
--- a/scripts/ensure-dev-deps.sh
+++ b/scripts/ensure-dev-deps.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(cd "$script_dir/.." && pwd)"
+force=0
+
+if [[ "${1:-}" == "--force" ]]; then
+  force=1
+  shift
+fi
+
+if [[ $# -ne 0 ]]; then
+  echo "Usage: scripts/ensure-dev-deps.sh [--force]" >&2
+  exit 2
+fi
+
+pnpm_bin="${PNPM_BIN:-pnpm}"
+dependency_stamp="$repo_root/node_modules/.berd-dev-dependency-inputs.sha256"
+installed_lock="$repo_root/node_modules/.pnpm/lock.yaml"
+sdk_stamp="$repo_root/sdk/dist/.berd-dev-build-inputs.sha256"
+
+hash_files() {
+  local path
+  for path in "$@"; do
+    if [[ -d "$path" ]]; then
+      find "$path" -type f -print
+    elif [[ -f "$path" ]]; then
+      printf '%s\n' "$path"
+    fi
+  done \
+    | LC_ALL=C sort -u \
+    | while IFS= read -r path; do
+        shasum -a 256 "$path"
+      done \
+    | shasum -a 256 \
+    | awk '{print $1}'
+}
+
+stamp_matches() {
+  local stamp="$1"
+  local expected="$2"
+  [[ -f "$stamp" ]] && [[ "$(<"$stamp")" == "$expected" ]]
+}
+
+write_stamp() {
+  local stamp="$1"
+  local value="$2"
+  local temp_stamp
+  mkdir -p "$(dirname "$stamp")"
+  temp_stamp="$(mktemp "${stamp}.XXXXXX")"
+  printf '%s\n' "$value" >"$temp_stamp"
+  mv "$temp_stamp" "$stamp"
+}
+
+dependency_inputs_hash() {
+  hash_files \
+    "$repo_root/package.json" \
+    "$repo_root/pnpm-lock.yaml" \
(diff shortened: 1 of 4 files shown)


scripts/ensure-dev-deps.sh
```diff
@@ -0,0 +1,103 @@
...
+  write_stamp "$dependency_stamp" "$dependency_inputs"
+else
+  echo "pnpm dependencies are current; skipping install."
+fi
+
+sdk_inputs="$(sdk_inputs_hash)"
+
+if [[ "$force" == "1" ]] \
+  || ! stamp_matches "$sdk_stamp" "$sdk_inputs" \
+  || [[ ! -f "$repo_root/sdk/dist/index.js" ]] \
```
morgmart: 🤖 **P2 · Partial SDK builds pass freshness** (blocking)

The freshness check retains any previous success stamp while rebuilding and validates only dist/index.js and dist/index.d.ts. A forced or missing-output rebuild can emit those two files and then fail, or another required emitted file such as the package-exported dist/resolve-binary.js can be absent; the next invocation still sees the old matching input stamp plus both sentinels and reports the SDK as current.

**User effect:** After a failed or interrupted SDK repair, a developer can rerun just dev and be told the SDK is current, only to hit downstream missing-module or stale-build failures instead of getting the promised automatic recovery.

**Recommended fix:** Invalidate the SDK completion stamp before starting a rebuild and publish a new stamp only after success, while validating a complete manifest of expected emitted files or replacing dist atomically rather than checking only the root entry-point artifacts.

**Test:** Start from a successful warm fixture, remove a required non-root output such as dist/resolve-binary.js and assert the guard rebuilds. Also make the fake SDK build recreate index.js and index.d.ts but exit nonzero, then assert the following invocation retries instead of skipping; both cases fail with the current implementation.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The incremental dependency guard can accept a partial or failed SDK build as current, so the automatic repair path is not reliable. The supplied GitHub evidence is structurally valid and all nine check runs for the exact head SHA completed successfully; required checks still govern merge readiness.

scripts/ensure-dev-deps.sh
johnmatthewtennant: 🤖 Accepted and fixed in 8afe43a2. The guard now removes the SDK success stamp before rebuilding, writes it only after a successful build, and requires all four package-exported outputs. Added regressions for a missing non-root export and for retrying after a failed forced rebuild; the release-script suite passes 98/98 and the full pre-push gate passes.

scripts/ensure-dev-deps.sh
```diff
@@ -0,0 +1,110 @@
...
+}
+
+dependency_inputs="$(dependency_inputs_hash)"
+
+if [[ "$force" == "1" ]] \
+  || ! stamp_matches "$dependency_stamp" "$dependency_inputs" \
+  || [[ ! -f "$installed_lock" ]] \
+  || ! cmp -s "$repo_root/pnpm-lock.yaml" "$installed_lock"; then
+  echo "Preparing pnpm dependencies."
+  (cd "$repo_root" && "$pnpm_bin" install)
```
morgmart, line 90: 🤖 **P2 · Failed installs retain success stamp** (blocking)

The dependency repair path leaves the previous dependency success stamp in place while pnpm install runs. If an install updates node_modules/.pnpm/lock.yaml to match the tracked lockfile and then fails—for example during later linking or lifecycle work—the next invocation can see unchanged hashed inputs, the old matching stamp, and matching lockfiles, then report dependencies as current without retrying the failed operation.

**User effect:** After a failed setup or dependency repair, a developer can retry just dev and have it skip installation, then encounter confusing missing-package or incomplete-install failures later in startup instead of recovering automatically.

**Recommended fix:** Remove the dependency success stamp immediately before running pnpm install, and write a replacement only after the install succeeds, matching the SDK repair path's completion-record lifecycle.

**Test:** Start from a successful warm fixture, force an install whose fake pnpm updates the internal lock and then exits nonzero, and assert that the following normal invocation runs install again. The current implementation skips that retry.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The prior SDK freshness issue is fixed with success-stamp invalidation, complete package-export checks, and discriminating regressions. One analogous dependency-repair path still retains an old success stamp across a failed install, allowing a later launch to skip the needed retry. All 10 supplied GitHub check runs for the exact head SHA completed successfully; the combined commit status is pending with no status contexts, so required checks still independently govern merge readiness.

scripts/ensure-dev-deps.sh
johnmatthewtennant: 🤖 Accepted and fixed in d1831083. The dependency guard now removes its success stamp immediately before pnpm install and writes a replacement only after success. Added a warm-fixture regression where the fake install updates the internal lock and then fails; the following normal invocation retries installation. Release scripts pass 99/99 and the full isolated pre-push gate passes.

morgmart (approved):
🤖 Automated code review

APPROVE. The full exact PR comparison now invalidates dependency and SDK success stamps before repair, restores them only after successful completion, verifies all package-exported SDK outputs, and includes discriminating regressions for both interrupted repair paths. Both prior automated findings are fixed, so they are not findings or duplicate suppressions. All 10 supplied GitHub check runs for the exact head SHA completed successfully; the combined commit status is pending with no status contexts, so required checks still independently govern merge readiness.

PR #190, opened 2026-08-24
fix(voice): resume false-positive interruptions

johnmatthewtennant:
## Summary

This fixes assistant speech being permanently abandoned when Earshot reports a brief interruption but Apple speech recognition produces no newer finalized transcript. Berd holds the interrupted delivery through the existing transcript settle window and resumes from the first synthesis chunk not known to be complete.

Finalized speech, microphone mute, explicit stop, call or session ownership changes, and voice backend changes discard the remainder. Native playback ownership is released before a resumed stream starts, and resume positions remain absolute so repeated interruptions can replay the current chunk without rewinding into earlier completed chunks.

### Related issue

None found.

### Testing

On macOS, start a voice conversation with Pocket TTS and request a long response. Briefly interrupt it without producing finalized STT. The response resumes after the settle window. Interrupt the resumed response repeatedly: the current incomplete chunk may replay, but playback does not move back into earlier completed chunks and does not report that Pocket playback is already active.

Speak a real utterance that produces finalized STT, mute the microphone, or end the call while speech is held. The held remainder stays discarded rather than resuming.

### DIFF
--- a/src/features/voice-conversation/lib/nativeAssistantSpeech.ts
+++ b/src/features/voice-conversation/lib/nativeAssistantSpeech.ts
@@ -50,7 +50,7 @@ type ActiveUtterance = {
   targetSpans: SpeechTargetSpan[];
   text: string;
   finishing: boolean;
-  nativeStreamStarted: boolean;
+  nativeStartInvoked: boolean;
   interruptionRequested: boolean;
   resumptionDiscarded: boolean;
   interruptionFallback: ReturnType<typeof setTimeout> | null;
@@ -183,7 +183,7 @@ function recordPlaybackNotice(
   estimate?: SpeechDeliveryEstimate,
   interruptionCause: InterruptionCause = "voiceStopped",
 ) {
-  const noticeKey = `${sessionId}\0${key}\0${status}`;
+  const noticeKey = `${sessionId}\0${key}`;
   const excerpt = text.length > 500 ? `${text.slice(0, 497).trimEnd()}…` : text;
   const outcome =
     status === "interrupted"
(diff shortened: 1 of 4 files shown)


src/features/voice-conversation/lib/nativeAssistantSpeech.ts
```diff
@@ -619,17 +728,17 @@ function interruptActiveUtterance(
...
   if (utterance && !utterance.interruptionRequested) {
     utterance.interruptionRequested = true;
     utterance.interruptionCause = cause;
-    if (terminalEventExpected) utterance.onInterrupted();
   }
   if (utterance && !terminalEventExpected) {
     finalizeInterruptedUtterance(
       utterance,
       utterance.interruptionCause ?? cause,
+      awaitTerminalDelivery && cause === "userSpeaking",
```
morgmart, line 741: 🤖 **P1 · Resume after terminal timeout** (blocking)

The newly added resumability flag is passed only when interruption is finalized before native startup. When native startup has occurred, the code waits for a terminal event; if that event never arrives, the one-second fallback calls finalization without allowing resume. That leaves the same false-positive interruption permanently abandoned in the exact never-completes case the fallback is intended to bound. The changed test titled "bounds a missing native terminal event and allows the next reply" confirms this behavior by expecting the first reply to remain interrupted rather than resume.

**User effect:** After a brief false-positive interruption, an assistant response can stop permanently whenever the native backend loses or delays its terminal event beyond one second, even though no finalized user speech arrived.

**Recommended fix:** When the terminal wait times out for a user-speaking interruption, finalize through the same resumable path used by a received interrupted event, while preserving the existing generation, transcript-ownership, mute, stop, and idempotence guards. Do not treat the presentation timeout as cancellation of the remainder.

**Test:** Change the missing-terminal regression test to simulate user speaking then becoming idle with no native terminal event, advance through both the one-second terminal bound and transcript-settle window, and assert that a replacement stream starts from the first incomplete synthesis segment. Also assert that a late terminal event cannot start a second replacement or duplicate notices.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The false-positive interruption flow resumes after a normal native terminal event, but the bounded missing-terminal fallback still permanently abandons the interrupted remainder. The supplied GitHub evidence was inspected: nine check runs at the exact head SHA completed successfully, while GitHub's combined status was pending; required checks still govern merge readiness.

src/features/voice-conversation/lib/nativeAssistantSpeech.ts
johnmatthewtennant: 🤖 Fixed in 5be02ae7. When the terminal bound expires, the renderer probes the same native backend again and resumes only if that probe reports the old playback token is gone. A still-active token or failed probe remains non-resumable, preserving single-stream ownership. The regression now covers resumption after released ownership, refusal while ownership remains active, and a late terminal event without duplicate streams or notices.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static inspection of the full exact three-dot comparison found no new publishable findings. The changed flows cover false-positive interruption resumption before and after native startup, safe resume cutoffs across segments and repeated interruptions, ownership release, terminal timeout and late-event races, genuine finalized speech, microphone mute, explicit stop, session/backend lifecycle changes, held/rewritten/appended text, failure notices, and Pocket/Siri parity. No user-interface surface changed, so accessibility, localization, navigation, consent, responsive layout, and design-system concerns are not implicated. Changed tests are behaviorally discriminating for the principal async and lifecycle paths. The supplied evidence is structurally valid: eight checks at the exact head SHA completed successfully, while GitHub's combined status was pending; required checks still govern merge readiness.

PR #194, opened 2026-08-24
feat(voice): add advanced VAD settings

johnmatthewtennant:
## Summary

Adds an Advanced voice detection picker to Voice settings for the native VAD thresholds Berd supports tuning:

- Speech detection sensitivity controls how easily Berd reports that the user started speaking.
- Interruption sensitivity controls how easily speech interrupts Berd while it is speaking.
- Friendly presets expose their exact threshold values and preserve the current defaults (0.50 for speech detection and 0.65 for interruptions).
- Preferences apply when the next voice conversation starts and are shared by the Apple and Parakeet input paths.

1. Open Settings, then Voice, then Advanced.
2. Choose Less sensitive for speech detection and More sensitive for interruptions.
3. Close and reopen Advanced. Both choices remain selected.
4. Choose Prevent feedback. Reopen Advanced and confirm interruption sensitivity is hidden because interruptions are disabled, while speech detection remains available.
5. Select Reset to defaults and confirm the values return to More sensitive (0.50) and Balanced (0.65).

## Screenshots

Advanced voice detection dialog


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison 4fe4bcb8b1a5bf069288f0cf7372ddaa03362c3c...32c92ae54fc35320d87568e0dd78b5379e7f86e6 found no publishable issues. Final self-check covered the advanced voice-detection settings flow, preference persistence and reset behavior, next-conversation application, both native input backends, accessibility, English and Spanish localization, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, design-system and project rules, duplicate overlap, and blocking-evidence quality. Supplied GitHub evidence was inspected: clean-room install, DCO, and Linux Rust checks passed; frontend and Windows/macOS Rust checks were in progress at capture time, so required checks still govern merge readiness.

**Pending checks:** 4 check(s) are not complete.

morgmart (commented):
🤖 Automated code review

COMMENT. Fresh static review of the exact three-dot comparison 4fe4bcb8b1a5bf069288f0cf7372ddaa03362c3c...5c795371bf620420a24a414e0f1d118ecd68987d found one non-blocking test-honesty issue and no blocking issues. The final self-check covered the settings dialog, threshold selection/reset/persistence, next-conversation application across Parakeet and Apple input, accessibility, English and Spanish localization, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, design-system and project rules, duplicate overlap, and evidence plus user effect for severity. Supplied GitHub evidence was structurally valid and inspected; all listed checks for this SHA completed successfully. Required checks still independently govern merge readiness.

src/features/voice-conversation/lib/voiceInterruptionPreference.test.ts
```diff
@@ -43,79 +45,54 @@ describe("voice interruption preference", () => {
...
+        mode: "allowInterruptions",
+        sensitivity: "more",
+        speechSensitivity: "less",
+      }),
     );
 
     expect(getVoiceInterruptionPreference()).toEqual({
       mode: "allowInterruptions",
       sensitivity: "more",
-    });
```
morgmart, line 85: 🤖 **P2 · Restore failed-write storage-event coverage** (non-blocking)

This comparison deletes the only discriminating test for the recovery path where a localStorage write fails and a later cross-window storage event must replace the renderer's volatile preference. That lifecycle is especially relevant because the production change now also clears volatile state from the subscribed storage-event handler. The remaining tests cover a later direct storage read with a different persisted value, but they do not exercise an actual storage event or prove a mounted useSyncExternalStore consumer updates, so the event/listener behavior could regress while the suite remains green.

**User effect:** After a temporary settings-storage failure, a Voice settings window could keep showing or applying stale sensitivity values even after another window resets or updates them, with no regression test catching it.

**Recommended fix:** Restore a focused storage-event regression test using a localStorage spy setup compatible with the current test environment. Keep the hook mounted, force setItem to fail, restore storage, dispatch a same-key or clear storage event, and assert both the hook snapshot and getVoiceInterruptionPreference adopt the external/default value.

**Test:** Mount useVoiceInterruptionPreference, force setItem to throw while selecting non-default values, restore setItem, dispatch StorageEvent for goose:voice-interruption-preference (including the null/clear case), and verify the mounted hook and getter leave the volatile value. The test should fail if clearVolatilePreference is removed from the storage-event path.

PR #202, opened 2026-08-25
feat(migration): move onboarding import into Berd

matt2e:
## Summary

- Replace Berd's renderer calls to Goose's deprecated onboarding import scan/apply methods with a native Tauri import command.
- Discover only fixed Goose and Claude Desktop locations, parse bounded config files, and copy legacy skills into the portable personal skills directory without overwriting collisions.
- Keep secrets and extension payloads out of renderer IPC by applying them through a backend-only ACP connection, while preserving existing extension entries and returning non-sensitive warnings for recoverable import failures.

### Related issue

https://github.com/aaif-goose/goose/issues/11451

### Testing

- `just check`
- `just tauri-check`
- `just clippy`
- `just _tauri-cargo-unix test --lib commands::onboarding_import::tests` (16 passed)
- `pnpm vitest run src/features/migration/runMigration.test.ts` (passed)
- `just test` (7021 passed, 1 skipped; one unrelated telemetry/localStorage test fails in this environment)
- `just _tauri-cargo-unix test --lib` (828 passed; seven unrelated base-branch failures: five layout default-zoom assertions and two Goose runtime-provider fixture assertions)

No UI layout or visual behavior changes.

### DIFF
--- a/src-tauri/src/commands/onboarding_import.rs
+++ b/src-tauri/src/commands/onboarding_import.rs
@@ -1,0 +1,1412 @@
+//! Native discovery and parsing for Berd's silent first-boot import.
+//!
+//! The renderer receives a bounded, non-secret plan. Imported secrets are
+//! applied through a native ACP connection so Goose remains authoritative and
+//! plaintext never crosses into the webview. No renderer-provided filesystem
+//! path is accepted by this module.
+
+use std::collections::{BTreeMap, HashSet};
+use std::fs;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+use futures_util::{SinkExt, StreamExt};
+use serde::{Deserialize, Serialize};
+use serde_json::Value as JsonValue;
+use tauri::{AppHandle, Manager};
+use tokio_tungstenite::tungstenite::Message;
+use yaml_serde::Value as YamlValue;
+
+use crate::services::goose_config;
+
+const MAX_CONFIG_BYTES: u64 = 1024 * 1024;
+const MAX_SKILLS_PER_ROOT: usize = 256;
+const MAX_SKILL_ROOT_ENTRIES: usize = 1024;
+const MAX_ENTRIES_PER_SKILL: usize = 512;
+const MAX_FILE_BYTES: u64 = 2 * 1024 * 1024;
+const MAX_SKILL_BYTES: u64 = 16 * 1024 * 1024;
+const MAX_SECRET_UPDATES: usize = 256;
+const MAX_EXTENSION_UPDATES: usize = 256;
+const DEFAULT_EXTENSION_TIMEOUT_SECONDS: u64 = 300;
+const ACP_REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
+const ACP_CLOSE_TIMEOUT: Duration = Duration::from_secs(1);
+
+#[derive(Debug, Clone, Default, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OnboardingImportPlan {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub provider_defaults: Option<ImportedProviderDefaults>,
+    pub imported_skills: u32,
+    pub skipped_skills: u32,
+    pub warnings: Vec<String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ImportedProviderDefaults {
+    pub provider_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub model_id: Option<String>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct ImportedSecretUpdate {
+    pub key: String,
+    pub value: JsonValue,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct ClaudeDesktopConfig {
+    #[serde(default)]
+    mcp_servers: BTreeMap<String, ClaudeMcpServer>,
+}
+
+#[derive(Debug, Deserialize)]
+struct ClaudeMcpServer {
+    command: Option<String>,
+    #[serde(default)]
+    args: Vec<String>,
+    #[serde(default)]
+    env: BTreeMap<String, String>,
+}
+
+#[derive(Debug, Default)]
+struct SkillCopyCounts {
+    imported: u32,
+    skipped: u32,
+}
+
+struct SkillCopyOutcome {
+    counts: SkillCopyCounts,
+    warnings: Vec<String>,
+}
+
+struct PreparedOnboardingImport {
+    plan: OnboardingImportPlan,
+    secret_updates: Vec<ImportedSecretUpdate>,
+    extensions: BTreeMap<String, JsonValue>,
+}
+
+struct ImportLocations {
+    goose_configs: Vec<PathBuf>,
+    claude_configs: Vec<PathBuf>,
+    personal_skills_root: PathBuf,
+}
+
+#[tauri::command]
+pub async fn prepare_onboarding_import(app: AppHandle) -> Result<OnboardingImportPlan, String> {
+    let active_goose_config = goose_config::config_path()?;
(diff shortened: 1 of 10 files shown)


src-tauri/src/commands/onboarding_import.rs
```diff
@@ -0,0 +1,1412 @@
...
+        };
+        let mut changed = false;
+        for (key, extension) in imported_extensions {
+            if !merged.contains_key(key) {
+                merged.insert(key.clone(), extension.clone());
+                changed = true;
+            }
+        }
+        if changed {
+            let _ = send_acp_request(
```
morgmart, line 351: 🤖 **P1 · Preserve concurrent extension changes** (blocking)

The import reads the entire current extensions object, merges imported entries into that snapshot, and then replaces the whole object with a later upsert. Any other Goose client or app instance that adds, edits, or removes an extension between the read and this write is silently overwritten. This is a persisted-config migration, so the snapshot-then-write race can lose a concurrent writer's saved intent.

**User effect:** A connection changed in another Berd window or Goose client during first-run migration can unexpectedly revert or disappear.

**Recommended fix:** Avoid replacing a stale full-map snapshot. Apply each missing imported extension through an atomic Goose extension-add/patch operation, or add a backend compare-and-set/merge operation that re-reads and verifies immediately before committing and retries on conflict.

**Test:** Add a discriminating ACP test that changes the server's extensions after the import read but before the import write, then assert both the concurrent change and imported extension survive.

src-tauri/src/commands/onboarding_import.rs
```diff
@@ -0,0 +1,1412 @@
...
+                        imported_extensions.entry(key).or_insert(extension);
+                    }
+                    plan.warnings.extend(config.warnings);
+                }
+                Err(error) => {
+                    plan.warnings.push(format!(
+                        "Skipped Goose configuration import at '{}': {error}",
+                        path.display()
+                    ));
+                    continue;
```
morgmart, line 198: 🤖 **P1 · Import sibling data independently** (blocking)

A missing, unreadable, oversized, or malformed config.yaml immediately continues the loop, so the same legacy directory's valid secrets.yaml and skills directory are never examined. The command still returns a warning, runMigration only writes it to the developer console, and the caller can mark migration complete, turning this recoverable source-file problem into a permanent skipped import. The existing malformed-extension test does not cover malformed top-level config.

**User effect:** A user with a damaged or absent legacy config can silently lose automatic migration of otherwise valid credentials and personal skills, with no in-product indication or later retry.

**Recommended fix:** Treat config, secrets, and skills as independent artifacts: record the config warning but continue scanning sibling secrets and skills. Ensure skipped artifacts remain retryable or surface a user-visible, localized recovery notice before finalizing the one-shot migration.

**Test:** Add a migration test with invalid (and separately absent) config.yaml beside valid secrets.yaml and a valid skill; assert the secret update and skill import still occur and that recoverable failures are not silently finalized.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review of exact range 20c005aefb9fa92ce6209a31475c6fca774d6ded...a627092d25926c08a82154b4cf1c6634034efc29 found 2 blocking findings and 0 non-blocking findings. Final self-check covered the complete first-boot import flow (backup, fixed-location discovery, config/secrets/extensions/skills import, provider defaults, extension disabling, migration completion and retry), accessibility and i18n (no new UI), navigation/consent guards, async timeout/error/never-completes/lifecycle/race behavior, test honesty, project laws and design-system rules, duplicate overlap, and concrete user effect. Supplied GitHub evidence was inspected: all reported head-SHA check runs completed successfully except the skipped close job; the combined commit status remained pending, so required checks still independently govern merge readiness.

src-tauri/src/commands/onboarding_import.rs
```diff
@@ -0,0 +1,1467 @@
...
+        return;
+    }
+    let Ok(directory) = fs::read_dir(target_root) else {
+        return;
+    };
+    for entry in directory.flatten() {
+        if !entry
+            .file_name()
+            .to_string_lossy()
+            .starts_with(SKILL_STAGING_PREFIX)
```
morgmart, line 741: 🤖 **P1 · Do not delete prefix-matched skills** (blocking)

The new retry cleanup treats every directory in the user's personal skills root whose name starts with `.berd-onboarding-` as disposable staging and recursively deletes it. There is no ownership marker, recorded staging path, age check, or exact generated-name validation. A legitimate user directory with that prefix is therefore destroyed, and a second Berd instance can also delete a staging directory while the first instance is actively copying it. This is destructive migration behavior based only on a filename resemblance.

**User effect:** On first boot or a migration retry, a user can permanently lose a personal skill directory, or a concurrent import can be corrupted, without any confirmation or recovery path.

**Recommended fix:** Track staging directories created by this migration with verifiable ownership metadata and clean up only directories proven stale and owned by Berd. Prefer a dedicated private staging parent outside the user-visible skills namespace, and coordinate concurrent import attempts before deleting anything.

**Test:** Add a discriminating test that places a user-authored `.berd-onboarding-custom` skill in the target root and asserts it survives cleanup; also cover two concurrent staging owners so one run cannot remove the other's active directory.

src-tauri/src/commands/onboarding_import.rs
```diff
@@ -0,0 +1,1467 @@
...
+            },
+        }),
+        "initialize secret import",
+        ACP_REQUEST_TIMEOUT,
+    )
+    .await?;
+
+    let mut next_id = 2;
+    for update in secret_updates {
+        let _ = send_acp_request(
```
morgmart, line 300: 🤖 **P1 · Bound the whole startup import** (blocking)

The migration performs as many as 256 secret writes serially and gives every request its own 15-second deadline. Because this command sits on the first-boot migration gate, a slow but responsive Goose peer can keep startup unsettled for roughly 64 minutes before extension work even begins. The per-request timeout prevents one request from waiting forever, but it does not provide a degraded path for the overall operation; the comment explicitly chooses repeated deadlines, and the delayed-peer test only proves that this long-running behavior succeeds.

**User effect:** A user can be unable to reach a ready Home chat for many minutes—or close to an hour—while a silent import waits on sequential backend writes.

**Recommended fix:** Remove network import work from the startup-critical path, or enforce a short aggregate presentation deadline that lets the app become usable while bounded background work continues. Preserve retry progress so an aggregate deadline does not repeatedly restart at the same entry.

**Test:** Add a fake-clock integration test with many just-under-timeout responses and assert the migration gate reaches a usable degraded state within the product-level startup budget while remaining able to resume safely.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Fresh static review of exact range 20c005aefb9fa92ce6209a31475c6fca774d6ded...11929da00a1e3b9e4f0fbe148820cdb64b75186e found 2 new blocking findings and 0 new non-blocking findings; 2 still-present findings were suppressed because they map to existing unresolved automation threads. The final self-check covered the complete first-boot flow (backup, fixed-location discovery, config/secrets/extensions/skills import, provider defaults, extension disabling, marker completion and retry), accessibility and localization (no new UI), navigation and consent guards, async error/timeout/never-completes/lifecycle/race behavior, test honesty, project laws and design-system rules, overlap/duplicate handling, and concrete evidence plus user effect for blocking severity. Supplied GitHub evidence was structurally valid and inspected: clean-room install and DCO passed, while frontend and platform Rust checks were still in progress; required checks independently govern merge readiness.

Deterministic publication result: 2 blocking and 0 non-blocking finding(s) publishable; 2 duplicate(s) suppressed.

morgmart (commented):
🤖 Automated code review

COMMENT. The completed static multi-lens review of exact range 20c005aefb9fa92ce6209a31475c6fca774d6ded...11929da00a1e3b9e4f0fbe148820cdb64b75186e has 0 new blocking and 0 new non-blocking publishable findings. Four current issues are suppressed because each maps to an existing unresolved automation thread. Atlas and Spar separately escalate durable ownership of cross-product onboarding import semantics for Morgan's decision. Vogue confirms this is non-graphical work and no screenshots are required. Supplied GitHub evidence was structurally valid and inspected; required checks independently govern merge readiness.

Deterministic publication result: 0 blocking and 0 non-blocking inline finding(s) publishable; 4 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

PR #204, opened 2026-08-25
fix(voice): recover from denied microphone access

johnmatthewtennant:
## Summary

Fixes the recovery path when macOS microphone access is turned off. Voice settings stays quiet before the first permission request, then shows a focused warning and System Settings shortcut after access is denied. Starting Voice Conversation keeps the normal first-use prompt, and any microphone capture startup failure returns the user to Voice settings.

1. With microphone access not yet requested, open Settings > Voice. No permission warning appears.
2. Start Voice Conversation and deny the macOS microphone prompt. Voice settings opens and shows the microphone-access warning.
3. Choose Open Microphone Settings, allow access, and return to Berd. The warning clears when the window regains focus.

### Microphone access denied

Voice settings showing the microphone access denied warning


morgmart (commented):
🤖 Automated code review

Reviewed the full exact three-dot comparison at head 9e6d30167fc1f4af1abde3ea10edbc2151778467 with the project code-review and wes-review lenses. Found one non-blocking state-consistency issue in microphone-permission recovery. Supplied GitHub check runs are successful, while the combined commit status is pending; required checks still govern merge readiness. Recommendation: COMMENT.

src/features/voice-conversation/hooks/useMicrophonePermission.ts
```diff
@@ -0,0 +1,62 @@
...
+    const id = ++refreshId.current;
+    try {
+      const next = await getMicrophonePermissionStatus();
+      if (id !== refreshId.current) return;
+      setStatus(next);
+      setError(false);
+    } catch (cause) {
+      if (id !== refreshId.current) return;
+      console.error("Failed to read microphone permission", cause);
+      setError(true);
```
morgmart: 🤖 **P2 · Separate status and settings errors** (non-blocking)

The hook uses one error flag for both permission-status refresh failures and failures to open System Settings. After a denied status has been loaded, a later focus refresh can fail and set error to true while retaining the denied status. VoiceSettings then renders the localized message that says Microphone Settings could not be opened, even though the failed operation was only the background status check (and the settings shortcut may have opened successfully). This makes two independent async operations disagree about what the UI reports.

**User effect:** After returning from System Settings, a transient permission-status read failure can incorrectly tell the user that Berd could not open Microphone Settings, sending them toward the wrong recovery action.

**Recommended fix:** Track permission-read failure separately from open-settings failure (or model each operation's state explicitly), clear only the relevant error on success, and render copy appropriate to the operation that actually failed.

**Test:** Add a hook/component regression test that first loads denied, then rejects the focus-triggered status refresh, and verifies that the settings-open error is not shown; retain the existing test proving that an actual openSettings rejection does show it.

johnmatthewtennant: 🤖 Fixed in d9e4d890. Permission-status refresh failures now only log and preserve the last known status; only failures from opening System Settings set `openSettingsError`. Added coverage for denied → focus refresh rejection while retaining the existing open-settings rejection coverage. Focused tests (20) and `just check` pass.

morgmart (approved):
🤖 Automated code review

Fresh full review completed for the exact three-dot comparison fcbaa72bbab9e92efb3e7f1687be18ca7470368d...d9e4d890f629e0aa46a7b078ead6c38b78c281b6 using the project code-review and wes-review lenses. The previously reported status/settings error-state issue is fixed and covered by a discriminating regression test. Final self-check covered the denied, not-determined, authorized, status-read failure, settings-open failure, capture-start failure, cleanup, focus-refresh, and cross-platform flows; accessibility; English and Spanish localization; navigation and permission consent behavior; async, error, never-completes, lifecycle, and race handling; test honesty; design-system and project rules; duplicate overlap; and severity evidence. No publishable findings remain. Supplied GitHub check runs are successful, while the combined commit status is pending; required checks still govern merge readiness. Recommendation: APPROVE.

PR #205, opened 2026-08-25
Fall back when the selected Siri voice is unavailable

johnmatthewtennant:
## Summary

Use any installed Siri voice when the saved selection is unavailable, and open Voice settings only when no installed Siri voice can be resolved. The saved preference remains intact, while each speech stream receives the current effective fallback across status refreshes, view remounts, and session handoffs.

### Testing

On macOS, select a downloaded Siri voice in Berd, remove that voice in System Settings while another Siri voice remains installed, then start a voice conversation. Berd should speak with the installed fallback without opening Voice settings. Remove all Siri voices and retry; Berd should open Voice settings instead.

### DIFF
--- a/src-tauri/src/commands/siri_voice.rs
+++ b/src-tauri/src/commands/siri_voice.rs
@@ -716,6 +716,7 @@ fn resolve_voice_selection(
     })
 }
 
+#[cfg(any(target_os = "macos", test))]
 fn resolve_stream_voice(
     selection: &SiriVoiceSelection,
     load_all_voices: impl FnOnce() -> Result<Vec<SiriVoice>, String>,
--- a/src-tauri/src/commands/siri_voice.rs
+++ b/src-tauri/src/commands/siri_voice.rs
@@ -37,8 +37,15 @@ pub struct SiriVoiceState {
 struct SiriVoiceRuntime {
     active: Option<Arc<AtomicBool>>,
     owner_window: Option<String>,
+    resolved_voice: Option<ResolvedSiriVoice>,
     #[cfg(target_os = "macos")]
     stream: Option<ActiveSiriStream>,
+}
+
+#[derive(Clone, Debug)]
+struct ResolvedSiriVoice {
+    configured: Option<SiriVoiceSelection>,
+    effective: Option<SiriVoiceSelection>,
 }
 
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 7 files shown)


src-tauri/src/commands/siri_voice.rs
```diff
@@ -1033,12 +1057,14 @@ fn run_siri_stream(
...
 
 #[tauri::command]
+#[allow(clippy::too_many_arguments)] // Tauri injects four runtime dependencies beside the stream payload.
 pub fn start_siri_voice_stream(
     app: AppHandle,
     webview_window: tauri::WebviewWindow,
     state: tauri::State<'_, SiriVoiceState>,
     native_voice: tauri::State<'_, NativeVoiceState>,
     stream_id: String,
+    voice: SiriVoiceSelection,
```
morgmart, line 1082: 🤖 **P1 · Validate the stream voice at ingress** (blocking)

The changed command accepts `voice` from the renderer and passes it directly into the native stream. Unlike `select_siri_voice`, this receiver performs no catalog lookup to prove that the requested name/language is currently installed. The status used by the renderer can become stale when a voice is removed, and a handcrafted IPC request can bypass status entirely, so producer-side resolution is not a sufficient boundary check.

**User effect:** A voice conversation can be admitted as ready and then fail its first Siri response instead of directing the user to install or choose an available voice; malformed or stale renderer state can also request any native voice identifier.

**Recommended fix:** At the start-command boundary, resolve the supplied name/language against the current native voice catalog and reject it unless it is installed, before acquiring playback state or spawning the stream. Return a setup-actionable error so the existing recovery flow can send the user to Voice settings.

**Test:** Add a receiver-side regression test that supplies an unavailable SiriVoiceSelection directly to stream startup and proves the command rejects it before beginning playback; include a case where status previously resolved the voice but the current catalog no longer marks it installed.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The fallback selection is now supplied across the renderer-to-Tauri boundary, but the changed Tauri command trusts that value without confirming that it still identifies an installed Siri voice. Supplied GitHub evidence was inspected: all captured checks for the exact head SHA completed successfully. Static review only; no local code, tests, builds, package managers, hooks, or project scripts were run.

src-tauri/src/commands/siri_voice.rs
johnmatthewtennant: 🤖 Accepted the ingress-boundary concern. Commit 5c02fd10 now re-resolves the supplied voice against the current unfiltered native catalog before playback state is acquired. To preserve the feature requirement, a stale/unavailable supplied voice falls back to another installed Siri voice; startup is rejected with a Voice-settings-actionable error only when no Siri voice is installed. Added regressions for a voice removed after status resolution and for the no-installed-voice rejection, with 16 focused Siri tests plus Tauri check and Clippy passing.

morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the full exact three-dot comparison found no new publishable findings. The prior stream-ingress validation issue is now rechecked against the native catalog before playback state is acquired, with focused regression coverage. A remaining concern in the same underlying recovery flow is suppressed because the resolved authoritative thread contains a substantive human response accepting and explaining the chosen handling. Supplied GitHub evidence for the exact head SHA was inspected; all captured check runs completed successfully. No local code, tests, builds, package managers, hooks, or project scripts were run.

PR #206, opened 2026-08-25
fix(voice): improve Pocket playback speed quality

johnmatthewtennant:
## Summary

Pocket speech at accelerated and reduced speeds on macOS now uses AVAudioUnitTimePitch, preserving natural pitch while avoiding the metallic and choppy artifacts from the previous time-stretching path. Playback at 1x bypasses the effect, selected output-device routing remains supported, and interruption accounting reports only audio that the native player completed.

The lower-quality sstretch dependency and cross-platform fallback are removed. Pocket playback is macOS-only in this final state.

### Related issue

None found.

1. On macOS, select Pocket TTS in Voice settings and choose an accelerated playback speed such as 1.5x or 2x.
2. Start a voice conversation and request a multi-sentence response.
3. Confirm the speech keeps a natural pitch and avoids metallic or choppy artifacts.
4. Switch to 1x and confirm normal-speed playback remains natural. Select a non-default audio output and confirm Pocket playback uses it.

### Testing

John manually compared accelerated Pocket playback in a macOS development build and accepted the improved audio quality.


src-tauri/src/commands/pocket_voice.rs
```diff
@@ -2672,8 +2281,14 @@ fn run_pocket_voice_stream(
...
                         });
                     }
+                    player.ensure_healthy()?;
+                    if player.is_empty() {
+                        player.ensure_healthy()?;
+                        break;
+                    }
                     std::thread::sleep(Duration::from_millis(10));
                 }
+                assistant_speech.take();
```
morgmart: 🤖 **P1 · Preserve output-latency capture grace** (blocking)

The new completion path drops `assistant_speech` immediately after `pendingBuffers` reaches zero. That counter reaches zero from `AVAudioPlayerNodeCompletionDataPlayedBack`, but output routes can still buffer audio after the player node has rendered it. The replaced implementation deliberately retained this guard for `playback_latency_safety_duration(output_device)` (100 ms local, 500 ms Bluetooth, and 2 s AirPlay/unknown); that route-aware helper and its tests remain, but Pocket no longer uses it. The 100 ms subtraction in delivery accounting does not delay dropping the guard, so it cannot protect microphone capture.

**User effect:** During Pocket voice conversations on Bluetooth, AirPlay, virtual, or otherwise buffered outputs, microphone capture and interruption detection can resume while the assistant is still audibly speaking. Users can get assistant audio transcribed as their own speech, false interruptions, or feedback/echo at the end of each playback burst.

**Recommended fix:** Keep the assistant-speech guard alive until the native queue is drained plus the route-specific `playback_latency_safety_duration(output_device)`, while preserving immediate release on explicit stop, interruption, and failure. Use one lifecycle owner so a newly enqueued burst cannot race with an older grace timer and release the newer guard.

**Test:** Add a discriminating lifecycle test with an injectable completion signal/clock: after native queue drain, assert the guard remains active through the configured Bluetooth/AirPlay grace, then releases; enqueue a new burst during that grace and assert the stale completion cannot release the new guard. Also assert stop/failure releases promptly.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking lifecycle regression is present in the exact three-dot comparison. Pocket playback now drops its assistant-speech/capture-protection guard as soon as AVAudioPlayerNode reports its final source buffer completed, removing the existing route-aware output-latency grace; buffered Bluetooth, AirPlay, virtual, and unknown routes can therefore resume capture while speech is still physically audible. I statically reviewed all eight changed files and the complete Pocket playback flow (creation, routing, speed processing, enqueue, progress accounting, completion, interruption, failure, and cleanup), plus accessibility, localization, navigation/consent, async never-completes and races, tests, project laws/design-system applicability, and overlap/duplicate status. No additional supported findings survived the final self-check. Supplied GitHub evidence was inspected: the listed DCO, frontend, clean-room, and platform Rust checks succeeded, while combined status was pending; required checks still govern merge readiness. No local code, tests, builds, package managers, hooks, binaries, or project scripts were run.

src-tauri/src/commands/pocket_voice.rs
johnmatthewtennant: 🤖 Fixed in 17b726f4 and bdb8e354. Pocket now retains the assistant-speech guard through the existing route-aware output-latency grace. Every successful non-empty native enqueue resets the drain timestamp, so a short burst that fully drains before the owner loop polls still receives a fresh grace interval. Explicit stop, interruption, and failure continue to release promptly. Added a discriminating missed-poll regression test; Tauri checks, clippy, 30 Pocket tests, push hooks, and a targeted lifecycle rereview pass.

src-tauri/src/commands/pocket_voice.rs
```diff
@@ -2672,6 +2292,20 @@ fn run_pocket_voice_stream(
                             delivery: Some(delivery),
                         });
                     }
+                    player.ensure_healthy()?;
+                    if player.is_empty() {
```
morgmart: 🤖 **P1 · Bound missing completion callbacks** (blocking)

The `Finish` loop now exits only when `player.is_empty()` observes the native `pendingBuffers` counter reach zero. That counter is decremented solely by the AVAudioPlayerNode completion handler, and `ensure_healthy()` only detects a stopped engine or an explicitly recorded callback failure. If a scheduled-buffer completion callback is lost or never arrives while the engine remains running, this loop sleeps forever. The replaced path deliberately had `POCKET_SOURCE_COMPLETION_TIMEOUT` to release ownership when source-completion bookkeeping went missing, together with `playback_monitor_times_out_a_missing_source_completion_and_releases_ownership`; this PR removes both without an equivalent watchdog.

**User effect:** A voice response can remain permanently in playback after its audible audio ends, leaving the stream unfinished and assistant-speech/capture policy active. The conversation may stop advancing, and subsequent Pocket playback attempts can remain blocked as already active until the user tears down the session or window.

**Recommended fix:** Restore a bounded completion-bookkeeping fallback around the native drain wait. After the expected audio duration or a short post-drain watchdog expires without the counter advancing, stop the native player, release assistant-speech and playback ownership, and return a defined completion or failure outcome rather than waiting forever. Keep the new route-aware output-latency grace separate so a bookkeeping timeout cannot bypass capture protection for audio still plausibly audible.

**Test:** Add a discriminating lifecycle test using an injectable player/clock where pending buffers never decrement while health remains nominal. Assert the finish path terminates within the configured bound, releases the assistant-speech and playback guards, and permits a subsequent stream; also verify ordinary drain still waits through the route-specific output-latency grace.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: one blocking lifecycle regression is present in the exact three-dot comparison. The new finish path has no degraded path if the native pending-buffer completion bookkeeping never reaches zero, so Pocket playback can remain active indefinitely; the replaced implementation explicitly bounded this condition and its regression test was deleted. I statically reviewed all eight changed files and the complete Pocket playback flow (player creation, speed processing, output routing, enqueue, progress/delivery accounting, burst/grace handling, completion, interruption, failure, and cleanup), plus accessibility, localization, navigation/consent, async error/never-completes/race behavior, test honesty, project laws/design-system applicability, and finding overlap. The prior output-latency capture-grace issue is fixed in the current comparison and its resolved thread has a substantive human reply, so it is neither a new finding nor a duplicate suppression. No other supported findings survived the final self-check. Supplied GitHub evidence was inspected: DCO, frontend, clean-room, and platform Rust checks were successful, Transcript virtualization was in progress, and combined status was pending; required checks still govern merge readiness. No local code, tests, builds, package managers, hooks, binaries, or project scripts were run.

src-tauri/src/commands/pocket_voice.rs
johnmatthewtennant: 🤖 Fixed in 5067ac79. Finish now bounds native completion bookkeeping by the expected remaining source-audio duration at the selected playback rate plus a two-second allowance. On timeout it stops native output, then independently observes the route-specific output-latency grace before releasing capture protection. Interruption and failure still release promptly. Added discriminating timeout and timeout-to-route-grace tests; Tauri checks, clippy, 32 Pocket tests, push hooks, and targeted lifecycle rereview pass.

morgmart (dismissed):
🤖 Automated code review

APPROVE: no publishable findings remain in the exact three-dot comparison. I statically reviewed all eight changed files and traced every changed Pocket flow: native player creation and ownership, speed processing, output routing, streaming append/flush/finish, one-shot preview/speak, enqueue and delivery accounting, route-aware capture grace, timeout fallback, interruption, stop, failure, and cleanup. The stream finish path now bounds missing native completion bookkeeping and independently preserves route-aware capture grace. A similar never-completes risk remains in the changed one-shot preview/speak drain loop, but it is the same underlying missing-native-completion issue as the resolved protected thread whose substantive human reply states the fix scope; policy therefore suppresses it rather than publishing it again. Final checks covered accessibility and i18n (no UI/string-surface change requiring localization), navigation/consent (not affected), async errors, never-completes, lifecycle and races, test honesty, design-system/project rules, duplicate overlap, and concrete user effects; no other supported issue survived. Supplied GitHub evidence was structurally valid and inspected: DCO, frontend, transcript virtualization, clean-room, and macOS/Windows/Linux Rust checks all completed successfully, while combined status was pending; required checks still govern merge readiness. No local code, tests, builds, package managers, hooks, binaries, or project scripts were run.

morgmart (approved):
🤖 Automated code review

APPROVE: no publishable findings remain in the exact three-dot comparison. I statically reviewed all eight changed files and traced every changed Pocket flow: native player creation and ownership, AVAudioUnitTimePitch speed processing, output-device routing, streaming append/flush/finish, one-shot preview/speak, enqueue and delivery accounting, route-aware capture grace, bounded native-completion fallbacks, interruption, explicit stop, output failure, and cleanup. Both previously reported issues are fixed in the current comparison: streaming playback preserves route-aware capture protection after drain, and missing completion bookkeeping is bounded for both streaming and one-shot playback. Final self-checks covered every changed user flow; accessibility and i18n/localization (no changed visual, interactive, or translatable UI surface); navigation and consent guards (not affected); async errors, never-completes, lifecycle, cancellation, and races; test honesty and discriminating coverage; project laws and design-system rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. No supported concern survived. Supplied GitHub evidence was structurally valid and inspected: DCO, frontend build/tests, transcript virtualization, clean-room install, and macOS/Windows/Linux Rust checks all completed successfully, while combined status was pending; required checks still govern merge readiness. No local code, tests, builds, package managers, hooks, binaries, or project scripts were run.

PR #214, opened 2026-08-25
feat: add distribution-owned response feedback UI

comp615:
_This PR description was generated by AMP._

## Summary

Berd distributions need a generic way to collect turn-level response ratings without putting a company-specific telemetry endpoint or schema in the public app.

This adds opt-in thumbs-up/down controls tied to the canonical assistant response, session ID, and message ID. Stock Berd uses a typed no-op transport and ordering seam; distributions can supply their own sink and enable the UI through the transport-neutral `feedbackSurveys` build capability plus runtime config. Mounting, hovering, focusing, or revealing controls emits nothing; only explicit good, bad, and cleared transitions reach the sink.

Response feedback controls

### Related issue

N/A — no duplicate issue or PR found.

### Testing

```bash
bin/just check
bin/pnpm vitest run src/features/chat/response-feedback src/shared/runtime-config/schema.test.ts
bin/just _tauri-cargo-unix check --features block-feedback
```

<details>
<summary>AI tool validation prompt</summary>

Copy and paste into your AI tool:

```
Review block/berd branch ccroom/response-feedback-ui and verify:
1. Run the focused response-feedback and runtime-config tests.
2. Confirm stock Berd sends no feedback events, while the typed sink receives session ID, message ID, appearance ID, and response only when the user changes a rating; confirm the distribution sink owns delivery ordering.
3. Confirm only the canonical terminal assistant response receives controls in classic and virtual transcripts.
```

</details>


src/features/chat/ui/MessageBubble.tsx
```diff
@@ -728,6 +733,10 @@ export const MessageBubble = memo(function MessageBubble({
   const { isCopied: isCopyConfirmed, copyToClipboard } = useCopyToClipboard();
   const hasPersonaAvatar = Boolean(persona?.avatar);
   const catalogEntries = useProviderCatalogStore((state) => state.entries);
+  const feedbackEnabled = useProfileCapability("feedback");
```
kalvinnchau: 🤖 Could we separate human-feedback surveys from the existing `feedback` capability? That capability has `requiresKgoose: true`, while the typed survey sink itself has no KGoose dependency. A distro can enable `responseRatingEnabled` and provide its own transport yet still get no controls unless it also configures KGoose, which undercuts the generic seam. I suggest a dedicated transport-neutral survey capability that retains explicit build/runtime opt-in without requiring KGoose.

src/features/chat/response-feedback/feedbackSurveyEvents.ts
```diff
@@ -0,0 +1,38 @@
...
+>;
+
+const FEEDBACK_SEQUENCE_STORAGE_PREFIX = "berd:feedback-event-sequence:v1:";
+const volatileSequences = new Map<string, number>();
+
+function nextFeedbackEventSequence(sessionId: string): number {
+  const key = `${FEEDBACK_SEQUENCE_STORAGE_PREFIX}${sessionId}`;
+  let current = volatileSequences.get(key) ?? 0;
+  try {
+    const stored = Number(localStorage.getItem(key));
```
kalvinnchau: 🤖 This sequence allocator is not atomic across source/detached renderers: both can read `N`, persist `N+1`, and emit different responses with the same highest sequence. A distribution sink can renumber by arrival order, but then the public seam no longer provides the advertised final-state ordering. Can we choose one authoritative owner here—preferably a public native mutation that atomically updates selection and allocates a persistent revision—or remove authoritative sequencing from the public event and explicitly make it sink-owned?

src-tauri/src/commands/runtime_config.rs
```diff
@@ -162,6 +162,8 @@ pub struct RuntimeFeedbackConfig {
     pub enabled: Option<bool>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub project_key: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub response_rating_enabled: Option<bool>,
```
kalvinnchau, line 170: 🤖 Is the admin runtime-config endpoint guaranteed to version responses by client? `RuntimeFeedbackConfig` uses `deny_unknown_fields`, but this adds a field while retaining schema version 1. An older managed client receiving the new field rejects the entire endpoint response and falls back to cache/default policy; the stacked survey PR adds another field under the same v1 shape. If mixed client versions can reach one endpoint, this needs an extensible field or negotiated schema response rather than an undeclared additive v1 change.

kalvinnchau (commented):
🤖 Reviewed in the context of the stacked session-survey work and a distribution-owned transport. These comments are limited to unresolved contracts owned by this PR; later-stack concerns are left on the PR that owns them.

src/features/chat/ui/MessageBubble.tsx
comp615: Addressed in bfe7788. Human surveys now use a dedicated transport-neutral `feedbackSurveys` capability; the existing KGoose-dependent `feedback` capability remains unchanged. `VITE_FEEDBACK=1` remains a compatibility umbrella, while distributions can opt into surveys independently with `VITE_FEEDBACK_SURVEYS=1`. — Amp 🤖

src/features/chat/response-feedback/feedbackSurveyEvents.ts
comp615: Addressed in bfe7788 by removing renderer-owned sequencing from the public event contract. The public seam now explicitly makes ordering distribution-sink-owned; the Block sink allocates durable native per-session sequences. — Amp 🤖

src-tauri/src/commands/runtime_config.rs
comp615: Confirmed the compatibility risk is conditional on the service contract. The client sends no visible schema/client negotiation signal, so I avoided a speculative local version change that would not protect already-shipped strict v1 clients. Can mixed Berd versions receive the same `/desktop/v1/runtime-config` response, and if so what service-side signal shapes fields for each client? — Amp 🤖

comp615: Clarified after tracing the active release path: current public and internal Berd builds do not fetch this endpoint. `admin-runtime-config` is default-off, so they load a packaged `runtime-config.json`; Berd Internal injects its private config during assembly/release. `/desktop/v1/runtime-config` is a latent, feature-gated path for a future remote-config contract. If enabled later, the current request provides no explicit client/schema negotiation signal and strict older v1 clients could reject additive fields, so the endpoint must define response shaping or a new version before that feature is enabled. There is no mixed-client production compatibility issue from these fields today, and I am deferring speculative client versioning until the service contract exists. — Amp 🤖

kalvinnchau (approved):

PR #215, opened 2026-08-25
Add sampled session feedback survey

comp615:
_This PR description was generated by AMP._

## Summary

This builds the generic sampled session survey on top of the response-feedback seam. It is stacked on the response-feedback PR and should merge after it.

After five visible user turns and ten minutes, each eligible completed response gets a distribution-owned basis-point sampling opportunity, matching Claude and Amp. A selected prompt starts a persisted 27–29 hour cooldown, with arbitration serialized within the app process; this best-effort feedback path intentionally does not coordinate independent app processes. Dismiss is the default focus, Escape dismisses only while the visible survey owns focus, and focus returns to its prior target. Stock Berd still sends nothing; distributions own transport and rollout values.

Virtualized transcripts include the survey in offscreen height measurements without letting the hidden measurement copy focus or emit feedback.

Session feedback survey

### Related issue

N/A — no duplicate issue or PR found. Depends on [response-feedback PR #214](https://github.com/block/berd/pull/214).

### Testing

```bash
bin/just check
bin/pnpm vitest run src/features/chat/response-feedback src/shared/runtime-config/schema.test.ts
bin/just _tauri-cargo-unix test --lib feedback_survey
bin/just tauri-check
```

<details>
<summary>AI tool validation prompt</summary>

Copy and paste into your AI tool:

```
Review block/berd branch ccroom/session-feedback-survey and verify:
1. Run the focused survey, runtime-config, and native cooldown tests.
2. Confirm sessions shorter than five visible user turns or ten minutes never claim a survey.
3. Confirm a visible prompt defaults focus to Dismiss, Escape dismisses only while it owns focus, and a session is never prompted again after an appearance.
4. Confirm cooldown state persists across restart, claims are serialized within one app process, and the delay is randomized from 27 through 29 hours.
```

</details>

### DIFF
--- a/src-tauri/src/commands/feedback_survey.rs
+++ b/src-tauri/src/commands/feedback_survey.rs
@@ -6,6 +6,9 @@ const SESSION_SURVEY_COOLDOWN_FILE: &str = "session-feedback-survey-cooldown-v1"
 const SESSION_SURVEY_COOLDOWN_MINIMUM_MS: u64 = 27 * 60 * 60 * 1_000;
 const SESSION_SURVEY_COOLDOWN_JITTER_MS: u64 = 2 * 60 * 60 * 1_000;
 
+/// Coordinates survey cooldown claims within one app process and restores the
+/// persisted deadline after restart. Independent app processes may race; this
+/// feedback path is intentionally best-effort.
 pub struct SessionFeedbackSurveyCooldownState {
     path: PathBuf,
     next_eligible_at_ms: Mutex<u64>,
@@ -98,7 +101,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn cooldown_claim_is_atomic_and_persistent() {
+    fn cooldown_claim_is_process_atomic_and_persists_across_restart() {
         let dir = tempfile::tempdir().unwrap();
         let state = Arc::new(SessionFeedbackSurveyCooldownState::new(
             dir.path().to_path_buf(),
--- a/src-tauri/src/commands/feedback_survey.rs
+++ b/src-tauri/src/commands/feedback_survey.rs
@@ -1,0 +1,171 @@
+use std::{fs, io::Write, path::PathBuf, sync::Mutex, time::SystemTime};
+
+use serde::Deserialize;
+
+const SESSION_SURVEY_COOLDOWN_FILE: &str = "session-feedback-survey-cooldown-v1";
+const SESSION_SURVEY_COOLDOWN_MINIMUM_MS: u64 = 27 * 60 * 60 * 1_000;
+const SESSION_SURVEY_COOLDOWN_JITTER_MS: u64 = 2 * 60 * 60 * 1_000;
+
+pub struct SessionFeedbackSurveyCooldownState {
+    path: PathBuf,
+    next_eligible_at_ms: Mutex<u64>,
+}
+
+impl SessionFeedbackSurveyCooldownState {
+    pub fn new(app_data_dir: PathBuf) -> Self {
+        let path = app_data_dir.join(SESSION_SURVEY_COOLDOWN_FILE);
+        let next_eligible_at_ms = fs::read_to_string(&path)
+            .ok()
+            .and_then(|value| value.trim().parse().ok())
+            .unwrap_or(0);
+        Self {
+            path,
+            next_eligible_at_ms: Mutex::new(next_eligible_at_ms),
+        }
+    }
+
+    fn claim(
+        &self,
+        input: SessionFeedbackSurveyCooldownInput,
+        now_ms: u64,
+    ) -> Result<bool, String> {
+        if input.sampling_rate_basis_points > 10_000
+            || !input.random.is_finite()
+            || !(0.0..=1.0).contains(&input.random)
+            || !input.cooldown_random.is_finite()
+            || !(0.0..=1.0).contains(&input.cooldown_random)
+        {
+            return Err("invalid session feedback survey cooldown input".to_string());
+        }
+        let mut next_eligible_at_ms = self
+            .next_eligible_at_ms
+            .lock()
+            .map_err(|_| "session feedback survey cooldown lock poisoned".to_string())?;
+        if input.sampling_rate_basis_points == 0
+            || now_ms < *next_eligible_at_ms
+            || input.random * 10_000.0 >= f64::from(input.sampling_rate_basis_points)
+        {
+            return Ok(false);
+        }
+
+        let jitter_ms = (input.cooldown_random * SESSION_SURVEY_COOLDOWN_JITTER_MS as f64) as u64;
+        let claimed_until = now_ms
+            .saturating_add(SESSION_SURVEY_COOLDOWN_MINIMUM_MS)
+            .saturating_add(jitter_ms);
+        let parent = self
+            .path
+            .parent()
+            .ok_or_else(|| "survey cooldown path has no parent".to_string())?;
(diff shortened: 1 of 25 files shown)


src/features/chat/response-feedback/sessionFeedbackSurveyState.ts
```diff
@@ -0,0 +1,223 @@
...
+  sessionCreatedAt,
+  userTurnCount,
+  samplingRateBasisPoints,
+  now = Date.now(),
+  random = Math.random(),
+  cooldownRandom = Math.random(),
+}: SessionFeedbackSurveyClaimInput): Promise<ActiveSessionFeedbackSurvey | null> {
+  let existing = readSessionRecord(sessionId);
+  const createdAt = Date.parse(sessionCreatedAt);
+  if (existing?.active && currentMessageIds.has(existing.active.messageId)) {
```
kalvinnchau, line 125: 🤖 This returns an active survey before considering `existing.appeared`. Since marking appearance leaves `active` populated, reload, navigation away/back, or a session-window handoff presents the same prompt again, despite the stated once-after-appearance contract. Could presentation eligibility reject an appeared record on a new mount while retaining enough state for the original mounted instance to submit? A remount-after-appearance test would pin this boundary.

src/features/chat/response-feedback/sessionFeedbackSurveyState.ts
```diff
@@ -0,0 +1,223 @@
...
+  if (rate === 0) return null;
+  const selected = await claimSessionFeedbackSurveyCooldown({
+    samplingRateBasisPoints: rate,
+    random,
+    cooldownRandom,
+  }).catch(() => false);
+  const active = selected
+    ? { appearanceId: crypto.randomUUID(), messageId }
+    : null;
+  writeSessionRecord(sessionId, {
```
kalvinnchau, line 164: 🤖 This writes a complete record after awaiting native arbitration, based on state read before the await. Two renderer realms can both read no record; one receives `true`, the other `false`, and the losing renderer can overwrite the winner with `active: null`, consuming the cooldown without a usable prompt. `sessionClaimQueues` only serializes one JS realm. The native claim should own the per-session active record in the same transaction and notify renderers of changes.

src-tauri/src/commands/feedback_survey.rs
```diff
@@ -0,0 +1,171 @@
...
+
+pub struct SessionFeedbackSurveyCooldownState {
+    path: PathBuf,
+    next_eligible_at_ms: Mutex<u64>,
+}
+
+impl SessionFeedbackSurveyCooldownState {
+    pub fn new(app_data_dir: PathBuf) -> Self {
+        let path = app_data_dir.join(SESSION_SURVEY_COOLDOWN_FILE);
+        let next_eligible_at_ms = fs::read_to_string(&path)
```
kalvinnchau, line 20: 🤖 The cooldown file is read once into process-local state and subsequent claims only lock that process's `Mutex`. Berd supports concurrent app instances, so two instances started before either claim retain the same stale eligibility and can both select inside the promised device-wide cooldown. Atomic replacement prevents torn writes but not an atomic cross-process read/check/write; this needs a file lock plus reread under lock, or a transactional persistent store.

src/features/chat/ui/VirtualMessageTimeline.tsx
```diff
@@ -3563,6 +3574,10 @@ function VirtualMessageTimelineSession({
       })}
       message={row.messageId ? stableMessageByRowId.get(row.rowId) : undefined}
       isStreaming={false}
+      feedbackSessionId={
+        responseFeedbackRowIds.has(row.rowId) ? sessionId : undefined
+      }
+      sessionFeedbackSurvey={sessionFeedbackSurveyForRow(row)}
```
kalvinnchau, line 3616: 🤖 This includes the survey in offscreen real-row measurement, but ordinary completed text responses use `measure-shell` and never take this path. Their virtual height excludes the survey until the live row mounts, causing a correction when the user scrolls or jumps to it. Could survey presence update the measurement revision/policy and either force safe real measurement or account for the localized, wrapping survey in shell measurement?

src/features/chat/response-feedback/SessionFeedbackSurvey.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }, [measurementOnly, sessionId, survey.appearanceId]);
+
+  const respond = useCallback(
+    (response: SessionFeedbackSurveyResponse) => {
+      recordSessionFeedbackSurveyResponse(
+        sessionId,
+        survey.appearanceId,
+        response,
+      );
+      setVisible(false);
```
kalvinnchau, line 64: 🤖 Because the survey programmatically focuses Dismiss, removing the focused fieldset here leaves focus on the document body. The next Tab can restart near the beginning of the app. Please restore a stable prior target, with the transcript row or composer as a fallback, and cover both Escape and button activation.

kalvinnchau (commented):
🤖 Product-semantics question for the stack: the basis-point rate is currently applied to every eligible completion until one is selected, not once per session. At 2.5%, 20 eligible opportunities produce about a 40% chance of a prompt and 100 produce about 92%, weighting the sample toward long, active sessions. If this is intentionally an opportunity hazard rate, the PR should name it that way; if it is meant to be a 2.5% session rollout, sample once and persist the sampled-out result.

src/features/chat/response-feedback/sessionFeedbackSurveyState.ts
comp615: Addressed in f202dab. A claim with `appeared` is no longer presentable on a new mount, while the mounted owner retains its active survey long enough to submit. Added remount coverage. — Amp 🤖

src/features/chat/response-feedback/sessionFeedbackSurveyState.ts
comp615: Addressed in f202dab without moving the full state machine native-side. After native arbitration, a renderer rereads and merges persisted state, preserving another renderer’s winner instead of overwriting it. A controlled interleaving test covers the race. — Amp 🤖

src-tauri/src/commands/feedback_survey.rs
comp615: We narrowed the contract rather than adding OS locking for noncritical best-effort feedback. Claims are serialized within one app process and the deadline persists across restart; independent simultaneous app processes may race. The PR and tests now state that boundary explicitly. — Amp 🤖

src/features/chat/ui/VirtualMessageTimeline.tsx
comp615: Addressed in f202dab. A virtual row that owns the active survey now forces real offscreen measurement and receives an appearance/locale-specific height revision. Regression coverage verifies the measurement policy. — Amp 🤖

src/features/chat/response-feedback/SessionFeedbackSurvey.tsx
comp615: Addressed in f202dab. The survey captures the stable element focused before Dismiss, restores it after button or Escape dismissal when connected, and falls back to the composer. Both paths are covered. — Amp 🤖

comp615:
Charlie confirmed the 250-bp sampling behavior is intentionally a per-eligible-completion opportunity hazard because it matches Claude and Amp. PR text, code comments, and tests now name that behavior rather than implying a 2.5% allocation of sessions. — Amp 🤖

kalvinnchau (approved):

PR #230, opened 2026-08-26
fix(chat): preserve assistant order after batched steers

johnmatthewtennant:
## Summary

Fixes a chat transcript ordering bug for live steering. When a prior assistant response is completing and multiple user steers are delivered before the next assistant response starts, Berd now anchors the continuation after the contiguous delivered steer batch instead of after only the first delivered steer. Undelivered queued steers still stay ahead of their own later continuation.

### Related issue

none found

### Testing

- `pnpm vitest run src/features/chat/stores/__tests__/chatStore.test.ts src/features/chat/acp/__tests__/acpNotificationHandler.test.ts`
- `just check`

Red/green regression proof:

Apply only the regression test to `origin/main`, then run the focused case:

```bash
pnpm vitest run src/features/chat/acp/__tests__/acpNotificationHandler.test.ts \
  -t 'keeps assistant responses in canonical order when multiple steers arrive before the next response'
```

Before this fix, the focused case fails because the continuation assistant is inserted between the two delivered steers:

```diff
 assistant-before-steers
 steer-1
-steer-2
-assistant continuation
+assistant continuation
+steer-2
```

On this PR, the same focused case passes. The seeded live transcript shape matches session `20260826_2`:

1. assistant response with `completionStatus: "inProgress"`
2. two delivered steer user messages with the same timestamp
3. next assistant chunk begins after the first steer boundary is processed

The resulting transcript order remains:

```text
assistant-before-steers
steer-1
steer-2
assistant continuation
```

The store test `does not cross a steering message that has not been delivered` covers the adjacent preserved behavior: if a later local steer is still `delivery: "steering"`, the continuation for the delivered steer is inserted before that undelivered steer.

### DIFF
--- a/src/features/chat/acp/__tests__/acpNotificationHandler.test.ts
+++ b/src/features/chat/acp/__tests__/acpNotificationHandler.test.ts
@@ -787,8 +787,8 @@ describe("acpNotificationHandler", () => {
       useChatStore.getState().messagesBySession["acp-session"],
     ).toMatchObject([
       { id: "backend-steer-1", metadata: { delivery: "steer" } },
-      { id: "local-steer-2", metadata: { delivery: "steering" } },
       { role: "assistant" },
+      { id: "local-steer-2", metadata: { delivery: "steering" } },
     ]);
     expect(
       useChatStore.getState().getSessionRuntime("acp-session")
--- a/src/features/chat/acp/__tests__/acpNotificationHandler.test.ts
+++ b/src/features/chat/acp/__tests__/acpNotificationHandler.test.ts
@@ -575,6 +575,95 @@ describe("acpNotificationHandler", () => {
     ).toBeNull();
   });
 
+  it("keeps assistant responses in canonical order when multiple steers arrive before the next response", async () => {
+    useChatStore.getState().setMessages("acp-session", [
+      {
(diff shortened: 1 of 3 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison 0821223fe847aaeb4742256981d82368ff9c9065...5be85725da20a043151e68788866318e3414ea3e found no publishable findings. The continuation now derives its insertion point from the contiguous batch of delivered steer messages while stopping before an undelivered steer, preserving canonical transcript order and the queue/steering laws. The added store and ACP regression tests are discriminating and cover both the repaired multiple-delivered-steer case and the adjacent undelivered-steer boundary. Final self-check covered every changed user flow, accessibility, i18n/localization, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, project laws/design-system rules, duplicate overlap, and blocking-evidence/user-effect requirements; this state-only change introduces no new UI, copy, navigation, consent, or accessibility surface. Supplied GitHub evidence was inspected: all listed check runs for the exact head SHA completed successfully, while the legacy combined commit status is pending with no statuses; required checks still govern merge readiness. No local code, tests, builds, package managers, binaries, hooks, or project scripts were run.

PR #232, opened 2026-08-26
feat(voice): add OpenAI speech-to-text

johnmatthewtennant:
## Summary

Adds OpenAI realtime speech-to-text as a Voice Conversation input backend.

- Adds an OpenAI API-key field in Voice settings.
- Uses the same Berd-owned OpenAI voice credential as text-to-speech. The speech input and speech output fields are two entry points for the same key.
- Never reads Goose provider credentials or the generic `OPENAI_API_KEY` environment variable.
- Waits for the realtime transcription session to be ready before accepting microphone audio.
- Preserves committed transcript delivery while discarding stale turns across mute boundaries.
- Stops active voice and playback before changing or clearing the shared credential.

## Configuration

Enter an OpenAI key in either **Settings → Voice → Speech input** or **Settings → Voice → Speech output**. Saving or clearing it from either location updates the same system credential-manager entry for both voice backends. Berd-scoped environment variables may override voice models and the OpenAI voice base URL, but not the credential.


src-tauri/src/commands/native_voice.rs
```diff
@@ -2473,6 +2580,613 @@ fn macos_stt_worker(
...
+    let mut resampler = match Fft::<f32>::new(48_000, 24_000, 960, 2, 1, FixedSync::Input) {
+        Ok(resampler) => resampler,
+        Err(error) => {
+            let _ = event_tx.blocking_send(SttMessage::Failed(format!(
+                "Could not initialize OpenAI audio resampling: {error}"
+            )));
+            return;
+        }
+    };
+    if startup_tx.send(()).is_err() {
```
morgmart: 🤖 **P1 · Wait for server readiness** (blocking)

The worker sends session.update and immediately signals startup_tx after constructing the local resampler. A successful WebSocket send only confirms that the client handed off the message; it does not confirm that OpenAI accepted and applied the transcription configuration. start_native_voice_conversation therefore returns and microphone audio can be appended before a session.updated acknowledgement (or an error response) is received, despite this flow's stated readiness contract.

**User effect:** On a slow connection or when OpenAI rejects the requested model/configuration, Voice Conversation can appear to start and accept speech before transcription is actually ready. The first utterance may be lost or the call may fail only after the user has already begun speaking.

**Recommended fix:** Keep startup pending while reading WebSocket events until the matching successful session.updated acknowledgement arrives; surface an OpenAI error, disconnect, or timeout as startup failure. Only then signal startup_tx and begin consuming microphone batches.

**Test:** Add a discriminating worker/protocol test where session.update is written but session.updated is delayed: startup must remain pending and no audio append may be sent. Also cover an error received before acknowledgement and verify startup fails rather than reporting ready.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the OpenAI STT startup path reports readiness after only writing session.update, before the server confirms that the transcription session is configured. Static review covered all 18 changed files and the complete exact three-dot comparison. Supplied GitHub evidence was inspected; all captured checks for the exact head SHA are completed successfully. No local code, tests, builds, package managers, hooks, binaries, or project scripts were run.

src-tauri/src/commands/native_voice.rs
johnmatthewtennant: 🤖 Accepted and fixed in b10e1159, with follow-up lifecycle hardening through 9a1bcd56. OpenAI STT startup now remains pending until the worker receives `session.updated`; a pre-ack OpenAI error, disconnect, or timeout is returned as startup failure, and microphone batches are not consumed before acknowledgement. Added local WebSocket regressions for delayed acknowledgement and pre-ack rejection. Validation: focused Rust tests (2 passed), `just check`, `just tauri-check`, `just clippy`, pre-push gates, and exact-change local review all pass.

morgmart (dismissed):
🤖 Automated code review

APPROVE: fresh static inspection of all 19 changed files in the exact three-dot comparison found no publishable findings. The prior server-readiness issue is fixed: startup now waits for session.updated, propagates pre-acknowledgement errors and timeouts, and has discriminating delayed-acknowledgement and rejection coverage. Final self-check covered every changed voice setup and conversation flow; accessibility; English and Spanish localization; navigation, foreground ownership, and consent guards; async failure, never-completes, shutdown, mute-boundary, transcript-ordering, and race behavior; test honesty; design-system and project rules; duplicates and overlap; and the evidence required for blocking severity. Supplied GitHub evidence is structurally valid and was inspected: DCO, frozen install, Semgrep, and zizmor succeeded, while frontend and Tauri Rust checks were still in progress, so required checks still govern merge readiness. No local code, tests, builds, package managers, generated binaries, hooks, or project scripts were run.

**Pending checks:** 5 check(s) are not complete.

src-tauri/src/commands/openai_audio.rs
```diff
@@ -319,32 +375,89 @@ pub async fn get_openai_voice_status(
...
+    api_key: String,
+) -> Result<(), String> {
+    let api_key = api_key.trim();
+    if api_key.is_empty() {
+        return Err("OpenAI speech-to-text API key cannot be empty".to_string());
+    }
+    native_voice
+        .stop_active_then(&app, &capture, || {
+            stop_openai_voice_inner(&state)?;
+            openai_voice_credentials::store(OpenAiVoiceCredential::SpeechToText, api_key)?;
```
morgmart, line 392: 🤖 **P2 · Commit key changes atomically** (blocking)

The shared credential store writes the new `api-key` first and then returns the result of deleting the legacy `tts-api-key`. If that cleanup fails, this command returns an error before incrementing `credential_revision`, updating `configured`, or emitting the settings-changed event even though the requested new key is already persisted. The clear path has the same partial-success shape in reverse: it deletes the shared key before a legacy cleanup failure can make the command report failure. This leaves the command result and in-memory/UI state disagreeing with the actual keychain state.

**User effect:** A user can be told that saving or removing the OpenAI key failed when the change actually happened. The current screen may continue to show the old readiness state, and retrying can repeat a destructive or confusing operation until the next full status refresh or restart reveals the real state.

**Recommended fix:** Treat migration cleanup as best-effort after the canonical shared-key mutation, or otherwise make the operation transactional. Once the canonical write/delete succeeds, update the revision and configured state and emit the settings event; surface legacy cleanup separately without reporting that the requested key change failed.

**Test:** Add discriminating credential-command tests where the canonical shared-key write/delete succeeds but deletion of the legacy account fails. Verify the command still reports the requested mutation consistently, updates readiness/revision, emits the refresh event, and leaves subsequent reads matching the visible state.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: fresh static inspection of all 19 changed files in the exact three-dot comparison found one blocking credential-state consistency issue. Final self-check covered every changed voice settings and conversation flow; accessibility; English and Spanish localization; navigation, foreground ownership, and consent guards; async failure, never-completes, shutdown, mute-boundary, transcript-ordering, startup, and race behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. The prior server-readiness issue remains fixed and is protected by a resolved thread with a substantive human reply, so it is neither a finding nor a suppression. Supplied GitHub evidence is structurally valid and was inspected; all captured checks for the exact head SHA completed successfully. No local code, tests, builds, package managers, generated binaries, hooks, or project scripts were run.

src-tauri/src/commands/openai_audio.rs
johnmatthewtennant: 🤖 Accepted and fixed in the updated base at 72ff2e04, carried by current #232 head 1279d4c0. Canonical credential save/clear now defines the mutation result; legacy TTS-account cleanup is best-effort and logs a warning. Therefore `set_openai_voice_api_key` continues through credential revision, configured-state, and settings-event updates whenever the requested canonical mutation succeeds. The shared fake-state regression covers canonical save and clear with failing legacy cleanup. Validation: focused credential tests (2 passed) plus formatting, frontend checks, Tauri checks, clippy, and pre-push gates on both stack layers.

morgmart (approved):
🤖 Automated code review

APPROVE: fresh static inspection of all 19 changed files in the exact three-dot comparison found no publishable findings. The prior readiness issue remains fixed by waiting for session.updated and covering delayed acknowledgement and pre-acknowledgement rejection. The prior credential consistency issue remains fixed: canonical save/delete determines success while legacy cleanup is best-effort, with discriminating regression coverage. Final self-check covered every changed voice settings and conversation flow; accessibility; English and Spanish localization; navigation, foreground ownership, and consent guards; async failure, never-completes, startup, shutdown, mute-boundary, transcript-ordering, credential migration, lifecycle, and race behavior; test honesty; design-system and project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. Supplied GitHub evidence is structurally valid and was inspected; all captured checks for the exact head SHA completed successfully. No local code, tests, builds, package managers, generated binaries, hooks, or project scripts were run.

PR #234, opened 2026-08-26
feat(agents): bundle native monitoring tools

johnmatthewtennant:
## Summary

Bundle long-lived monitoring and cross-session orchestration directly with Berd:

- Add a native `berd-monitor` sidecar that detaches producers, durably buffers line-oriented output, wakes the owning session through `berdctl`, and stops complete process trees on macOS, Linux, and Windows.
- Bundle `berd-monitor` and `berd-orchestrator` skills so agents can discover these workflows without private configuration.
- Add optional `--from` attribution to `berdctl session create` and `session send`, while retaining trusted Berd provenance. Monitor deliveries use `--from berd-monitor`.
- Add optional `--delivery-id` to `session send`. Repeated IDs for one session are accepted without creating duplicate user turns, so a monitor can safely retry after a crash between delivery and local acknowledgement.

Monitor state is private, session-scoped, and exclusively owned. Failed or timed-out launches terminate and reap their detached child. Producer trees are cleaned up on every exit path. Delivery subprocesses have a monitor-owned deadline. Unterminated and oversized output is bounded and safely chunked, and final delivery retries until success or explicit stop.

### Related issue

none found

### Testing

- `cargo test -p berdctl`: 52 passed, 1 ignored
- `cargo test -p berd-monitor`: 22 passed
- `cargo check -p berd-monitor --target x86_64-pc-windows-msvc`
- Live smoke tests verified custom create/send attribution, detached monitor wake-up, duplicate-owner rejection, synchronous launch failure, and process-tree termination.

### DIFF
--- a/src-tauri/crates/berd-monitor/src/main.rs
+++ b/src-tauri/crates/berd-monitor/src/main.rs
@@ -21,6 +21,7 @@ const MAX_LABEL_CODE_UNITS: usize = 120;
 const MAX_LOCK_CANDIDATES: usize = 8;
 const DELIVERY_POLL_INTERVAL: Duration = Duration::from_millis(50);
 const LAUNCH_TIMEOUT: Duration = Duration::from_secs(10);
+const LAUNCH_TERMINATION_GRACE: Duration = Duration::from_secs(1);
 
 #[derive(Clone, Copy, Debug, ValueEnum)]
 enum RunningMode {
@@ -259,10 +260,19 @@ fn spawn_detached(state_key: &str, session_id: &str) -> Result<(), String> {
     let mut process = child
         .spawn()
         .map_err(|error| format!("start detached monitor: {error}"))?;
-    let deadline = Instant::now() + LAUNCH_TIMEOUT;
+    wait_for_launch(&mut process, &paths, &status_path, LAUNCH_TIMEOUT)
+}
+
+fn wait_for_launch(
+    process: &mut std::process::Child,
+    paths: &StatePaths,
+    status_path: &Path,
+    timeout: Duration,
+) -> Result<(), String> {
+    let deadline = Instant::now() + timeout;
     loop {
-        if let Ok(status) = fs::read_to_string(&status_path) {
-            let _ = fs::remove_file(&status_path);
+        if let Ok(status) = fs::read_to_string(status_path) {
+            let _ = fs::remove_file(status_path);
             if status.trim() == "ready" {
                 println!("{} {}", process.id(), paths.root.display());
                 return Ok(());
@@ -273,17 +283,41 @@ fn spawn_detached(state_key: &str, session_id: &str) -> Result<(), String> {
                 .to_owned());
         }
         if Instant::now() >= deadline {
+            terminate_timed_out_launch(process, paths, status_path);
             return Err(format!(
                 "monitor did not become ready within {} seconds; inspect {}",
-                LAUNCH_TIMEOUT.as_secs(),
+                timeout.as_secs_f64(),
                 paths.log.display()
             ));
         }
         if let Ok(Some(status)) = process.try_wait() {
+            let _ = fs::remove_file(status_path);
             return Err(format!("monitor exited before becoming ready ({status})"));
         }
         thread::sleep(Duration::from_millis(50));
     }
+}
+
+fn terminate_timed_out_launch(
+    process: &mut std::process::Child,
+    paths: &StatePaths,
+    status_path: &Path,
+) {
+    let _ = fs::write(&paths.stop, b"stop\n");
+    let deadline = Instant::now() + LAUNCH_TERMINATION_GRACE;
+    while Instant::now() < deadline {
+        match process.try_wait() {
+            Ok(Some(_)) => break,
+            Ok(None) => thread::sleep(Duration::from_millis(50)),
+            Err(_) => break,
+        }
+    }
+    if process.try_wait().ok().flatten().is_none() {
+        let _ = process.kill();
+        let _ = process.wait();
+    }
+    let _ = fs::remove_file(status_path);
+    let _ = fs::remove_file(&paths.stop);
 }
 
 #[cfg(unix)]
@@ -1060,6 +1094,41 @@ fn render_command(command: &[OsString]) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[cfg(unix)]
+    #[test]
+    fn launch_timeout_terminates_and_reaps_the_child() {
+        let key = format!(
+            "launch-timeout-test-{}-{}",
+            std::process::id(),
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .unwrap()
+                .as_nanos()
+        );
+        let paths = StatePaths::for_key(&key, "test-session");
+        ensure_private_directory(&paths.root).unwrap();
+        let status_path = paths.launch_status("never-ready");
+        let mut child = Command::new("sh")
+            .args(["-c", "exec sleep 30"])
+            .spawn()
+            .unwrap();
+        let pid = child.id();
+
+        let error = wait_for_launch(
+            &mut child,
+            &paths,
+            &status_path,
+            Duration::from_millis(100),
+        )
+        .unwrap_err();
+
+        assert!(error.contains("did not become ready"));
+        assert!(!test_process_exists(pid));
+        assert!(!status_path.exists());
+        assert!(!paths.stop.exists());
+        fs::remove_dir_all(paths.root).unwrap();
+    }
 
     #[test]
     fn failed_atomic_replace_preserves_pending_output() {
--- a/src-tauri/crates/berd-monitor/src/main.rs
+++ b/src-tauri/crates/berd-monitor/src/main.rs
@@ -1,0 +1,1102 @@
+use clap::{Parser, Subcommand, ValueEnum};
+use fs2::FileExt;
+use std::env;
+use std::ffi::OsString;
+use std::fs::{self, File, OpenOptions};
+use std::io::{self, BufRead, BufReader, Read, Write};
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitCode, Stdio};
+use std::sync::mpsc::{self, RecvTimeoutError};
+use std::thread;
+use std::time::{Duration, Instant, SystemTime};
+
+const FOREGROUND_ENV: &str = "BERD_MONITOR_FOREGROUND";
+const LAUNCH_TOKEN_ENV: &str = "BERD_MONITOR_LAUNCH_TOKEN";
+const RETRY_INTERVAL: Duration = Duration::from_secs(15);
+const BATCH_WINDOW: Duration = Duration::from_millis(250);
+const MAX_DELIVERY_BYTES: usize = 40_000;
+const MAX_PROMPT_CODE_UNITS: usize = 49_000;
+const MAX_INSTRUCTIONS_CODE_UNITS: usize = 4_000;
+const MAX_LABEL_CODE_UNITS: usize = 120;
+const LAUNCH_TIMEOUT: Duration = Duration::from_secs(10);
+
+#[derive(Clone, Copy, Debug, ValueEnum)]
+enum RunningMode {
+    Steer,
+    Queue,
+}
+
+impl RunningMode {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Steer => "steer",
+            Self::Queue => "queue",
+        }
+    }
+}
+
+#[derive(Debug, Parser)]
+#[command(
+    name = "berd-monitor",
+    about = "Run a long-lived command and wake its owning Berd session"
+)]
+struct Cli {
+    #[command(subcommand)]
+    command: MonitorCommand,
+}
+
+#[derive(Debug, Subcommand)]
+enum MonitorCommand {
+    /// Start a detached line-oriented command monitor.
+    Run {
+        /// Stable key used to identify and stop this monitor.
+        #[arg(long)]
+        state_key: String,
+        /// Concise source name included with delivered events.
+        #[arg(long)]
+        label: String,
+        /// Guidance appended to every delivered event.
+        #[arg(long, default_value = "")]
+        instructions: String,
+        /// Berd session that owns the monitor.
+        #[arg(long, env = "AGENT_SESSION_ID")]
+        session_id: String,
+        /// Whether events steer a running turn or queue behind it.
+        #[arg(long, value_enum, default_value_t = RunningMode::Steer)]
+        if_running: RunningMode,
+        /// Producer command and arguments, following `--`.
+        #[arg(last = true, required = true, allow_hyphen_values = true)]
+        command: Vec<OsString>,
+    },
+    /// Request that a running monitor stop its producer.
+    Stop {
+        /// Stable key passed when the monitor was started.
+        #[arg(long)]
+        state_key: String,
+        /// Berd session that owns the monitor.
+        #[arg(long, env = "AGENT_SESSION_ID")]
+        session_id: String,
+    },
+}
+
+struct StatePaths {
+    root: PathBuf,
+    log: PathBuf,
+    pending: PathBuf,
+    owner: PathBuf,
+    stop: PathBuf,
+}
+
+impl StatePaths {
+    fn for_key(key: &str, session_id: &str) -> Self {
+        let identity = format!("{session_id}\0{key}");
+        let suffix = stable_hash(identity.as_bytes());
+        let safe = key
+            .chars()
+            .take(80)
+            .map(|character| {
+                if character.is_ascii_alphanumeric() || matches!(character, '.' | '_' | '-') {
+                    character
+                } else {
+                    '-'
+                }
+            })
+            .collect::<String>();
+        let root = state_base_dir().join(format!("{safe}-{suffix:016x}"));
+        Self {
+            log: root.join("watcher.log"),
+            pending: root.join("pending.txt"),
+            owner: root.join("owner.pid"),
+            stop: root.join("stop-requested"),
+            root,
+        }
+    }
+
+    fn launch_status(&self, token: &str) -> PathBuf {
+        self.root.join(format!("launch-{token}.status"))
+    }
+}
+
+fn state_base_dir() -> PathBuf {
+    if let Some(explicit) = env::var_os("BERD_MONITOR_STATE_DIR") {
+        return PathBuf::from(explicit);
+    }
+    #[cfg(target_os = "windows")]
+    if let Some(local) = env::var_os("LOCALAPPDATA") {
+        return PathBuf::from(local).join("Berd").join("monitor");
+    }
+    #[cfg(target_os = "macos")]
+    if let Some(home) = env::var_os("HOME") {
+        return PathBuf::from(home)
+            .join("Library")
+            .join("Caches")
+            .join("Berd")
+            .join("monitor");
+    }
+    #[cfg(all(unix, not(target_os = "macos")))]
+    {
+        if let Some(runtime) = env::var_os("XDG_RUNTIME_DIR") {
+            return PathBuf::from(runtime).join("berd-monitor");
+        }
+        if let Some(state) = env::var_os("XDG_STATE_HOME") {
+            return PathBuf::from(state).join("berd").join("monitor");
+        }
+        if let Some(home) = env::var_os("HOME") {
+            return PathBuf::from(home)
+                .join(".local")
+                .join("state")
+                .join("berd")
+                .join("monitor");
+        }
+    }
+    env::temp_dir().join(format!("berd-monitor-{}", current_user_suffix()))
+}
+
+#[cfg(unix)]
+fn current_user_suffix() -> u32 {
+    unsafe { libc::geteuid() }
+}
+
+#[cfg(windows)]
+fn current_user_suffix() -> u32 {
+    std::process::id()
+}
+
+fn stable_hash(bytes: &[u8]) -> u64 {
+    bytes.iter().fold(0xcbf29ce484222325, |hash, byte| {
+        (hash ^ u64::from(*byte)).wrapping_mul(0x100000001b3)
+    })
+}
+
+fn main() -> ExitCode {
+    match run_cli() {
+        Ok(()) => ExitCode::SUCCESS,
+        Err(error) => {
+            eprintln!("berd-monitor: {error}");
+            ExitCode::FAILURE
+        }
+    }
+}
+
+fn run_cli() -> Result<(), String> {
+    let cli = Cli::parse();
+    match cli.command {
+        MonitorCommand::Run {
+            state_key,
+            label,
+            instructions,
+            session_id,
+            if_running,
+            command,
+        } => {
+            validate_single_line("--label", &label)?;
+            if label.encode_utf16().count() > MAX_LABEL_CODE_UNITS {
+                return Err(format!(
+                    "--label must not exceed {MAX_LABEL_CODE_UNITS} characters"
+                ));
+            }
+            if instructions.encode_utf16().count() > MAX_INSTRUCTIONS_CODE_UNITS {
+                return Err(format!(
+                    "--instructions must not exceed {MAX_INSTRUCTIONS_CODE_UNITS} characters"
+                ));
+            }
+            if env::var_os(FOREGROUND_ENV).is_some() {
+                run_foreground(
+                    &state_key,
+                    &label,
+                    &instructions,
+                    &session_id,
+                    if_running,
+                    &command,
+                    env::var(LAUNCH_TOKEN_ENV).ok().as_deref(),
+                )
+                .map_err(|error| error.to_string())
+            } else {
+                spawn_detached(&state_key, &session_id)
+            }
+        }
+        MonitorCommand::Stop {
+            state_key,
+            session_id,
+        } => request_stop(&state_key, &session_id).map_err(|error| error.to_string()),
+    }
+}
+
+fn validate_single_line(flag: &str, value: &str) -> Result<(), String> {
+    if value.trim().is_empty() {
+        return Err(format!("{flag} must not be empty"));
+    }
+    if value.contains('\n') || value.contains('\r') {
+        return Err(format!("{flag} must be a single line"));
+    }
+    Ok(())
+}
+
+fn spawn_detached(state_key: &str, session_id: &str) -> Result<(), String> {
+    let executable =
+        env::current_exe().map_err(|error| format!("resolve current executable: {error}"))?;
+    let token = format!(
+        "{}-{}",
+        std::process::id(),
+        SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos()
+    );
+    let paths = StatePaths::for_key(state_key, session_id);
+    let status_path = paths.launch_status(&token);
+    let mut child = Command::new(executable);
+    child
+        .args(env::args_os().skip(1))
+        .env(FOREGROUND_ENV, "1")
+        .env(LAUNCH_TOKEN_ENV, &token)
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null());
+    configure_detached(&mut child);
+    let mut process = child
+        .spawn()
+        .map_err(|error| format!("start detached monitor: {error}"))?;
+    let deadline = Instant::now() + LAUNCH_TIMEOUT;
+    loop {
+        if let Ok(status) = fs::read_to_string(&status_path) {
+            let _ = fs::remove_file(&status_path);
+            if status.trim() == "ready" {
+                println!("{} {}", process.id(), paths.root.display());
+                return Ok(());
+            }
+            return Err(status
+                .strip_prefix("error: ")
+                .unwrap_or(status.trim())
+                .to_owned());
+        }
+        if Instant::now() >= deadline {
+            return Err(format!(
+                "monitor did not become ready within {} seconds; inspect {}",
+                LAUNCH_TIMEOUT.as_secs(),
+                paths.log.display()
+            ));
+        }
+        if let Ok(Some(status)) = process.try_wait() {
+            return Err(format!("monitor exited before becoming ready ({status})"));
+        }
+        thread::sleep(Duration::from_millis(50));
+    }
+}
+
+#[cfg(unix)]
+fn configure_detached(command: &mut Command) {
+    use std::os::unix::process::CommandExt;
+    unsafe {
+        command.pre_exec(|| {
+            if libc::setsid() == -1 {
+                return Err(io::Error::last_os_error());
+            }
+            Ok(())
+        });
+    }
+}
+
+#[cfg(unix)]
+fn configure_producer(command: &mut Command) {
+    use std::os::unix::process::CommandExt;
+    unsafe {
+        command.pre_exec(|| {
+            if libc::setpgid(0, 0) == -1 {
+                return Err(io::Error::last_os_error());
+            }
+            Ok(())
+        });
+    }
+}
+
+#[cfg(windows)]
+fn configure_producer(command: &mut Command) {
+    use std::os::windows::process::CommandExt;
+    const CREATE_NEW_PROCESS_GROUP: u32 = 0x0000_0200;
+    const CREATE_SUSPENDED: u32 = 0x0000_0004;
+    command.creation_flags(CREATE_NEW_PROCESS_GROUP | CREATE_SUSPENDED);
+}
+
+#[cfg(unix)]
+fn resume_producer(_child: &std::process::Child) -> io::Result<()> {
+    Ok(())
+}
+
+#[cfg(windows)]
+fn resume_producer(child: &std::process::Child) -> io::Result<()> {
+    use std::mem::{size_of, zeroed};
+    use windows_sys::Win32::Foundation::{CloseHandle, INVALID_HANDLE_VALUE};
+    use windows_sys::Win32::System::Diagnostics::ToolHelp::{
+        CreateToolhelp32Snapshot, Thread32First, Thread32Next, TH32CS_SNAPTHREAD, THREADENTRY32,
+    };
+    use windows_sys::Win32::System::Threading::{OpenThread, ResumeThread, THREAD_SUSPEND_RESUME};
+
+    let snapshot = unsafe { CreateToolhelp32Snapshot(TH32CS_SNAPTHREAD, 0) };
+    if snapshot == INVALID_HANDLE_VALUE {
+        return Err(io::Error::last_os_error());
+    }
+    let mut entry: THREADENTRY32 = unsafe { zeroed() };
+    entry.dwSize = size_of::<THREADENTRY32>() as u32;
+    let mut found = unsafe { Thread32First(snapshot, &mut entry) } != 0;
+    while found && entry.th32OwnerProcessID != child.id() {
+        found = unsafe { Thread32Next(snapshot, &mut entry) } != 0;
+    }
+    unsafe {
+        CloseHandle(snapshot);
+    }
+    if !found {
+        return Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "could not find the suspended producer thread",
+        ));
+    }
+    let thread = unsafe { OpenThread(THREAD_SUSPEND_RESUME, 0, entry.th32ThreadID) };
+    if thread.is_null() {
+        return Err(io::Error::last_os_error());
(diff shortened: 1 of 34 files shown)


src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1102 @@
...
+        ));
+    }
+    write_launch_status(paths, launch_token, "ready");
+    let stdout = child.stdout.take().expect("stdout was piped");
+    let (sender, receiver) = mpsc::sync_channel::<Vec<u8>>(1024);
+    thread::spawn(move || {
+        let mut reader = BufReader::new(stdout);
+        loop {
+            let mut line = Vec::new();
+            match reader.read_until(b'\n', &mut line) {
```
morgmart: 🤖 **P1 · Bound unterminated producer output** (blocking)

The stdout reader calls read_until(newline) before handing data to the bounded channel. A producer can continuously emit bytes without a newline, causing the local line buffer to grow without limit; the later delivery-size cap cannot engage until a delimiter or EOF arrives.

**User effect:** A long-running command that emits a very large unterminated line can exhaust the detached monitor's memory, losing monitoring and potentially destabilizing the user's machine instead of waking the session.

**Recommended fix:** Read producer stdout in bounded chunks, or enforce a per-record limit and split oversized unterminated records before enqueueing them, while preserving UTF-8-safe delivery chunking.

**Test:** Add a discriminating monitor test whose producer writes substantially more than the delivery limit without a newline and verify output is bounded, persisted, and delivered in chunks without unbounded buffering.

src/features/chat/ui/MessageBubble.tsx
```diff
@@ -1007,7 +1010,11 @@ export const MessageBubble = memo(function MessageBubble({
                   data-role="berdctl-cross-session-message-label"
                   className="leading-4"
                 >
-                  {t("message.berdctlCrossSessionLabel")}
+                  {berdSenderLabel
```
morgmart, line 1013: 🤖 **P1 · Preserve trusted Berd provenance** (blocking)

When a caller supplies --from, transcript chrome replaces the trusted 'Sent by Berd from another session' provenance with 'Sent by {{sender}}'. The sender string is only length/single-line validated and does not authenticate identity, so any berdctl caller can make trusted-looking chrome claim attribution such as 'Sent by Morgan' or 'Sent by Security'.

**User effect:** Users can be misled about who initiated a cross-session prompt and may trust or act on agent-generated content as though it came from a named person or trusted system.

**Recommended fix:** Always retain the product-authenticated Berd/cross-session provenance and present the caller-controlled value only as an explicitly untrusted descriptor, for example 'Sent by Berd from another session · source: {{sender}}'.

**Test:** Update the named-label component test to require both the trusted Berd provenance and the supplied descriptor, and include a person-like label to prove it cannot replace the authenticated source.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1102 @@
...
+        atomic_write(&paths.pending, pending)?;
+        log_line(paths, "delivered one event batch")?;
+    }
+    Ok(())
+}
+
+fn deliver(session_id: &str, prompt: &str, if_running: RunningMode) -> bool {
+    for lock in lock_candidates() {
+        for binary in berdctl_candidates() {
+            let status = Command::new(&binary)
```
morgmart: 🤖 **P2 · Keep stop requests responsive** (non-blocking)

Delivery synchronously tries each discovered lock against each berdctl candidate with a 10-second timeout, without checking the stop marker between attempts. Since sibling control files are unbounded, stale locks can make a flush block for many timeout intervals before the main loop observes an explicit stop.

**User effect:** After asking a monitor to stop, users may see the underlying command continue running for a long and unpredictable period.

**Recommended fix:** Bound or prune lock candidates, deduplicate executable attempts, and make delivery cancellation-aware so stop requests interrupt or bypass remaining probes.

**Test:** Add a test with multiple non-responsive lock candidates, request stop during delivery, and assert the producer tree is terminated within a fixed short bound.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1102 @@
...
+    let temporary = path.with_extension(format!("tmp.{}", std::process::id()));
+    fs::write(&temporary, data)?;
+    if path.exists() {
+        fs::remove_file(path)?;
+    }
+    fs::rename(temporary, path)
+}
+
+fn log_line(paths: &StatePaths, message: &str) -> io::Result<()> {
+    writeln!(
```
morgmart, line 1358: 🤖 **P2 · Replace pending output atomically** (non-blocking)

The pending-file rewrite writes a temporary file, removes the existing durable file, and then renames the temporary file. A crash, forced termination, or rename failure in that gap removes the only durable copy of undelivered output.

**User effect:** A monitor interrupted while recording delivery progress can silently lose output that should be retried when Berd becomes available again.

**Recommended fix:** Use a platform-appropriate atomic replacement that does not delete the destination first, keep the temporary file in the same directory, and apply the required file/directory flushes for the intended durability guarantee.

**Test:** Inject a replacement failure after the temporary write and verify the previous pending file remains readable and retryable.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the exact three-dot comparison introduces two blocking issues and two non-blocking reliability issues. Static review covered every changed flow and file, including monitor launch/stop/delivery and persistence, cross-session create/send/steer/queue/replay behavior, sender-label rendering, packaging on Unix and Windows, accessibility, localization, navigation/consent boundaries, async/error/never-completes/lifecycle/race behavior, test honesty, and project/design-system rules. Findings were consolidated by underlying issue and anchored only to current changed lines. Supplied GitHub evidence was inspected: DCO, frontend checks/unit tests, Rust checks on Windows/macOS/Linux, and clean-room install passed; Transcript virtualization and Frontend build smoke were still in progress at capture time. Required checks independently govern merge readiness.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Fixed in f982eb7e: producer output is now read in fixed-size chunks and flushed at a 40,000-byte bound even when the producer never emits a newline. Added a regression test for an unbounded unterminated record.

src/features/chat/ui/MessageBubble.tsx
johnmatthewtennant: 🤖 Fixed in 1201f012: custom  labels no longer replace trusted provenance. The UI now renders , with an updated person-like-label regression test and localized copy.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Fixed in 006cd245: delivery probes are bounded to eight lock candidates, deduplicated, and polled so a stop request kills a nonresponsive  probe promptly. Added coverage for both responsiveness and candidate bounds.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Fixed in dc22c923: pending output is written and synced to a temporary file, then atomically replaced without deleting the previous file first (native write-through replacement on Windows and rename plus parent sync on Unix). A failure-injection test verifies the prior payload survives replacement failure.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1355 @@
...
+            if status.trim() == "ready" {
+                println!("{} {}", process.id(), paths.root.display());
+                return Ok(());
+            }
+            return Err(status
+                .strip_prefix("error: ")
+                .unwrap_or(status.trim())
+                .to_owned());
+        }
+        if Instant::now() >= deadline {
```
morgmart, line 343: 🤖 **P1 · Terminate timed-out monitor launch** (blocking)

When the detached monitor does not publish its launch-status file within ten seconds, the launcher returns an error without terminating or reaping the child it spawned. The child can become ready after the caller has already been told startup failed, so reported startup state can disagree with the actual monitor lifecycle.

**User effect:** An agent can retry after an apparent startup failure while the original monitor is still running, causing duplicate-owner errors or leaving an apparently nonexistent command alive and sending events.

**Recommended fix:** On launch timeout, terminate and wait for the detached child before returning failure, and remove the launch-status artifact associated with the attempt.

**Test:** Add a discriminating test with a foreground child that stays alive but never publishes readiness; verify startup reports failure, the child is reaped, and no stale status file remains.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1355 @@
...
+    let mut batch = Vec::new();
+    let mut batch_deadline: Option<Instant> = None;
+    let mut retry_deadline = Instant::now();
+    let mut producer_status = None;
+    let mut receiver_closed = false;
+    let mut termination_requested = false;
+
+    loop {
+        if paths.stop.exists() && !termination_requested {
+            log_line(paths, "stop requested")?;
```
morgmart, line 800: 🤖 **P1 · Clean up producers on errors** (blocking)

After the producer process group starts, several logging, persistence, delivery, and polling operations can return through `?`. On Unix, dropping `ProducerTree` and `Child` does not kill the process group, so such an error exits the monitor and releases its owner lock while the producer and descendants remain alive. Windows has kill-on-job-close behavior, but Unix has no equivalent unconditional cleanup guard.

**User effect:** A filesystem or monitor-state error can make monitoring disappear while the watched command continues indefinitely; a subsequent stop request cannot reach the orphaned process tree.

**Recommended fix:** Give producer-tree ownership an unconditional cross-platform cleanup guard, or route every post-spawn exit through one termination-and-wait path that kills and reaps the entire process group.

**Test:** On Unix, inject a failure after producer spawn in a fallible logging or pending-write path and verify both the direct child and a spawned descendant are terminated and reaped.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1355 @@
...
+                Ok(child) => child,
+                Err(_) => continue,
+            };
+            loop {
+                if paths.stop.exists() {
+                    let _ = child.kill();
+                    let _ = child.wait();
+                    return false;
+                }
+                match child.try_wait() {
```
morgmart, line 1114: 🤖 **P1 · Bound delivery subprocess lifetime** (blocking)

The monitor passes a ten-second request timeout to berdctl but enforces no deadline on the berdctl subprocess itself. If that process hangs before or outside its internal request timeout, this polling loop never exits unless a stop marker appears, permanently blocking delivery and retries.

**User effect:** A hung delivery process can stop session wakeups forever and eventually backpressure producer stdout, causing the monitored command itself to stall without an automatic recovery path.

**Recommended fix:** Track a monitor-owned deadline for every delivery child; on expiry, kill and reap it, then continue to the next candidate or return to the bounded retry cycle.

**Test:** Use a fake berdctl that never exits without issuing a stop request; assert delivery returns after the monitor-owned deadline and the child is reaped.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1355 @@
...
+                io::ErrorKind::InvalidData,
+                "monitor delivery prompt exceeded its internal size bound",
+            ));
+        }
+        if !deliver_to_session(paths, session_id, &prompt, if_running) {
+            log_line(paths, "delivery failed; buffered output will be retried")?;
+            return Ok(());
+        }
+        pending.drain(..end);
+        atomic_write(&paths.pending, pending)?;
```
morgmart: 🤖 **P1 · Make event delivery idempotent** (blocking)

A pending chunk is sent to the session before its durable pending record is removed. If the monitor crashes, loses power, or fails the subsequent atomic write after berdctl successfully creates the user turn, the old pending file still contains that chunk and a restarted monitor sends it again. Reversing these operations would instead lose events, so the boundary needs an idempotency contract.

**User effect:** One completion event can create duplicate user turns, causing an agent to repeat work or rerun an actionable operation such as a deployment response.

**Recommended fix:** Assign each persisted delivery chunk a stable event ID and make receiver-side session delivery idempotent by recording accepted IDs; acknowledge and remove a chunk only after that deduplicated acceptance.

**Test:** Inject failure after a successful session send but before pending acknowledgement, restart delivery from the persisted record, and verify the same event ID creates only one user turn.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found four new blocking monitor lifecycle and delivery issues. All changed files and user flows were reviewed, including monitor launch, ownership, producer execution, stop, delivery, retries, durable buffering, and packaging; cross-session create/send/steer/queue/drain/replay and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; duplicate memory and overlap. The four supplied prior threads are resolved with substantive human-account replies and their fixes remain present. Supplied GitHub evidence was structurally valid and inspected: all listed check runs passed, while combined status remained pending at capture time. Required checks independently govern merge readiness.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Accepted and fixed in abf9cf4d. Launch timeout now requests stop, waits for graceful exit, force-terminates and reaps the detached child if necessary, and removes both launch-status and stop artifacts. Added a regression test that verifies the child is gone and no stale status remains.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Accepted and fixed in e698da2f. Producer ownership now has an unconditional cross-platform cleanup guard that terminates the process tree and waits for the child on every post-spawn error path. Added a Unix regression test covering both the direct child and a descendant.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Accepted and fixed in 9ac883fe. Each delivery subprocess now has a monitor-owned 15-second deadline; expiry kills and reaps it before trying another candidate or returning to retry. Added a hanging fake-berdctl regression test.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Accepted and fixed in 92ff33c3. Persisted chunks now retain a stable delivery ID until acknowledgement, and session send records/deduplicates that ID across queued and transcript-replayed messages. A retry after successful send but failed local acknowledgement is therefore a successful no-op. Added monitor persistence/rotation coverage plus command and ACP replay deduplication tests.

src/features/berdctl/bridge/useBerdctlQueuedMessageDrain.ts
```diff
@@ -110,7 +110,13 @@ function drainQueuedMessage(queuedSessionId: string, ownerId: string): void {
       );
       assertQueuedSessionReady(state.getSessionRuntime(queuedSessionId));
     },
-    { returnOnDispatch: true },
+    {
+      returnOnDispatch: true,
+      ...(queuedMessage.payload.sendOptions?.userMessageMetadata
```
morgmart, line 128: 🤖 **P1 · Preserve queued delivery IDs** (blocking)

The queued-message drain forwards stored sendOptions only when berdSenderLabel exists. A cross-session send with --delivery-id but no --from therefore loses berdDeliveryId when the queued record becomes dispatchable. Once committed without that ID, subsequent retries cannot recognize the accepted delivery and create another user turn.

**User effect:** A monitor event queued behind a running session can be delivered more than once, causing the agent to repeat work or rerun an action even though the sender supplied an idempotency key.

**Recommended fix:** Forward sendOptions whenever the queued message contains Berd delivery metadata, including berdDeliveryId, rather than gating the entire object on the optional sender label.

**Test:** Queue a cross-session message with a delivery ID and no sender label, drain it, and assert the background send receives user and ACP metadata containing the original delivery ID; retry the same ID and verify only one user turn exists.

src/features/berdctl/commands/impl/sendSession.ts
```diff
@@ -88,10 +112,21 @@ Result:
...
     ]);
+    const sendOptions = berdctlCrossSessionSendOptions({
+      senderLabel: args.from,
+      deliveryId: args.delivery_id,
+    });
 
     await loadSessionForBerdctl(args.session_id);
     const session = requireSession(args.session_id);
 
+    if (
```
morgmart: 🤖 **P1 · Serialize delivery ID admission** (blocking)

The idle-send deduplication check is a read performed before asynchronous background-send acquisition and message commit. Two concurrent sends with the same delivery ID can both observe no accepted message, then both acquire or proceed through separate attempts and commit duplicate user turns. The check therefore does not provide the advertised at-most-once admission guarantee under races.

**User effect:** Concurrent retries of one monitor event can create duplicate prompts, causing duplicate agent responses or repeated external work.

**Recommended fix:** Reserve each (session ID, delivery ID) inside the same serialized dispatch or queue-admission boundary that commits the message, and release the reservation only when admission fails before commit.

**Test:** Issue two concurrent idle sends with the same delivery ID, pause both after their initial lookup, then release them together and verify exactly one is admitted and one reports deduplicated.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1697 @@
...
+    session_id: &str,
+    if_running: RunningMode,
+) -> io::Result<()> {
+    while !pending.bytes.is_empty() {
+        if pending.active_len == 0 {
+            pending.active_len = pending_chunk_end(&pending.bytes, MAX_DELIVERY_BYTES);
+            persist_pending(&paths.pending, pending)?;
+        }
+        let end = pending.active_len;
+        let text = String::from_utf8_lossy(&pending.bytes[..end]);
```
morgmart: 🤖 **P1 · Sanitize NUL output bytes** (blocking)

Producer bytes are converted with String::from_utf8_lossy, which preserves embedded NUL characters, and the resulting prompt is passed directly as a process argument. Rust rejects arguments containing NUL before spawning berdctl. That failure is treated like a transient candidate failure while the same pending chunk remains durable, so the monitor retries it forever and cannot advance to later output.

**User effect:** A monitored command that emits a NUL byte can permanently prevent its completion and all subsequent output from reaching the owning session, leaving the detached monitor stuck retrying after the producer exits.

**Recommended fix:** Sanitize or visibly encode embedded NUL bytes before constructing the berdctl argument while preserving the rest of the producer output and delivery ID.

**Test:** Feed a pending chunk containing an embedded NUL through delivery and verify it is transformed into a valid argument, accepted once, and followed by later queued output rather than retried forever.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found three new blocking correctness issues in delivery deduplication and monitor output handling. The review covered every changed file and user flow, including monitor launch, ownership, producer cleanup, stop, bounded delivery, retry, durable buffering and idempotency; cross-session create/send/steer/queue/drain/replay and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; and duplicate overlap. All eight supplied threads were read: the four resolved threads have substantive human replies, and the four unresolved findings are fixed in the current comparison, so none is repeated or suppressed as a still-present issue. Supplied GitHub evidence was structurally valid and inspected: DCO passed and the remaining listed checks were in progress at capture time. Required checks independently govern merge readiness.

src/features/berdctl/bridge/useBerdctlQueuedMessageDrain.ts
johnmatthewtennant: 🤖 Accepted and fixed in 748a6b01. The queued drain now forwards stored send options when either a sender label or delivery ID is present, so delivery metadata survives admission without requiring --from. Added drain coverage for a delivery ID with no sender label; the command-level deduplication tests verify retries create one admitted turn.

src/features/berdctl/commands/impl/sendSession.ts
johnmatthewtennant: 🤖 Accepted and fixed in c9295c89. Admission now atomically reserves each session/delivery-ID pair before any async dispatch or queue path and holds that reservation until admission settles. Added a concurrent idle-send regression that releases both initial lookups together and verifies one dispatch, one deduplicated result, and one total admitted turn.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Accepted and fixed in 571ae7da. Monitor prompt construction now replaces embedded NUL bytes with the visible Unicode replacement character before spawning berdctl, without increasing the bounded chunk length. Added coverage that delivers the sanitized event exactly once and then successfully delivers later output.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1769 @@
...
+    let paths = StatePaths::for_key(state_key, session_id);
+    ensure_private_directory(&paths.root)?;
+    let _owner = claim_owner(&paths).inspect_err(|error| {
+        write_launch_status(&paths, launch_token, &format!("error: {error}"));
+    })?;
+    let result = (|| {
+        paths
+            .stop
+            .exists()
+            .then(|| fs::remove_file(&paths.stop))
```
morgmart: 🤖 **P1 · Do not erase fresh stops** (blocking)

The foreground monitor publishes ownership by acquiring the owner lock, then unconditionally removes the stop marker before starting the producer. Once the lock is held, `berd-monitor stop` treats the monitor as active and can write a fresh marker in the interval before this removal; startup then deletes that valid request and launches the producer anyway.

**User effect:** A user or agent can receive confirmation that stop was requested while the newly launched monitored command continues running, potentially until another stop request is made.

**Recommended fix:** Clear stale stop state before publishing ownership, or generation-stamp stop requests so startup removes only a marker proven to predate the current launch and honors any request written after ownership becomes visible.

**Test:** Add a deterministic race test that pauses startup after owner-lock acquisition, writes the stop marker through the stop path, resumes startup, and verifies the producer never starts or is immediately terminated.

src/features/chat/ui/ChatInput.tsx
```diff
@@ -107,7 +107,11 @@ function stripCrossSessionOrigin<T extends Record<string, unknown>>(
     return metadata;
   }
 
-  const { origin: _origin, ...rest } = metadata;
+  const {
```
morgmart, line 110: 🤖 **P2 · Clear IDs on manual edit** (non-blocking)

Manual editing strips the cross-session origin and sender label but leaves berdDeliveryId in the queued message metadata. The edited message is therefore treated as user-authored while temporarily reserving the automation delivery ID. Replay restores that ID only for cross-session origins, so the reservation disappears after restart and retry behavior changes across the lifecycle.

**User effect:** A retry may be deduplicated before restart but accepted as a second turn afterward, and a user-authored edit can unexpectedly block the original monitor event ID during the current run.

**Recommended fix:** When converting a cross-session queued message into a manual user edit, remove berdDeliveryId from both user-message and ACP metadata along with origin and sender attribution; alternatively preserve all cross-session provenance consistently if edits are intended to remain automation deliveries.

**Test:** Edit a queued message carrying origin, sender label, and delivery ID; assert all cross-session delivery metadata is removed from both metadata objects and that retry behavior is consistent before and after replay.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found one new blocking lifecycle race and one new non-blocking metadata consistency issue. The review covered every changed file and flow: monitor launch, ownership, producer cleanup, stop, bounded delivery, retries, persistence and idempotency; cross-session create/send/steer/queue/drain/replay, manual edits, and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; and duplicate overlap. All eleven supplied threads were read; each is resolved with a substantive human-account reply and its discussed fix is present, so none is repeated or suppressed as a still-present issue. Supplied GitHub evidence was structurally valid and inspected: DCO, Windows and Linux Rust checks, and clean-room install passed; macOS Rust and frontend checks were in progress at capture time. Required checks independently govern merge readiness.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Fixed in 8271666a. Stop markers are now scoped to the active owner generation, so a request written after ownership becomes visible cannot be erased by startup. Detached launch timeout targets that same generation, and a deterministic test pauses immediately after owner claim, requests stop through the public stop path, then verifies the producer never starts and the marker is acknowledged.

src/features/chat/ui/ChatInput.tsx
johnmatthewtennant: 🤖 Fixed in 2935acd3. Converting a queued cross-session delivery into a manual edit now removes berdDeliveryId together with origin and berdSenderLabel from both user and ACP metadata. The edited-queue test now seeds all three fields in both metadata objects and verifies only unrelated metadata survives.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,1879 @@
...
+}
+
+fn forward_producer_output<R: Read>(mut reader: R, sender: mpsc::SyncSender<Vec<u8>>) {
+    let mut read_buffer = [0_u8; 8 * 1024];
+    let mut record = Vec::with_capacity(MAX_DELIVERY_BYTES);
+    loop {
+        let read = match reader.read(&mut read_buffer) {
+            Ok(0) => break,
+            Ok(read) => read,
+            Err(_) => return,
```
morgmart: 🤖 **P1 · Surface stdout read failures** (blocking)

The producer-output forwarding thread treats every stdout read error as a normal end of stream and returns without reporting it to the monitor loop. Any bytes buffered in the partial record are lost, subsequent producer output is no longer captured, and the monitor can later report the producer's exit as though capture completed normally.

**User effect:** A transient or terminal stdout read failure can silently drop actionable command output, including the result the owning session was waiting for, while the monitor appears to have completed normally.

**Recommended fix:** Send data-or-error events through the channel. On a read error, preserve any already buffered bytes, log and surface the capture failure, and terminate the producer cleanly rather than converting the error into EOF.

**Test:** Use a reader that returns bytes and then an injected error; verify buffered bytes are retained, the monitor records or delivers an explicit failure, and the producer cleanup path runs instead of reporting normal capture completion.

src/features/berdctl/commands/runtime/sessionSend.ts
```diff
@@ -23,22 +24,76 @@ export { isBerdctlCrossSessionQueuedMessage } from "@/features/chat/lib/queuedMe
...
   };
 }
 
+export function hasAcceptedBerdctlDelivery(
+  sessionId: string,
+  deliveryId: string,
+): boolean {
+  const chatStore = useChatStore.getState();
+  return (
+    (chatStore.messagesBySession[sessionId] ?? []).some(
```
morgmart: 🤖 **P1 · Deduplicate after transcript hydration** (blocking)

Delivery reservation checks only messages currently hydrated in the renderer store and the current queue. `loadSessionForBerdctl` loads session information, not transcript messages; a cold session after restart or eviction can therefore reserve a delivery ID that already exists in persisted history. The transcript is hydrated later by the background-send path, but the delivery ID is not checked again before committing the new user message.

**User effect:** Retrying a previously accepted monitor event after restart or transcript eviction can create a duplicate user turn, violating the at-most-once promise of --delivery-id and potentially repeating agent work.

**Recommended fix:** Keep the in-memory reservation for concurrency control, but perform an accepted-delivery check after transcript hydration and immediately before message commit; return deduplicated if persisted history contains the same ID.

**Test:** Start with no messages hydrated, make background acquisition restore a transcript containing the same delivery ID, retry that ID, and verify no new message is dispatched or queued. The test must fail if only the pre-hydration store check exists.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found two new blocking reliability issues. The review covered every changed file and flow: monitor launch, ownership, producer cleanup, stdout capture, stop, bounded delivery, retries, persistence and idempotency; cross-session create/send/steer/queue/drain/replay, cold-session hydration, manual edits, and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; and duplicate overlap. All thirteen supplied threads were read; each is resolved with a substantive human-account reply and its discussed fix is present, so none is repeated or suppressed as a still-present issue. Supplied GitHub evidence was structurally valid and inspected: DCO, Linux Rust checks, and clean-room install passed; frontend, Windows Rust, and macOS Rust checks were in progress at capture time. Required checks independently govern merge readiness.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Fixed in e3f20252. Producer stdout forwarding now emits typed data-or-error events. A read failure first preserves any partial record, then records and delivers an explicit capture-failure summary, terminates and reaps the producer process tree, and returns an error instead of reporting normal completion. Tests cover bytes followed by an injected read error and verify the failure log plus direct-child and descendant cleanup.

src/features/berdctl/commands/runtime/sessionSend.ts
johnmatthewtennant: 🤖 Fixed in 9744ca21. The existing in-memory reservation still serializes concurrent retries, and idle sends now re-check the hydrated transcript after background acquisition and preparation, immediately before entering the message dispatch path. A restored matching delivery returns deduplicated without calling ACP, adding a message, or queueing another turn. The cold-transcript test is empty before hydration and restores the accepted ID inside loadSessionMessages.

src/features/berdctl/bridge/useBerdctlQueuedMessageDrain.ts
```diff
@@ -110,7 +110,14 @@ function drainQueuedMessage(queuedSessionId: string, ownerId: string): void {
       );
       assertQueuedSessionReady(state.getSessionRuntime(queuedSessionId));
     },
-    { returnOnDispatch: true },
+    {
```
morgmart, line 126: 🤖 **P1 · Deduplicate stale queued deliveries** (blocking)

The queued-message drain now forwards berdDeliveryId but does not apply the accepted-delivery check or the post-hydration validation used by the direct-send path. If a user turn is accepted and persisted but the app exits before its queued record is dismissed, restart recovery can hydrate both the accepted transcript message and the stale queued record. The drain then sends that same delivery again.

**User effect:** A monitor event can create two user turns after an app restart, causing duplicate agent responses or repeated external work despite the delivery ID's at-most-once promise.

**Recommended fix:** Before dispatching a queued delivery, check the hydrated transcript for its delivery ID and dismiss only that matching stale record when it was already accepted. Also pass a post-acquisition transcript validator to close the cold-session hydration race before commit.

**Test:** Add a restart-recovery drain test with both an accepted transcript message and a stale queued record carrying the same delivery ID; verify the queued record is removed without dispatch. Add a cold-hydration variant where the accepted transcript appears during acquisition and still produces no second turn.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found one new blocking restart-recovery issue in queued delivery deduplication. The review covered every changed file and flow: monitor launch, ownership, producer cleanup, stdout capture and failures, stop, bounded delivery, retries, persistence and idempotency; cross-session create/send/steer/queue/drain/replay, cold-session hydration, manual edits, and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; and duplicate overlap. All fifteen supplied threads were read; each is resolved with a substantive human-account reply and its discussed fix is present, so none is repeated or suppressed as a still-present issue. Supplied GitHub evidence was structurally valid and inspected: DCO, frontend checks/unit tests, and clean-room install passed; build smoke, transcript virtualization, and Rust checks were in progress at capture time. Required checks independently govern merge readiness.

src/features/berdctl/bridge/useBerdctlQueuedMessageDrain.ts
johnmatthewtennant: 🤖 Fixed in 52c271da. The queued drain now checks transcript-only acceptance before dispatch and removes only the still-current matching stale record. Delivery-bearing drains also pass the same post-hydration validator used by direct sends; if history appears during acquisition, the expected dedupe rejection dismisses that record and continues FIFO draining without a second turn. Tests cover both already-hydrated restart recovery and acceptance during cold hydration.

src-tauri/crates/berd-monitor/src/main.rs
```diff
@@ -0,0 +1,2019 @@
...
+impl ProducerProcess {
+    fn terminate_and_wait(&mut self) {
+        terminate_producer_tree(&mut self.child, &self.tree);
+        let _ = self.child.wait();
+    }
+}
+
+impl Drop for ProducerProcess {
+    fn drop(&mut self) {
+        self.terminate_and_wait();
```
morgmart, line 565: 🤖 **P1 · Disarm producer cleanup once** (blocking)

ProducerProcess::terminate_and_wait sends termination to the stored process tree and waits, but it does not record that cleanup completed. Normal completion, explicit stop, and capture-failure paths call this method directly; Drop then calls it again. On Unix the second call unconditionally sends SIGKILL to the negative stored process-group ID after the original child has been reaped. If that numeric process-group ID is reused between calls, cleanup can signal an unrelated process group.

**User effect:** Finishing or stopping a monitored command can unexpectedly kill unrelated processes belonging to the user if the operating system reuses the old process-group ID before the duplicate cleanup signal.

**Recommended fix:** Make cleanup idempotent by tracking a terminated/disarmed state or consuming the process-tree guard. After the first termination and wait, Drop must become a no-op while still guaranteeing cleanup on early error paths.

**Test:** Inject or wrap the process-tree termination operation with a call counter, invoke explicit cleanup and then drop the guard, and assert termination is issued exactly once; retain a separate test proving Drop alone still cleans up an active producer tree.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh full review of the exact three-dot comparison found one new blocking process-lifecycle issue. The review covered every changed file and flow: monitor launch, ownership, producer cleanup, stdout capture and failures, stop, bounded delivery, retries, persistence and idempotency; cross-session create/send/steer/queue/drain/replay, restart recovery, cold-session hydration, manual edits, and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; and duplicate overlap. All sixteen supplied threads were read; each is resolved with a substantive human-account reply and its discussed fix is present, so none is repeated or suppressed as a still-present issue. Supplied GitHub evidence was structurally valid and inspected: macOS, Windows, Linux, frontend checks/unit tests, clean-room install, and DCO passed; transcript virtualization and frontend build smoke were in progress at capture time. Required checks independently govern merge readiness.

src-tauri/crates/berd-monitor/src/main.rs
johnmatthewtennant: 🤖 Fixed in 0ddd1953. ProducerProcess now owns the process-tree guard as an Option and takes it before signaling or waiting, so explicit cleanup, repeated cleanup, and Drop can terminate the stored tree at most once. The new injected call-counter test proves explicit cleanup disarms both a second call and Drop; the existing early-error test still proves Drop alone terminates and reaps the producer plus descendant.

morgmart (approved):
🤖 Automated code review

APPROVE: the fresh full review of the exact three-dot comparison found no new publishable findings. The review covered every changed file and flow: monitor launch, ownership, producer cleanup, stdout capture and failures, stop, bounded delivery, retries, persistence and idempotency; cross-session create/send/steer/queue/drain/replay, restart recovery, cold-session hydration, manual edits, and provenance; accessibility; English and Spanish localization; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system and project-law compliance; and duplicate overlap. All seventeen supplied threads were read; each is resolved with a substantive human-account reply and its discussed fix remains present, with no regression or still-existing issue requiring suppression. Supplied GitHub evidence was structurally valid and inspected: all listed checks passed. Required checks independently govern merge readiness.

PR #248, opened 2026-08-28
fix(chat): restore table vertical scroll and cell word wrapping

mrohan-sq:
## Summary

Tables in chat responses were cutting off text. Two CSS issues combined to cause this:

1. **`overflow-y: hidden`** on the table scroll container overrode Streamdown's default `overflow-y: auto`, completely disabling vertical scrolling. Combined with Streamdown's default `maxHeight: 300px`, any table taller than 300px had its bottom rows clipped with no way to scroll.

2. **No `overflow-wrap` on `td` cells** — long unbreakable strings (URLs, file paths, code identifiers) caused horizontal overflow instead of wrapping within cells.

PR #207 (smaller wrapping headers) made this worse: wrapping headers narrow columns → body text wraps more → rows get taller → more content falls below the 300px cutoff.

### Changes
- `overflow-y: hidden` → `overflow-y: auto` on the table scroll container (`globals.css`)
- Added `overflow-wrap: break-word` to `td` cells (`globals.css`)
- Updated `globals.streamdown-table.test.ts` to assert the new behavior

### Related issue
none found

### Testing
- All existing table-related tests pass (2 table style tests + 2 scrollbar tests + 9 message tests)
- Pre-commit hooks passed (biome, typecheck, i18n, design-system checks, berdctl contract)

### DIFF
--- a/src/shared/styles/globals.css
+++ b/src/shared/styles/globals.css
@@ -1980,7 +1980,7 @@ html[data-window-kind="voice-buddy"] #root {
 
 [data-streamdown="table"] td {
   font-size: 0.8125rem;
-  overflow-wrap: break-word;
+  overflow-wrap: anywhere;
 }
 
 /* Streamdown paints its table scroll container with bg-background. Keep the
--- a/src/shared/styles/globals.css
+++ b/src/shared/styles/globals.css
@@ -1967,7 +1967,7 @@ html[data-window-kind="voice-buddy"] #root {
 }
 
 [data-streamdown="table-wrapper"] > div:has(> [data-streamdown="table"]) {
-  overflow-y: hidden;
+  overflow-y: auto;
   padding-block-end: var(--streamdown-table-scrollbar-block-size, 0px);
 }
 
@@ -1980,6 +1980,7 @@ html[data-window-kind="voice-buddy"] #root {
 
 [data-streamdown="table"] td {
   font-size: 0.8125rem;
+  overflow-wrap: break-word;
 }
 
 /* Streamdown paints its table scroll container with bg-background. Keep the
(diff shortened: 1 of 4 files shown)


src/shared/styles/globals.css
```diff
@@ -1980,6 +1980,7 @@ html[data-window-kind="voice-buddy"] #root {
 
 [data-streamdown="table"] td {
   font-size: 0.8125rem;
+  overflow-wrap: break-word;
```
kalvinnchau: 🤖 `overflow-wrap: break-word` does not reduce a table cell’s intrinsic minimum width under automatic table layout, so long unbroken URLs, paths, or identifiers can still expand the table and scroll horizontally instead of wrapping. I reproduced this in Chromium with a 100-character token: the table remained 749px wide inside a 300px container. `overflow-wrap: anywhere` kept it at 300px and wrapped the token. Could we use `anywhere` here and, ideally, cover the rendered behavior with a browser/layout test? The current test only verifies that the CSS declaration exists.

mrohan-sq:
🤖 Good catch — updated in 084d619f.

Changed `overflow-wrap: break-word` → `overflow-wrap: anywhere`, which participates in min-content sizing so long unbreakable strings wrap within the cell instead of expanding the table.

Also added a Playwright browser layout test (`table-overflow.spec.ts`) that verifies:
1. A 200-char unbreakable string in a `td` wraps within a 300px container (no horizontal expansion)
2. A 50-row table scrolls vertically (scrollHeight > clientHeight, `overflow-y: auto`)

The browser test uses `page.setContent()` with inline CSS so it runs from a standalone Playwright config (`playwright.table-overflow.config.ts`) without needing a built `dist/`.

kalvinnchau (approved):

PR #252, opened 2026-08-31
feat: remote SSH sessions (experiment)

damienrj:
Experiment-gated **remote SSH sessions**: pick an SSH host and a folder on it, and the session's agent runs there while the Berd UI stays local — the Codex-desktop model. Behind the `remote-ssh-sessions` experiment.

## How it works

Berd starts `goose serve` on the remote host as a detached daemon over SSH, then reaches it through a local port-forward, so the renderer speaks ACP to `ws://127.0.0.1:<port>` exactly as it does locally. Because goosed spawns the ACP bridges itself, remoting it remotes every agent at once, and all file IO and tool execution happen natively on the remote host.

- **Bootstrap over ssh stdin**, so no script text, secret, or user path ever lands in remote argv. It reuses a healthy daemon or starts a fresh one.
- **The daemon outlives the tunnel**: losing the network or closing the laptop drops only the tunnel. A supervisor reconnects with backoff and sessions resume.
- **Hosts** come from `~/.ssh/config` plus anything you type; auth stays OpenSSH's job (key/agent). A doctor probe reports what's missing on a host.
- **Local-only surfaces degrade** for remote sessions: no terminal or git/worktree UI, artifacts render a "file on `<host>`" placeholder, local path resolution is bypassed.

Two changes worth a closer look: the ACP client became a **per-backend connection registry**, and remote session ids are **namespaced** (`ssh:<host>#<wireId>`) because goose ids are only unique per backend.

## Limitations

- A turn that is streaming when the connection drops is cancelled by current goose (transcripts persist and resume). Upstream fix: aaif-goose/goose#11668 — until it ships, a per-host goose-binary override lets a host run a patched build.
- Key/agent ssh auth only, no Windows remotes, and goose must already be installed on the host.
- The model picker shows the local backend's inventory; remote agents apply their own config.

## Validation

`just check`, `just test`, `just tauri-check`, and `just clippy` pass, plus new Rust tests that exercise the real bootstrap script. Manually verified against a Blox workstation: connect, remote session inside a project, tunnel kill → auto-reconnect → resume, app restart → rehydrate, and (with the patched goose) a turn surviving a mid-turn disconnect.


src/shared/api/acpApi.ts
```diff
@@ -194,8 +228,12 @@ export async function forkSession(
   }
 
   const response = await client.unstable_forkSession(params);
+  // The fork lives on the source session's backend; its renderer id is the
+  // composite of that backend and the returned wire id.
+  const forkedSessionId = compositeSessionId(backendId, response.sessionId);
+  registerSessionBackend(forkedSessionId, backendId, response.sessionId);
   return {
-    sessionId: response.sessionId,
+    sessionId: forkedSessionId,
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in df579267. Session mapping now infers remoteHost from composite SSH session IDs, and addSession persists inserted remote sessions. This keeps both UI and berdctl forks remote-aware and durable across restarts, with regression coverage.

src/features/chat/stores/remoteSessionPersistence.ts
```diff
@@ -0,0 +1,189 @@
...
+ * through the normal session-load path.
+ *
+ * The chat session store is imported lazily to keep this module free of an
+ * import cycle with `chatSessionStore`, which calls the persistence writers
+ * above.
+ */
+export async function rehydrateRemoteSessions(): Promise<void> {
+  const records = readRemoteSessionRecords().filter(
+    (record) => !record.archivedAt,
+  );
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in df579267. Remote-session rehydration now restores archived records with their archivedAt state, so they remain available in session history after restart. Added regression coverage for both persistence-level rehydration and the archive/restart store flow.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,263 @@
...
+  case "$rec_port" in
+  '' | *[!0-9]*) return 1 ;;
+  esac
+  [ -n "$rec_secret" ]
+}
+
+# Terminates the pid from the last read_record, TERM then KILL.
+stop_recorded_daemon() {
+  if kill -0 "$rec_pid" 2>/dev/null; then
+    kill -TERM "$rec_pid" 2>/dev/null || true
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in df579267. Daemon records now store an OS process-start identity and revalidate it before TERM, throughout the wait, and before KILL. Older unverifiable records are discarded without signaling their PID. Added real-script tests proving a mismatched PID remains alive and valid daemons still reuse/restart/shutdown correctly.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -0,0 +1,686 @@
...
+        daemon_reused: daemon_info.reused,
+    };
+
+    let generation = {
+        let mut shared = slot.shared.lock().expect("slot poisoned");
+        shared.generation += 1;
+        shared.daemon = Some(daemon_info);
+        shared.local_port = Some(local_port);
+        shared.tunnel_pid = tunnel.child.id();
+        shared.state = state.clone();
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in a310cb41. Establishment now carries the generation it observed before async setup and can publish Ready only if that generation still owns the slot. Superseded tunnels are killed instead, with a regression test covering stale generation advancement.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,309 @@
...
+
+STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/berd/remote"
+RECORD="$STATE_DIR/daemon.record"
+LOG="$STATE_DIR/goose-serve.log"
+RECORD_FORMAT="v3"
+
+emit() { printf '%s %s\n' "$NONCE" "$*"; }
+
+b64() { printf %s "$1" | base64 | tr -d '\n'; }
+unb64() { printf %s "$1" | base64 --decode; }
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in a310cb41. The bootstrap now decodes with portable base64 -d, supported by GNU coreutils and macOS/BSD, instead of the GNU-only --decode long option.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -0,0 +1,686 @@
...
+fn spawn_supervisor(
+    app: AppHandle,
+    slot: Arc<HostSlot>,
+    mut tunnel: tunnel::TunnelProcess,
+    generation: u64,
+    prior_attempts: u32,
+) {
+    tauri::async_runtime::spawn(async move {
+        let established_at = tokio::time::Instant::now();
+        let status = tunnel.child.wait().await;
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in a310cb41. The supervisor now clears the exited tunnel PID immediately after wait returns, guarded by both generation and PID ownership so an old supervisor cannot clear a replacement tunnel. Added regression coverage for current versus stale supervisors.

src/app/lib/chatRuntimeStartup.ts
```diff
@@ -330,6 +332,19 @@ async function startChatRuntime(
...
     );
+    // After the local page load so mergeAcpSessionPage cannot race the
+    // placeholder insert; the merge itself is additive for remoteHost either
+    // way.
+    if (getExperiment(REMOTE_SSH_SESSIONS_EXPERIMENT_ID)?.enabled) {
+      try {
+        const { rehydrateRemoteSessions } = await import(
+          "@/features/chat/stores/remoteSessionPersistence"
+        );
+        await rehydrateRemoteSessions();
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in a310cb41. AppShell now subscribes to the remote-session experiment at runtime and serializes reconciliation after session startup. Enabling rehydrates persisted remote sessions; disabling removes their in-memory state and backend routes, disconnects their hosts, and preserves persistence for later re-enable. Added enable/disable tests.

src/features/chat/ui/ChatInputToolbar.tsx
```diff
@@ -321,7 +336,9 @@ export function ChatInputToolbar({
               </DropdownMenuItem>
               <DropdownMenuItem
                 onSelect={() => onAttachFolders?.()}
-                disabled={disabled}
+                // Local folder attachments are meaningless for a session that
+                // runs its backend on a remote host.
+                disabled={disabled || Boolean(selectedRemoteHost)}
               >
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in f6576615. Remote sends and steers now sanitize attachments at the dispatch boundary: local files and directories are dropped, while byte-backed images keep their content with the local path stripped. Selecting a remote host also clears unsafe drafts and disables local attachment ingress. Added regression coverage.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -0,0 +1,748 @@
...
+pub async fn shutdown(
+    app: &AppHandle,
+    registry: &RemoteBackendRegistry,
+    host_input: &str,
+) -> Result<(), RemoteBackendError> {
+    let aliases = ssh_config::load_ssh_config_hosts();
+    let spec = RemoteHostSpec::parse(host_input, &aliases)?;
+    let shell_env = dir_env::capture_home_interactive_env().await;
+    daemon::shutdown_daemon(&spec, &shell_env).await?;
+    disconnect(app, registry, &spec.key());
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in f6576615. Shutdown now acquires the host slot connect lock, invalidates and disconnects the active generation, then stops the daemon while still holding that lock. Establish and reconnect cannot recreate the daemon during shutdown.

src/app/AppShell.tsx
```diff
@@ -719,9 +727,34 @@ export function AppShell({
...
+      remoteSessionReconciliationRef.current
+        .catch(() => undefined)
+        .then(async () => {
+          await runChatRuntimeStartup();
+          const { reconcileRemoteSessionsForExperiment } = await import(
+            "@/features/chat/stores/remoteSessionPersistence"
+          );
+          await reconcileRemoteSessionsForExperiment(
+            remoteSshSessionsExperimentEnabled,
+          );
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in f6576615. Experiment reconciliation now runs through a shared per-window hook used by both AppShell and SessionWindowApp. When disabled, a remote session window immediately stops rendering its remote session while shared reconciliation removes routes and disconnects hosts. Added hook and session-window tests.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -0,0 +1,748 @@
...
+            );
+            set_state(&app, &slot, RemoteBackendState::Disconnected);
+            return;
+        }
+
+        for attempt in start_attempt..=MAX_RECONNECT_ATTEMPTS {
+            set_state(
+                &app,
+                &slot,
+                RemoteBackendState::Reconnecting {
```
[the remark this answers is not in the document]
damienrj: 🤖 Addressed in f6576615. Reconnecting and terminal supervisor states now use a generation-guarded transition. A stale supervisor returns without overwriting Disconnected, and status emission is ordered under the same slot mutex. Added a regression test.

src/shared/api/acpConnection.ts
```diff
@@ -159,73 +281,64 @@ async function initializeConnection(): Promise<GooseClient> {
...
-  return client;
+function createBackendConnection(backendId: AcpBackendId): AcpConnection {
+  if (backendId === LOCAL_BACKEND_ID) {
+    return createAcpConnection(resolveLocalWsUrl);
+  }
+  const host = remoteHostFromBackendId(backendId);
+  if (!host) {
+    throw new Error(`Unknown ACP backend id: ${backendId}`);
+  }
+  return createAcpConnection(async () => {
```
kalvinnchau: 🤖 **[P2] Enforce experiment disablement at the transport boundary.** This resolver reconnects every `ssh:*` backend without checking whether remote sessions are enabled. Renderer reconciliation is not authoritative: composite session IDs reconstruct SSH routing after unregister, and non-UI callers such as `berdctl session send --session-id 'ssh:<host>#<id>'` can reach ACP with a stale ID. Gate remote backend acquisition here or in the Tauri command, including cached clients, and invalidate remote ACP clients when the experiment is disabled.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,311 @@
...
+  umask 077
+  mkdir -p "$STATE_DIR" || {
+    emit "ERR state-dir"
+    exit 46
+  }
+
+  resolve_goose_bin
+  b64binary="$(b64 "$goose_bin")"
+
+  if [ -f "$RECORD" ] && read_record; then
```
kalvinnchau, line 411: 🤖 **[P2] Serialize daemon record mutation across Berd clients.** The local Rust `connect_lock` protects only one process. Two app instances or laptops using the same remote Unix account can both pass this record check, start separate daemons, and overwrite the single record; the shipped `shutdown` then kills only the recorded PID. This was reproduced against this script at the reviewed head. Protect both `ensure` and `shutdown` with one portable per-account remote lock, and write the record atomically.

src-tauri/src/services/remote_backend/tunnel.rs
```diff
@@ -0,0 +1,169 @@
...
+            let mut lines = BufReader::new(stderr).lines();
+            while let Ok(Some(line)) = lines.next_line().await {
+                let redacted = redact_log_line(&line);
+                log::warn!("[remote-backend tunnel stderr] {redacted}");
+                if let Ok(mut tail) = tail.lock() {
+                    tail.push_str(&redacted);
+                    tail.push('\n');
+                    if tail.len() > 8 * 1024 {
+                        let excess = tail.len() - 8 * 1024;
+                        tail.drain(..excess);
```
kalvinnchau: 🤖 **[P3] Avoid draining a `String` at an arbitrary byte offset.** `len()` is bytes, while `drain` requires a UTF-8 character boundary. An 8193-byte tail beginning with a multibyte character makes `excess == 1` and panics this detached reader task, losing SSH failure diagnostics. Trim at the next valid `char_indices()` boundary or bound whole bytes/lines, with a multibyte regression test.

kalvinnchau (changes requested):
🤖 Requesting changes for two transport/process-ownership defects. I also left one non-blocking diagnostic finding inline. The follow-up fixes for shutdown ordering, stale supervisor state, cross-window reconciliation, attachment filtering, and status initialization hold at this head.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,311 @@
...
+    if recorded_process_is_current; then
+      if port_listening "$rec_port" && [ "$rec_b64binary" = "$b64binary" ]; then
```
kalvinnchau: 🤖 **[P2] Include the effective launch specification in daemon identity.** Reuse compares only the Goose path, so a second client can inherit an incompatible daemon. This is reachable today across supported platforms: packaged Windows requests `http://tauri.localhost`, Unix requests `tauri://localhost`, and Goose's explicit origin allowlist contains only the first launch's value. The second client's ACP WebSocket is rejected; different Goose overrides instead make clients mutually evict daemons forever, even with a lock. Key the daemon by binary plus normalized serve args/protocol contract, or return an ownership conflict.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,311 @@
...
+  while [ "$attempt" -lt 5 ]; do
+    attempt=$((attempt + 1))
+    port=$(((RANDOM % 40000) + 20000))
+    if port_listening "$port"; then
+      continue
+    fi
+    # Detached: nohup + fully redirected stdio survives the ssh session ending
+    # (no PTY is allocated, and non-interactive bash does not forward SIGHUP).
+    # shellcheck disable=SC2086
+    GOOSE_SERVER__SECRET_KEY="$secret" nohup "$goose_bin" serve --host 127.0.0.1 --port "$port" $extra_args >>"$LOG" 2>&1 </dev/null &
```
kalvinnchau: 🤖 **[P2] Bound the lifetime log before spawning the daemon.** This process intentionally survives tunnels, app exits, and updates, but every stdout/stderr byte is appended to one file with no rotation or size cap. A noisy long-running agent can exhaust the remote account's disk and impact unrelated work. Rotate/truncate with bounded retention or use a bounded logger.

kalvinnchau: 🤖 **[P2] Clean up a child that has not committed its record.** Goose is detached with `nohup` before its identity is recorded. The local 60-second command timeout can cancel this bootstrap during the script's roughly 75-second five-attempt readiness budget, leaving a live PID that future ensure/shutdown calls cannot identify. Install a trap that terminates the current uncommitted PID, clearing it only after atomic record commit; make the outer timeout exceed the script's bounded runtime.

kalvinnchau (changes requested):
🤖 Consolidated red-team delta: the existing experiment, cross-client record-race, and UTF-8 comments remain valid. These additional inline findings tighten the daemon ownership and operational contracts. The transport design remains viable; no SSH/ACP rewrite is requested.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,505 @@
...
+      return 0
+    fi
+
+    if lock_owner_is_current; then
+      stale_observations=0
+    else
+      stale_observations=$((stale_observations + 1))
+      # Allow the mkdir winner time to publish its owner before treating a
+      # missing/malformed owner as stale after an interrupted invocation.
+      if [ "$stale_observations" -ge 20 ]; then
```
kalvinnchau, line 247: 🤖 **[P2] Stale-lock recovery can wedge forever or destroy a newer lock generation.** A crash after creating `.owner.$$` but before renaming it leaves a non-empty lock directory with no valid `owner`; this recovery removes only `owner`, so `rmdir` cannot remove the partial artifact and future operations eventually fail `state-dir`. Concurrent stale observers also perform an unowned check-then-delete: after one removes the stale directory and a new process acquires it, a delayed observer can delete the new owner's file/directory, collapsing mutual exclusion. Both paths were reproduced. Atomically rename/claim the observed lock generation before deleting it, clean every partial publication artifact within the claimed directory, and make the wait budget cover the bounded critical section. Add crash-at-publication and synchronized multi-reclaimer tests.

src/shared/api/acpConnection.ts
```diff
@@ -92,49 +119,153 @@ function createClientCallbacks(): () => Client {
...
-        "[acp] Connection error. Will reconnect on next getClient().",
+    if (stream) {
+      await stream.writable.abort();
+    }
+  }
+
+  async function initializeConnection(): Promise<GooseClient> {
+    assertAvailable?.();
+    const tStart = performance.now();
+    const wsUrl = await resolveWsUrl();
```
kalvinnchau: 🤖 **[P2] Disabling the experiment during setup can leave a live forbidden transport.** While this await is unresolved, `activeStream` is still null, so `invalidate()` clears the shared promise but cannot abort the initializer. It can later create and initialize a WebSocket; the rejection path clears only `clientPromise` and does not close the initializer-owned stream or Rust tunnel. Re-enabling before completion can also publish work that crossed a disable boundary. Give each initialization an invalidation generation/cancellation token checked after every await, and close every locally owned stream/tunnel on rejection or supersession. Add a regression test that disables while remote connect is unresolved, then resolves it and proves no client survives.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
```diff
@@ -0,0 +1,453 @@
...
+                // Failures surface via the store's failed status for this row.
+                void ensureHostConnected(host).catch(() => {});
+              }}
+            >
+              {state === "connecting"
+                ? t("remoteHosts.status.connecting")
+                : t("remoteHosts.actions.connect")}
+            </Button>
+          )}
+          {isConnected ? (
```
kalvinnchau: 🤖 **[P2] A launch-spec conflict has no usable product recovery path.** The script correctly refuses to evict an incompatible live daemon, but after restart or a documented Goose override change, Connect returns `daemon-conflict` and the row becomes disconnected/Failed. This shutdown action is rendered only while `isConnected`, leaving only another Connect that deterministically fails. Keep ordinary Connect non-destructive, but expose an explicit confirmed stop/takeover action for conflicts and persisted disconnected daemons, with enough instance metadata to target the conflicting daemon safely.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -0,0 +1,505 @@
...
+  else
+    rm -f "$log_tmp"
+  fi
+}
+
+# Consume the daemon's pipe for its whole lifetime while keeping only a bounded
+# tail on disk. Each append reopens the path so atomic trims take effect.
+bounded_log_writer() {
+  writer_log="$1"
+  while IFS= read -r writer_line || [ -n "$writer_line" ]; do
```
kalvinnchau: 🤖 **[P2] Log retention is still unbounded for one record or a no-newline stream.** `read -r` buffers until newline, then the complete record is appended before trimming. A huge line therefore consumes unbounded shell memory and transient disk; a continuous no-newline stream never reaches trimming. Consume fixed-size byte chunks and rotate/truncate independently of newline boundaries. Test both a large record's peak/final bounds and continued bounding while a no-newline producer remains open.

kalvinnchau (changes requested):
🤖 Consolidated re-review: four P2 lifecycle/resource defects remain. The v3/v4 coexistence concern is withdrawn because the prior writer never shipped. The underlying SSH/ACP architecture remains sound; these are bounded ownership and recovery corrections.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
damienrj: 🤖 Addressed in 9b8ee3a1. Failed launch-spec conflicts and persisted disconnected hosts now expose a confirmed stop action. Conflict errors carry inspected PID/version/binary metadata plus an opaque process-identity token; takeover stops only that exact daemon generation and then reconnects. If the daemon changed after confirmation, shutdown refuses without signaling it.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Addressed in 9b8ee3a1. The lifetime logger now consumes fixed 64 KiB byte chunks and trims before each append, keeping the on-disk log at or below 4 MiB without buffering newline records. Real-script tests assert peak/final bounds for a huge record and for a no-newline producer that remains open.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Addressed in 9b8ee3a1. Stale lock generations are now serialized through an atomic claim-and-rename before exact-path cleanup, including partial owner publications. The wait budget now exceeds the bounded critical section, and real-script regressions cover crash-at-publication plus synchronized multi-reclaimers without deleting a successor lock.

src/shared/api/acpConnection.ts
damienrj: 🤖 Addressed in 9b8ee3a1. ACP initialization now carries an invalidation generation checked after each async boundary. Superseded setup aborts its owned stream and disconnects only the Rust tunnel generation it created, so it cannot tear down a newer replacement. The regression test disables while remote connect is unresolved and proves the stale setup leaves no client or transport.

kalvinnchau (dismissed):

kalvinnchau (approved):

PR #255, opened 2026-09-01
fix(sidebar): always show "View all chats" when more sessions exist

### LINKED ISSUE #256: "View all chats" link hides itself exactly when session loading fails or returns few sessions
### Before filing

### What's broken
The sidebar's "View all chats" link (which opens Session History) is only rendered when enough sessions are already loaded. When session loading fails or returns very few sessions — the exact failure mode where the user most needs a route to older chats — the link never appears, leaving no in-app path to Session History.

### Steps to reproduce
1. Launch Berd
2. Be in a state where the first session-list page returns few or zero sessions (e.g. a backend `session/list` error, or fewer than 30 sessions loaded while `hasMoreSessions` is true)
3. Look for the "View all chats" row at the bottom of the sidebar chat list

The "View all chats" affordance should be available whenever there may be more sessions on the backend (`hasMoreSessions`), or unconditionally, since Session History is a legitimate destination regardless of how many chats the sidebar currently shows.

### What actually happened
The link is gated on loaded-session count. Flat sidebar path: `showViewAllInHistory={hasFlatChatOverflow}` (`src/features/sessions/ui/session-list/SidebarProjectsSection.tsx:296`), where `hasFlatChatOverflow` requires `flatSessionCandidates.length >= MAX_FLAT_SIDEBAR_CHATS` (30) (`src/features/sessions/capabilities/SessionListCapability.tsx:891-893`). The grouped path has the same flaw via `hasMoreSessions && hasVisibleChats` (`SidebarProjectsSection.tsx:209-212`) when zero sessions load. Counting loaded rows to decide whether to show the link *to the place that loads more rows* is circular: it fails closed exactly when it should fail open.

### How often does it happen?
Every time — reliably reproducible (whenever loaded session count is below the threshold)

### Berd version
0.6.2

### Operating system
macOS (Apple Silicon)

### Model and provider
n/a

### Relevant log output
no relevant log output (UI gating logic; no error is logged by the link itself)

Surfaced while diagnosing a user whose sidebar showed only ~2 chats after an update; they had no "View all chats" link and therefore no UI route to Session History. Fix: drop the loaded-count dependency in `hasFlatChatOverflow` / the grouped guard so the link renders whenever `hasMoreSessions` is true (or always).

kalvinnchau:
**Category:** fix
**User Impact:** Users whose chats fail to load in the sidebar now always see the "View all chats" link into Session History, instead of losing the escape hatch exactly when they need it.

**Problem:** When session loading doesn't reach a user's chats — e.g. BOT-1688, where scheduler-generated sessions push real chats past the first page — the sidebar shows empty projects AND hides the "View all chats" link, leaving no UI path to sessions that are still intact on disk. Both sidebar paths gated the link on loaded-session counts: the flat path required ≥30 loaded candidates before `hasMoreSessions` could surface it, and the grouped path required `hasVisibleChats` alongside `hasMoreSessions`. Counting loaded rows to decide whether to show the link to the place that loads more rows is circular — the link hides itself exactly when loading falls short.

**Solution:** Show the link whenever the backend reports more sessions (`hasMoreSessions`), regardless of how many are loaded. Brand-new users still never see it: with zero sessions there is no second page and `hasMoreSessions` is false. This is Fix 0 (immediate unblock) from the BOT-1688 handoff; the durable fixes (per-project query, server-side search) remain follow-ups.

<details>

**src/features/sessions/capabilities/SessionListCapability.tsx**
Removed the `flatSessionCandidates.length >= MAX_FLAT_SIDEBAR_CHATS` term from `hasFlatChatOverflow` so `hasMoreSessions` alone surfaces the flat-mode link.

**src/features/sessions/ui/session-list/SidebarProjectsSection.tsx**
Removed the `hasVisibleChats` dependency from `showGroupedHistoryLink` so `hasMoreSessions` alone surfaces the grouped-mode link; updated the comment to explain why loaded-count gating is circular.

**src/features/sessions/ui/session-list/__tests__/SidebarFlatChatsSection.test.tsx**
Added regression tests for both paths: flat mode shows the link with a single loaded chat, and grouped mode shows it with zero loaded chats but more pages on the backend (the BOT-1688 condition). Verified the grouped test fails against the pre-fix code.

</details>

Closes #256

Refs BOT-1688 (Linear)

### DIFF
--- a/src/features/sessions/capabilities/SessionListCapability.tsx
+++ b/src/features/sessions/capabilities/SessionListCapability.tsx
@@ -889,8 +889,7 @@ export function SessionListCapability({
   );
 
   const hasFlatChatOverflow =
-    flatSessionCandidates.length > MAX_FLAT_SIDEBAR_CHATS ||
-    (flatSessionCandidates.length >= MAX_FLAT_SIDEBAR_CHATS && hasMoreSessions);
+    flatSessionCandidates.length > MAX_FLAT_SIDEBAR_CHATS || hasMoreSessions;
   const standaloneChatCount = groupChatsByProject
     ? projectSessions.standalone.length
     : flatSessionCandidates.length;
--- a/src/features/sessions/ui/session-list/SidebarProjectsSection.tsx
+++ b/src/features/sessions/ui/session-list/SidebarProjectsSection.tsx
@@ -198,18 +198,15 @@ export function SidebarProjectsSection({
   const showChatsEmptyState = projectSessions.standalone.length === 0;
   const showCombinedEmptyState = showProjectsEmptyState && !hasVisibleChats;
   const showProjects = collapsed || projectsSectionOpen;
-  // Only surface the Session History route when the grouped view actually
-  // hides chats: loaded standalone chats were truncated to the recents cap,
-  // or the backend has more sessions than are loaded. The hasMoreSessions
-  // case is gated on having any loaded chats (not standalone ones
-  // specifically): grouped auto-loading is bounded, so a user whose loaded
-  // chats all belong to projects must still get a route to older sessions.
-  // Brand-new users have no chats and no backend pages, so they never see
-  // the link.
+  // Surface the Session History route whenever the grouped view can hide
+  // chats: loaded standalone chats were truncated to the recents cap, or the
+  // backend has more sessions than are loaded. Gating on loaded chat counts
+  // would hide the link exactly when loading failed to reach the user's
+  // chats, which is when the escape hatch is needed most. Brand-new users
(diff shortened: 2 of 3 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison found no publishable findings. The change consistently uses hasMoreSessions as the source of truth for exposing Session History in both flat and grouped sidebar flows, and the added tests discriminate the two prior failure modes. Supplied GitHub evidence was inspected: 9 checks completed successfully, 1 was skipped, and Transcript virtualization remained in progress at capture time; required checks still govern merge readiness.

PR #260, opened 2026-09-01
fix(berdctl): surface goose backend error detail instead of dropping it

### LINKED ISSUE #258: berdctl drops backend error detail (reports only "Internal error")
### Before filing

### What's broken
When a berdctl session command fails because the goose backend returned an error, berdctl reports only the generic ACP error code (e.g. `RequestError: Internal error`) and discards the `.data` payload where goose puts the actual underlying message. This makes backend failures undiagnosable from the CLI output.

### Steps to reproduce
1. Launch Berd
2. Trigger a backend session-list failure (e.g. a session row that fails to decode)
3. Run `berdctl session list --limit 30`

The CLI should print the backend's actual error detail (the contents of the ACP error's `data` field), e.g. the real SQLite/decode message, so the failure is diagnosable.

### What actually happened
```
backend_read_failed: Failed to read sessions from the app backend: RequestError: Internal error
```
`loadAllSessionsForBerdctl` and `loadSessionForBerdctl` wrap the error with `String(error)` (`src/features/berdctl/commands/runtime/sessions.ts:22,42`), which for an ACP/JSON-RPC error yields only the code+message and drops `.data`. Goose's `internal_err()` maps the underlying storage error into `.data` (`crates/goose/src/acp/server.rs:157-159`), so the real cause never reaches the user. The existing `formatAcpErrorMessage` helper (`src/shared/api/acpErrors.ts:31-54`) already surfaces `.data`; these call sites just don't use it.

### How often does it happen?
Every time — reliably reproducible

### Berd version
0.6.2

### Operating system
macOS (Apple Silicon)

### Model and provider
n/a

### Relevant log output
CLI output shown above. The goose-side log contains the real message, e.g.:
```
session/list … data: "error occurred while decoding column \"accumulated_total_tokens\": out of range integral type conversion attempted"
```

Fix is two lines: use `formatAcpErrorMessage(error)` instead of `String(error)` at both call sites in `runtime/sessions.ts`. (Change already prepared on a local branch.)

kalvinnchau:
## Summary
berdctl session/project commands reported only the generic ACP error message ("Internal error") and discarded the error's `.data` payload, where goose puts the real underlying message. The `String(error)` wrap at the backend-read catch sites dropped it. This makes backend failures undiagnosable from the CLI.

This replaces the `String(error)` wraps with `formatAcpErrorMessage` (which already surfaces `.data`) via a single bounded helper `berdctlErrorDetail` (`commands/helpers.ts`) that caps extraction before serializing. Applied at every point a backend error crosses the berdctl wire: `runtime/sessions.ts`, `runtime/projects.ts`, the `toCommandError` bridge boundary, `impl/archiveProject.ts`, and the archive-session facade. Tauri-invoke and UI-notification sites that carry no ACP `.data` payload are left unchanged.

### Related issue
Fixes #258

### Testing
- `pnpm generate:berdctl-contract` ✓ (no contract diff)
- `cargo test -p berdctl` — 52 passed ✓
- `pnpm vitest run src/features/berdctl` — 259 passed ✓
- `just check` ✓, `just test` — 7281 passed ✓, `just lint` ✓, `just typecheck` ✓, `just fmt-check` ✓
- New `helpers.test.ts` + additions to `commands.test.ts`, `berdctlBridge.test.tsx`, `AppShell.berdctl.test.tsx` assert `.data` detail surfaces in the CLI error message
- 3 adversarial review rounds (4 reviewers each), final ratings all ≥ 9/10, no open findings

### DIFF
--- a/src/app/AppShell.berdctl.test.tsx
+++ b/src/app/AppShell.berdctl.test.tsx
@@ -476,7 +476,14 @@ describe("AppShell berdctl integration", () => {
   });
 
   it("archiveSession reports backend failure and keeps the session unarchived", async () => {
-    mockAcpArchiveSession.mockRejectedValue(new Error("backend down"));
+    // ACP-shaped error: the formatted detail must surface the `data` payload,
+    // not String(error)'s "Error: Internal error".
+    mockAcpArchiveSession.mockRejectedValue(
+      Object.assign(new Error("Internal error"), {
+        code: -32603,
+        data: "session row missing",
+      }),
+    );
     useChatSessionStore.setState({ sessions: [makeSession()] });
     render(<AppShell />);
 
@@ -484,7 +491,11 @@ describe("AppShell berdctl integration", () => {
       getAppNavigationController().archiveSession("session-1", "reject"),
     );
 
-    expect(outcome).toEqual({ ok: false, reason: "backend_archive_failed" });
+    expect(outcome).toEqual({
+      ok: false,
+      reason: "backend_archive_failed",
+      detail: "session row missing",
+    });
     expect(
       useChatSessionStore.getState().getSession("session-1")?.archivedAt,
     ).toBeUndefined();
@@ -506,7 +517,11 @@ describe("AppShell berdctl integration", () => {
       getAppNavigationController().archiveSession("session-1", "reject"),
     );
 
-    expect(outcome).toEqual({ ok: false, reason: "backend_archive_failed" });
+    expect(outcome).toEqual({
+      ok: false,
+      reason: "backend_archive_failed",
+      detail: "backend down",
+    });
     expect(useChatSessionStore.getState().activeSessionId).toBe("session-1");
     expect(screen.getByTestId("active-view")).toHaveTextContent("chat");
     expect(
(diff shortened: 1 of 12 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison 1a4c43f3688bfca1eeb4060868dc2452dd9e1f97...1e930a53c9a2b59609ca33b9be304cd73bd50b09 (merge base 37e706def3025557ae4614b7f083e697a47a9bc2) found no publishable issues. Both the project code-review and wes-review passes were applied. Final self-check covered every changed berdctl error-reporting flow (session list/get/archive, project list/archive, and uncaught bridge failures), accessibility, i18n/localization completeness, navigation and consent guards, async/error/never-completes/lifecycle/race behavior, test honesty, project and design-system rules, duplicate overlap, and blocking-evidence standards. The change is non-visual, introduces no navigation or consent behavior, preserves existing failure state, bounds relayed details, and adds discriminating coverage for ACP data payloads and oversized details. Supplied GitHub evidence was inspected: DCO, clean-room install, Semgrep, and zizmor succeeded, while frontend and Tauri platform checks were still in progress; required checks still govern merge readiness.

**Pending checks:** 5 check(s) are not complete.

PR #263, opened 2026-09-01
fix(search): search full session set via server-side session/list query

### LINKED ISSUE #257: Search only covers loaded sessions; should use the server-side session/list query filter
### Before filing

### What's broken
Search (both Cmd-K and Session History) only searches sessions that are already loaded into the renderer. Sessions beyond the first `session/list` page are never searched, so users with many sessions get "no results" for conversations that exist. There is no way to make search cover the full session set.

### Steps to reproduce
1. Launch Berd
2. Have more sessions than one backend page (page size is 50)
3. Open search (Cmd-K) and search for a term that only appears in an older session (one past the first page)

Search should cover all sessions, not just the ones currently loaded into memory.

### What actually happened
No results. Cmd-K sweeps only the loaded `visibleSessions` (`src/features/search/ui/SearchView.tsx:140-170`) with no pagination trigger. Content search (`src/features/sessions/hooks/useSessionSearch.ts` → `src/shared/api/sessionSearch.ts`) exports each loaded session's full JSON client-side to substring-match, which is too expensive to run over the whole set, so it stays limited to the loaded slice. Session History does paginate, but only when the user scrolls (`src/features/sessions/ui/SessionHistoryView.tsx:497-535`), so typing a query without scrolling still only searches loaded pages.

The goose backend already implements server-side full-message-content search: `session/list` accepts a `_meta.query` filter that translates to a SQL `EXISTS` over `json_extract(message.text)` (`crates/goose/src/session/session_manager.rs:363-386`), cursor-paginated and exposed on the ACP wire (`crates/goose/src/acp/server/list_sessions.rs:187,205`). Berd just never sends it — `LIST_SESSIONS_META` only sets `includeLastMessageSnippet` (`src/shared/api/acpApi.ts:60-64`).

### How often does it happen?
Every time — reliably reproducible

### Berd version
0.6.2

### Operating system
macOS (Apple Silicon)

### Model and provider
n/a

### Relevant log output
no relevant log output

Fix: point search at the existing server-side `_meta.query` filter on `session/list` instead of client-side export sweeps. One wire call, server-paginated, full-corpus — no export storm, no scroll-to-load-before-searching. This is a Berd-only change; the backend capability already ships.

kalvinnchau:
## Summary

Cmd-K and Session History search only covered sessions already loaded in the renderer: content search exported each loaded session's full JSON client-side, which is too expensive to run over the whole set — so anything past the first `session/list` page was never searched.

Discovery now runs server-side: `listSessionsPage`/`acpListSessionsPage` send `_meta.query`, which goose translates to a SQL keyword filter over message text with cursor pagination. The client export sweep remains, but only to enrich server-matched loaded sessions with snippet/messageId/matchCount; every server match rows as a content hit even without enrichment.

- `searchSessions` walks the filtered `session/list` pages (cursor cycles and the page cap are errors, never silent truncation) and returns the full match set (`matchedInfos`)
- `useSessionSearch` treats the server match set as authoritative for content membership: per-query retention across page sweeps, generation-gated commits for overlapping sweeps, snippet preservation across failed enrichments, and admission predicates (`includeDiscoveredSession`) so Cmd-K excludes archived sessions and History applies its scope/project filters to discovered sessions
- Selecting a discovered result hydrates it into the session store synchronously (restoring persisted workspace metadata) before activation; store-backed card actions (rename/fork/archive/restore) are hidden until a row is hydrated
- Progress narration now reflects export enrichment only; metadata/title matching, debouncing, and the result model are unchanged

### Related issue

Fixes #257

### Testing

- `just check` — pass (biome, design-system, berdctl contract, i18n, typecheck)
- `just lint` — pass
- `just typecheck` — pass
- `just fmt-check` — pass
- `just test` — pass (full Vitest suite, 7285 passed / 1 skipped), including new coverage: `_meta.query` wire shape, server-driven discovery across pages, OR-keyword enrichment agreement, cursor-cycle/page-cap/mid-pagination errors, zero-target discovery, searchMore carry-over, export-failure degradation, admission predicates, discovered-row selection routing, hydration-before-activation, and store-action gating for non-hydrated rows

Note: `src/app/AppShell.navigation.test.tsx` "focuses a detached chat selected from search" is flaky on main (fails ~50% without these changes) — pre-existing timing issue, unrelated to this fix.

### DIFF
--- a/src/app/AppShell.navigation.test.tsx
+++ b/src/app/AppShell.navigation.test.tsx
@@ -93,6 +93,7 @@ const mockAcpCreateSession = vi.hoisted(() => vi.fn());
 const mockAcpPrepareSession = vi.hoisted(() => vi.fn());
 const mockAcpSetSessionConfigOption = vi.hoisted(() => vi.fn());
 const mockAcpListSessionsPage = vi.hoisted(() => vi.fn());
+const mockAcpSearchSessions = vi.hoisted(() => vi.fn());
 const mockBuildFeatures = vi.hoisted(() => ({
   byoKeyProviders: false,
   voiceConversation: false,
@@ -492,6 +493,7 @@ vi.mock("@/shared/api/acp", () => ({
   acpGetSessionInfo: (...args: unknown[]) => mockAcpGetSessionInfo(...args),
   acpListSessionsPage: (...args: unknown[]) => mockAcpListSessionsPage(...args),
   acpLoadSession: (...args: unknown[]) => mockAcpLoadSession(...args),
+  acpSearchSessions: (...args: unknown[]) => mockAcpSearchSessions(...args),
   discoverAcpProviders: vi.fn().mockResolvedValue([]),
 }));
 
@@ -1003,6 +1005,15 @@ describe("AppShell global navigation", () => {
     mockAcpGetSessionInfo.mockResolvedValue(null);
     mockAcpLoadSession.mockReset();
     mockAcpLoadSession.mockResolvedValue(undefined);
+    mockAcpSearchSessions.mockReset();
+    // Default: the server matches nothing; tests that exercise discovery
+    // override with their own match set.
+    mockAcpSearchSessions.mockImplementation(async () => ({
+      results: [],
+      searchedIds: [],
+      failedIds: [],
+      matchedInfos: [],
+    }));
     mockToastError.mockReset();
     mockListenSessionDeepLinkErrors.mockReset();
     mockListenSessionDeepLinkErrors.mockResolvedValue(vi.fn());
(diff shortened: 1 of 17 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of all 17 files in the exact three-dot comparison found no publishable blocking or non-blocking findings. The code-review and wes-review passes covered server-wide session discovery, result admission and hydration, Cmd-K and Session History selection/navigation, pagination, overlapping searches, failures and partial enrichment, accessibility and localization impact, test honesty, and project laws/design-system rules. Supplied GitHub evidence was inspected; checks remain pending/in progress, including security scanning, so CI still governs merge readiness.

**Completed checks currently reporting failure:**
- Frontend checks and unit tests: failure

PR #265, opened 2026-09-01
feat(berdctl): attach and detach folders on existing projects via berdctl

### LINKED ISSUE #226: berdctl: no way to add a working directory to an existing project
### Before filing
- [x] I am using a recent release (0.6.2) and this doesn't already exist.
- [x] This is one request, not several bundled together.

### Closest existing issue
none found (#63, which added the gated `automation` noun to berdctl, is the nearest precedent for extending a berdctl noun)

### Is this new, or an improvement?
Improvement to the existing `berdctl project` commands

### The problem, in your terms
A project created without a working directory can't be given one from the command line. `project create --working-dir` is the only place that setting is accepted — there is no equivalent for a project that already exists.

### What you do today
I edited the project's state file in `~/.local/share/goose/projects/<id>.md` by hand, adding the `workingDirs` and `projectWorkspaces` keys — copied from a throwaway project I created purely to learn the format, then archived. Berd picked the change up immediately and it works, but hand-editing app state clearly isn't the intended path.

### What you'd like to see
Something like `berdctl project set-working-dir --project-id <id> --path <path>` (repeatable, ideally with a matching removal), so an existing project's folders can be managed the way its startup mode already can via `project set-startup-mode`.

### Why this belongs in Berd itself
berdctl already exposes project create / list / get / set-startup-mode / archive. Working directories are the one project attribute that's write-once-at-creation, which leaves the noun incomplete and pushes users (and the agents driving berdctl on their behalf) toward editing state files by hand.

### Non-goals
Not asking to change multi-workspace behaviour, folder attachment semantics, or anything in the UI — just CLI parity for an existing project attribute.

### Alternatives you considered
Hand-editing the project markdown file (works, fragile); creating a new project with the right folder and moving all sessions into it (loses project identity/groups).

---
*Filed via `gh` (an agent helped investigate and draft; I reviewed before posting). Laid out with the feature-request form's fields since the API path doesn't render the form.*

kalvinnchau:
**Category:** improvement
**User Impact:** Agents and users can add or remove a project's working folders from the command line after the project already exists.
**Problem:** `berdctl project create --working-dir` is the only write path for project folders. A project created without one, or with the wrong one, cannot be fixed from the CLI, so people hand-edit `~/.local/share/goose/projects/<id>.md`.
**Solution:** Add `berdctl project attach-folder` and `berdctl project detach-folder`, matching the existing project-dialog add/remove folder actions and the session `folder attach` / `folder detach` verbs.

Closes #226

## Why this shape

The issue asked for something like `project set-working-dir` with a matching removal. Attach/detach is the better fit: it is reversible, visible in the UI, and already how session folders work. A single "set" verb that both adds and replaces would hide the existing-folder case.

New folders start with startup mode `none`, the same default as the project dialog. Worktree behavior stays on `project set-startup-mode`. We do not refresh Git identity on a repeated attach: that path had already started rewriting startup mode when a checkout disappeared or moved, which is more than this issue asked for. Repeating the command is a no-op before existence or Git probes.

Relative paths are rejected. `resolve_path` would otherwise treat them relative to the app process, not the CLI cwd. `~` is allowed and expanded. A folder stored as `~/src/api` by `project create --working-dir` is treated as the same folder as `/Users/me/src/api`.

Detach does not require the directory to still exist. The folder just has to be on the project. Detaching the last folder is allowed; folder-less projects are already a legal create state.

## Why we did not do some things

- **No project-wide CAS / mutation queue.** Concurrent attach/detach can last-write-wins, same as `project set-startup-mode` and the rest of `editProject`. Fixing that is a store/backend change, not this CLI gap.
- **No `getGitState` cancellation.** The native Git probe still has its own 90s budget. Attach's bridge timeout is 120s so a real attach can finish; we refuse the deadline immediately before persist so a timed-out caller does not still write. Cancelling the sidecar probe is shared infrastructure.
- **No UI, multi-workspace, or folder-attachment semantic changes.** The issue's non-goal.
- **No telemetry.** berdctl mutations stay off the human-only project edit/create events, matching `set-startup-mode` and `agent create`.
- **No protocol bump.** Additive commands and optional fields are not a reshape.

## Commands

```
berdctl project attach-folder --project-id <id> --path ~/src/api
berdctl project detach-folder --project-id <id> --path ~/src/api
```

<details>

**src/features/berdctl/commands/impl/attachProjectFolder.ts**
New attach command: validate an absolute/`~` path, no-op if already attached, otherwise classify Git identity and persist with startup mode `none`.

**src/features/berdctl/commands/impl/detachProjectFolder.ts**
New detach command: expand the requested path, drop the matching stored folder (including stored `~` forms), skip persist when nothing matched.

**src/features/berdctl/commands/runtime/paths.ts**
Shared helpers for `~`/absolute admission, existence checks on an already-resolved path, and comparing a stored path to a resolved one.

**src/features/berdctl/commands/registry.ts**
Register `attach_folder` / `detach_folder` on the project noun and list them in help.

**src/features/berdctl/__tests__/commands/commands.test.ts**
Cover attach, re-attach no-op, relative reject, missing dir, Git fail, live-list merge, stored `~` identity, detach, last folder, missing-on-disk detach, and timeout-before-mutate.

**src-tauri/crates/berdctl/api-surface.json**
**src-tauri/crates/berdctl/cli-surface.json**
**src-tauri/crates/berdctl/api-surface-feedback.json**
**src-tauri/crates/berdctl/cli-surface-feedback.json**
Regenerated contract surfaces. Not hand-edited.

**src-tauri/crates/berdctl/src/tree.rs**
Top-level help lists the new project verbs.

**src-tauri/crates/berdctl/src/main.rs**
Minimal CLI invocations so the wire-mapping test covers the new verbs.

</details>

### DIFF
--- a/src-tauri/crates/berdctl/api-surface-feedback.json
+++ b/src-tauri/crates/berdctl/api-surface-feedback.json
@@ -688,7 +688,7 @@
       }
     },
     "projects": {
-      "description": "Manage the user's projects: create, list, get, set startup mode, archive.",
+      "description": "Manage the user's projects: create, list, get, attach folder, detach folder, set startup mode, archive.",
       "actions": {
         "create": {
           "description": "Create a new project; it appears immediately in the app's project list.",
@@ -769,6 +769,80 @@
               }
             },
             "required": ["project_id"],
+            "additionalProperties": false
+          }
+        },
+        "attach_folder": {
+          "description": "Add an existing folder to a project's configured folders, matching the project dialog's add-folder action. Git checkouts are detected so identity behaves as if the folder was added in the UI. Repeating the command is safe.",
+          "fields": [
+            {
+              "name": "project_id",
+              "required": true,
+              "kind": "string",
+              "description": "Id of the project to update.",
+              "min": 1
+            },
+            {
+              "name": "path",
+              "required": true,
+              "kind": "string",
+              "description": "Existing absolute folder or Git checkout path; ~ is expanded.",
+              "min": 1
+            }
+          ],
+          "schema": {
+            "$schema": "https://json-schema.org/draft/2020-12/schema",
+            "type": "object",
+            "properties": {
+              "project_id": {
+                "type": "string",
+                "minLength": 1,
+                "description": "Id of the project to update."
+              },
+              "path": {
+                "type": "string",
+                "minLength": 1,
+                "description": "Existing absolute folder or Git checkout path; ~ is expanded."
+              }
+            },
+            "required": ["project_id", "path"],
+            "additionalProperties": false
+          }
+        },
+        "detach_folder": {
+          "description": "Remove a folder from a project's configured folders without deleting anything from disk or Git, matching the project dialog's remove-folder action. Detaching the last folder leaves the project without folders, which is allowed.",
+          "fields": [
+            {
+              "name": "project_id",
+              "required": true,
+              "kind": "string",
+              "description": "Id of the project to update.",
+              "min": 1
+            },
+            {
+              "name": "path",
+              "required": true,
+              "kind": "string",
+              "description": "Attached folder path to remove; ~ is expanded.",
+              "min": 1
+            }
+          ],
+          "schema": {
+            "$schema": "https://json-schema.org/draft/2020-12/schema",
+            "type": "object",
+            "properties": {
+              "project_id": {
+                "type": "string",
+                "minLength": 1,
+                "description": "Id of the project to update."
+              },
+              "path": {
+                "type": "string",
+                "minLength": 1,
+                "description": "Attached folder path to remove; ~ is expanded."
+              }
+            },
+            "required": ["project_id", "path"],
             "additionalProperties": false
           }
         },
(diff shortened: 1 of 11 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison 40ef23b50663a4b6a4c48f62c39ee2250e4e88fb...f08b0a2335baf4527220033f1fc744ddf29ca311 found no publishable findings. The attach and detach project-folder flows validate path admission, preserve live project state across asynchronous inspection, avoid mutation after deadline expiry, handle idempotent and missing-folder cases, and include discriminating command tests. Final self-check covered both changed user flows; accessibility and localization (no UI surface or localized UI copy changed); navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; project laws and berdctl architecture rules; duplicate overlap; and blocking-severity evidence/user effect. Supplied GitHub evidence was inspected: it is structurally valid and tied to the exact head SHA, with some checks successful and the combined status still pending; required checks independently govern merge readiness.

PR #266, opened 2026-09-01
feat(chat): order model picker by usage recency

### LINKED ISSUE #228: Let me set the order models appear in the picker
### Before filing
- [x] I am using a recent release (0.6.2) and this doesn't already exist.
- [x] This is one request, not several bundled together.

### Closest existing issue
#122 (Claude effort controls) is the same picker, different control. Nothing found on ordering.

### Is this new, or an improvement?
Improvement

### The problem, in your terms
I use Opus for almost everything and it sits near the bottom of my model list, so every model switch means scanning past models I rarely pick. The order doesn't reflect how I actually use them. From a quick look at the picker code, it appears to fall back to alphabetical by display name once the currently-selected model has been pinned to the top — so how often I use a model has no bearing on where it appears.

### What you do today
Nothing that helps. I scroll to it each time. The current-model-pinned-to-top behaviour only helps for the model I've already selected, which is the opposite of what I need when I'm switching.

### What you'd like to see
Some way for my own ordering to win — whether that's setting it myself or Berd learning it from what I actually use. I don't have a strong view on which.

### Why this belongs in Berd itself
The ordering happens inside Berd's picker, after the harness hands over its model list. A harness-level setting can shorten that list but can't change the order Berd displays it in, so there's no way to solve this from outside.

### Non-goals
- Not asking to add or hide any models.
- Not asking to remove the current-model-pinned-to-top behaviour. It's useful, just not sufficient.
- Not asking for this to sync across machines.

### Berd version
0.6.2

### Operating system
macOS (Apple Silicon)

---
*Filed via `gh` (an agent helped investigate and draft; I reviewed before posting). Laid out with the feature-request form's fields since the API path doesn't render the form.*

kalvinnchau:
Closes #228.

## What

The model picker pinned the current model on top, then fell back to provider → harness `sortOrder` → alphabetical. Usage had no bearing on order, so a heavily-used model (the reporter's Opus) sat near the bottom. The issue asked for either a user-set order or one learned from actual use — this implements the learned option.

- **Record at the single chokepoint.** `AgentModelPicker.handleModelSelect` now records each selection (per agent, with provider) into a small bounded localStorage map via a new `src/features/chat/lib/modelRecency.ts`. One call site covers every picker surface (chat toolbar, global composer pill).
- **Order by recency.** In `AgentModelPickerLists`, the full-list sort inserts recency right after the current-model pin: recently-used models first (newest first), then the existing provider / `sortOrder` / name fallbacks for never-used models.
- **Recent models join the compact view.** Up to 3 recently-used models fold into the recommended shortlist ahead of harness recommendations, so a frequently-used non-recommended model is reachable without "View more."

Non-goals respected: current-model pin preserved, no sync across machines, no settings UI.

## Module design (`modelRecency.ts`)

- `useSyncExternalStore` + a cached snapshot, patterned on the existing `autoArchivePreference.ts` (CustomEvent for same-window, `storage` events cross-window, SSR-safe).
- Keys are `encodeURIComponent(agent)/encodeURIComponent(provider)/encodeURIComponent(model)` so slash-bearing ids can't collide.
- Ranks are monotonic (survive tied/regressing clocks), safe-integer validated, normalized at the parse boundary and on every record, and pruned to the 50 highest-ranked entries on read and write.
- Cross-window writes merge by max rank (no last-write-wins data loss); a read landing in the write/event gap also merges so local recency isn't dropped.

## Tests

`modelRecency.test.ts` (17) and a new "model recency" block in `AgentModelPicker.test.tsx` cover: recording, newest-first ordering, the 3-model shortlist cap, cross-agent/provider isolation, slash-id collisions, pruning boundaries, corrupt/oversized storage, quota failure, cross-window merge, safe-integer ceiling edges, and tied-rank determinism.

## Validation

- `just check` — format, lint, i18n, typecheck: pass
- `pnpm vitest run` — 611 files, 7291 tests: pass
- `pnpm build` — pass

## Known limitation

An extreme corner (ceiling-level rank corruption **and** two concurrently-open windows racing a renormalization) can transiently mix rank scales in one window's merged map, mis-ordering it until reload. It produces no data loss, can't arise in normal single-window use, and self-corrects on next launch; judged not worth the scale-rebase complexity.

---

🤖 Implemented with iterative adversarial review (6 rounds, 4 parallel reviewers each).


morgmart (approved):
🤖 Automated code review

APPROVE. Fresh static review of the exact three-dot comparison 40ef23b50663a4b6a4c48f62c39ee2250e4e88fb...e8e950e0af9eecf11e0212a0787918a500e3cc9a found no publishable findings. The changed model-selection flow, compact and expanded picker ordering, per-agent/provider recency identity, localStorage failure handling, cross-window lifecycle, accessibility, localization, navigation/consent behavior, race handling, test honesty, and project/design-system rules were checked. The supplied GitHub evidence was inspected: several checks succeeded, while Frontend checks and unit tests and Tauri Rust checks (macOS) were still in progress at capture time; required checks continue to govern merge readiness.

**Pending checks:** 3 check(s) are not complete.

Issue #171, opened 2026-08-22
Files panel opens Markdown externally instead of in the artifact viewer


AnobleSCM:
## Before filing

- [x] I searched open and closed issues for duplicates.
- [x] I reproduced this on the latest release.
- [x] This report covers one bug.

## Closest existing issue

None found.

## What's broken

Clicking a Markdown file in a chat's right-rail **Files** tab always delegates to macOS. It opens the registered external editor even though Berd already supports Markdown in its in-app artifact viewer.

## Steps to reproduce

1. Launch Berd v0.6.2 on macOS Apple Silicon.
2. Open a chat with a project or workspace attached. No agent turn is required.
3. Open the right rail's **Files** tab.
4. Click `README.md`.

## Expected behavior

`README.md` opens in Berd's sidebar artifact viewer.

## Actual behavior

macOS opens `README.md` in the registered external editor. On this machine, that editor is Xcode.

## Frequency

Every time. The path is deterministic.

## Version

Berd v0.6.2, the latest release as of 2026-08-22.

## Operating system

macOS 15.7.1 on Apple Silicon.

## Model or provider

Not applicable. The bug occurs without an agent turn.

## Relevant logs

No relevant log output. The problem is in the frontend click-routing path.

## Cause and verified repair direction

`FilesList.tsx` calls Tauri's `openPath(path)` for every file. This bypasses `ArtifactPolicyContext.openInApp`.

A local proof patch routes file clicks through `openInApp(path, basename(path))`. Supported Markdown and images then open in Berd, while unsupported files still use the operating system.

Verification completed against that patch:

- The new integration test failed before the change because the artifact-viewer state stayed empty.
- The focused `FilesList` tests pass: 6 of 6.
- The surrounding artifact-viewer tests pass: 49 of 49.
- `just check` passes.
- The full test run passes: 6,776 tests across 575 files, with one existing skip.
- An isolated app-driver run clicked a real fixture `README.md`. It created one `[data-artifact-viewer-panel]`, rendered one Markdown `h1`, matched the exact file in the viewer header, and did not start Xcode.

PR #267, opened 2026-09-01
fix(chat): restore artifact open actions ("Open in editor") broken by provider scoping

badideasforsale:
## Summary

"Open in editor" in the artifact viewer's ⋯ menu did nothing — a silent no-op. This PR fixes the root cause and four adjacent defects in the same failure family, found during pre-review and dogfooding of the fix.

**Root cause:** #178 moved `ArtifactPolicyProvider` from `ChatView` into the extracted `ChatTranscriptSurface`. The transcript kept its provider, but sibling consumers in the chat row — `ArtifactViewerPanel` ("Open in editor"), the right rail's `ArtifactsWidget` (row opens), and `ArtifactAutoOpenMount` (artifact auto-open) — fell outside it and received the inert default context, whose `openResolvedPath` is an async no-op and whose artifact list is empty. Call sites swallow errors (`.catch(() => {})`), so nothing surfaced anywhere.

Seven commits, each independently revertable:

1. **`fix(chat)`: outer `ArtifactPolicyProvider` around ChatView's chat row.** Restores context for the viewer panel, right-rail artifacts widget, and auto-open. The transcript's inner provider stays (Home canvas cards rely on it); nesting is benign — actions are memoized and artifact collection is signature-cached. The outer provider intentionally receives ungated `controller.messages`: the auto-open hook baselines whatever is present while history loads and on its first settled pass, so history artifacts cannot pop the viewer — and baselining the real list is safer than the paint-gated (transiently empty) one.
2. **`fix(tauri)`: session-window opener capability.** Popped-out `session:*` windows only allowed `opener:allow-open-path` for `$DOWNLOAD`, so even with the provider fixed, "Open in editor" was rejected by capability scope in popped-out windows — same silent symptom. The scope now mirrors the main window's.
3. **`fix(home)`: canvas card composer provider.** Same #178 family: `CanvasCardComposer` is a sibling of `ChatTranscriptSurface`, so its @-file mentions silently read an empty session-artifact list. It now gets its own provider fed the full session messages.
4. **`fix(chat)`: failed opens no longer consume the retry debounce.** `openResolvedPath` recorded its 1200 ms dedupe timestamp before awaiting the opener, so a rejected hand-off blocked an immediate retry.
5. **`fix(chat)`: no "Open in editor" for deleted files.** The viewer already knows when the file is gone (the "File deleted from disk." strip), yet the error body still offered an "Open in editor" button and the ⋯ menu kept the action enabled — guaranteed dead clicks on a nonexistent file. The error-body button now hides and the menu item disables when the divergence kind is `missing` (the same signal that already hides the strip's Reload button).
6. **`fix(chat)` (review follow-up): the outer provider is keyed by the effective session.** The outer provider received the raw requested `sessionId` while its messages/cwd came from the controller's effective session; during session replacement those can briefly disagree, letting artifact policy or viewer-store writes act under a stale identity. The provider, `ArtifactAutoOpenMount`, and `ArtifactViewerPanel` now all use `timelineSessionId`, with a discriminating test where the requested and effective ids differ.
7. **`fix(chat)` (review follow-up): Reveal is also disabled for deleted files.** Commit 5 kept Reveal enabled claiming file managers could show the containing folder — but the implementation passes the deleted file's path (not the parent) to `revealItemInDir` and swallows the rejection: another silent dead end. Both hand-offs now disable together and re-enable when polling sees the file reappear (new reappearance test; the deleted-state test verifies clicks reach neither OS boundary).

### Screenshots (deleted-file viewer states)

Captured via the Playwright harness (tauri-mock driven to the `missing` stat state), viewer panel shown at 2x:

| Before (PR head `90cfe94`, as reviewed) | After (current head) |
|---|---|
| ⋯ menu on a deleted file: "Open in editor" disabled but **Reveal still enabled** (dead click) | Both OS hand-offs disabled; re-enabled automatically if the file reappears |

_(images attached in PR comments below)_

Known follow-ups deliberately not in this PR (happy to file issues):
- Surface open/reveal failures to the user (toast) instead of `.catch(() => {})` — that swallowing is what made both the #178 regression and the capability gap invisible.
- Make the default actions context warn or throw in dev; the silent no-op default is what let this regress unnoticed.
- `AutomationBuilderView` renders `MessageTimeline`/`ChatInput` with no artifact provider (artifact links/opens inert there); pre-existing, out of scope here.

### Related issue

None found for this exact bug. Closest: #171 (Files panel opens Markdown externally instead of the viewer) — adjacent surface, different defect; not addressed here.

### Testing

- New regression test: the `ArtifactPolicyProvider` test mock now stamps its props on a structural marker; a ChatView test asserts the viewer panel mounts inside a provider that received real session data. **Fails against the unfixed ChatView** (verified both ways via stash).
- New integration-style tests (`ArtifactViewer.openActions.test.tsx`) render the **real** provider with only the Tauri boundary mocked: viewer "Open in editor" and right-rail artifact row both reach `openPath`; a failed open doesn't consume the retry debounce (also verified to fail pre-fix). The provider-scope bug class is invisible to tests that mock the context, so these keep the real wiring covered.
- `just check` clean; `just tauri-check` clean (capability schema regenerated); full chat + home suites: **3,116 tests pass**.
- Manual verification of "Open in editor" pending a dev build; the wiring is covered by the integration tests above.

Two independent adversarial review passes were run on this branch before opening; findings from both (session-window capability gap, canvas composer orphan, debounce lockout, test-mock drift risks, commit-message inaccuracies) are incorporated.


badideasforsale:
🤖 Rebased onto main (`8e7e350`) after #252 (remote SSH sessions) landed — it touches the same two files (`ChatView.tsx`, `ArtifactPolicyContext.tsx`). Clean rebase, no conflicts; the remote-session guard in `openResolvedPath` composes correctly with the debounce fix (guard throws before the debounce bookkeeping). `just check` clean and full chat+home suites (3,205 tests) pass on the rebased head.

src/features/chat/ui/ChatView.tsx
```diff
@@ -793,7 +794,18 @@ export function ChatView({
...
+    // because siblings of the transcript consume the artifact context too:
+    // ArtifactViewerPanel ("Open in editor"), the right rail's
+    // ArtifactsWidget (row opens), and ArtifactAutoOpenMount (the artifact
+    // list). ChatTranscriptSurface renders its own provider for the
+    // transcript, but without this outer one its siblings get the inert
+    // default context and their actions silently no-op.
+    <ArtifactPolicyProvider
+      messages={controller.messages}
+      sessionCwd={controller.sessionArtifactCwd}
+      sessionId={sessionId}
```
morgmart: 🤖 **P1 · Provider uses stale session identity** (blocking)

ChatView derives timelineSessionId from effectiveSession?.id ?? sessionId and gives that identity to the rendered transcript, but the newly added outer ArtifactPolicyProvider receives the raw requested sessionId while receiving controller messages and cwd. During session replacement or reconciliation, its identity can disagree with the session snapshot it governs. This provider owns artifact behavior for the viewer, rail, and auto-open subtree, including remote/local policy and viewer-store identity.

**User effect:** While a chat is being restored or reconciled, a remote artifact can incorrectly expose a local file action, a valid local action can be blocked, or an artifact can open under the obsolete chat session.

**Recommended fix:** Pass timelineSessionId, or the same effective-session identity used by the timeline, to the outer ArtifactPolicyProvider so its ID, messages, cwd, and remote/local policy describe one session snapshot.

**Test:** Add a discriminating ChatView test where the requested sessionId differs from controller.session.id and assert that the enclosing artifact policy provider receives the effective controller session ID; the current matching-ID test does not catch this case.

src/features/chat/ui/ArtifactViewer.tsx
```diff
@@ -598,7 +598,14 @@ export function ArtifactViewer({
                 />
               </DropdownMenuTrigger>
               <DropdownMenuContent align="end">
+                {/* A deleted file cannot be handed to an editor, so the
+                    action is disabled rather than left as a dead click.
+                    Reveal stays enabled: file managers can still show the
```
morgmart: 🤖 **P1 · Deleted reveal can fail silently** (blocking)

The changed code explicitly keeps Reveal enabled after the artifact is known to be missing, asserting that file managers can show the containing folder. The action still passes the deleted artifact path to revealItemInDir rather than targeting an existing parent directory, and its rejection is swallowed. The implementation therefore does not establish the recovery behavior promised by the new state decision.

**User effect:** After Berd reports that a file was deleted, the remaining enabled Reveal action can appear to do nothing, leaving another silent dead end beside the action this PR removes.

**Recommended fix:** For a missing artifact, either disable or hide Reveal, or implement an explicit containing-folder action that targets an existing parent directory and uses an accurate label.

**Test:** Add a discriminating missing-file test that selects the remaining action and verifies either that it is unavailable or that the file-manager boundary receives the existing parent-directory path. Do not only assert the absence of aria-disabled.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the new outer artifact policy can apply filesystem behavior using the wrong session identity during reconciliation, and the deleted-file menu deliberately retains a Reveal action whose deleted target may fail silently. This PR changes graphical artifact-viewer states but the supplied PR evidence contains no screenshots or recording; please add screenshots or a short screen recording. The supplied GitHub evidence is structurally valid: all captured check runs passed, while combined status remained pending, so required checks still govern merge readiness.

Deterministic publication result: 2 blocking and 0 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 1 blocking screenshot-evidence requirement(s) in this review body.

badideasforsale:
🤖 Both blocking findings addressed, and screenshot evidence added below.

**P1 · Provider uses stale session identity** → fixed in `dcf8b24`. The outer `ArtifactPolicyProvider`, `ArtifactAutoOpenMount`, and `ArtifactViewerPanel` now all key off `timelineSessionId` (the controller's effective session), so identity, messages, cwd, and remote/local policy describe one session snapshot — and the viewer panel reads the same viewer-store entry that `openInApp` writes. Added the requested discriminating test: the requested `sessionId` differs from `controller.session.id`, and both the enclosing provider and the panel must carry the effective id (fails against the previous head).

**P1 · Deleted reveal can fail silently** → fixed in `3dd1684`, taking the disable option. The review is right that the code contradicted the comment — Reveal passes the deleted file's path (not the parent directory) to `revealItemInDir` with the rejection swallowed. Both OS hand-offs now disable together on the shared `fileIsMissing` state, and polling re-enables them if the file reappears (new reappearance test). The deleted-state test now verifies clicks on both disabled items reach neither OS boundary (`openResolvedPath` / `revealInFileManager`), not just `aria-disabled`.

**Screenshots** (deleted-file viewer, captured via the Playwright harness with the tauri-mock driven to a `missing` stat verdict):

*Before — PR head `90cfe94` as reviewed: "Open in editor" disabled but Reveal still enabled (silent dead click):*

before: reveal still enabled on a deleted file

*After — current head: both OS hand-offs disabled:*

after: both hand-offs disabled on a deleted file

*After — error body for a deleted file: "File deleted from disk." strip with no dead "Open in editor" button:*

after: error body without the dead button

All chat + home suites pass (3,210 tests); `just check` clean.

src/features/chat/ui/ChatView.tsx
```diff
@@ -943,7 +959,9 @@ export function ChatView({
         </div>
 
         {sessionId && !isAgentBuilderSession ? (
-          <ArtifactViewerPanel sessionId={sessionId} />
+          // Keyed by the same effective identity the providers use, so the
+          // panel reads the viewer-store entry that openInApp writes.
+          <ArtifactViewerPanel sessionId={timelineSessionId} />
```
morgmart, line 1005: 🤖 **P1 · Viewer layout reads stale session** (blocking)

ArtifactViewerPanel now correctly reads timelineSessionId, but isArtifactViewerOpen is still derived near the top of ChatView with useOpenArtifact(sessionId). During the reconciliation state this follow-up explicitly supports, the panel and the layout calculation therefore read different artifact-store entries. Width allocation and compact-mode decisions can disagree with the viewer actually rendered.

**User effect:** While a chat is being restored or reconciled, an open artifact viewer can appear without reserved space and overlap or compress the chat, while a stale viewer can leave an empty reserved column.

**Recommended fix:** Derive the effective timeline session identity before the viewer-store lookup and use it consistently for useOpenArtifact, ArtifactViewerPanel, and every layout calculation driven by viewer visibility.

**Test:** Extend the reconciliation test so requested and effective session IDs differ, seed viewer state only for the effective session, and assert that the rendered chat layout enters its viewer-open configuration; also verify stale requested-session viewer state does not reserve space.

src/features/chat/ui/ChatView.tsx
```diff
@@ -793,11 +794,26 @@ export function ChatView({
...
+    // ArtifactViewerPanel ("Open in editor"), the right rail's
+    // ArtifactsWidget (row opens), and ArtifactAutoOpenMount (the artifact
+    // list). ChatTranscriptSurface renders its own provider for the
+    // transcript, but without this outer one its siblings get the inert
+    // default context and their actions silently no-op.
+    // The identity, messages, and cwd must describe one session snapshot:
+    // timelineSessionId (the controller's effective session) rather than the
+    // raw requested sessionId, which can briefly disagree with
+    // controller.messages during session replacement or reconciliation.
+    <ArtifactPolicyProvider
```
morgmart, line 846: 🤖 **P2 · Artifact policy has duplicate owners** (non-blocking)

The new ChatView provider wraps a ChatTranscriptSurface that still creates its own provider, and ChatCanvasCard similarly gives its transcript and composer separate providers. Each provider independently derives artifact inventory and owns its own per-path open debounce. The same visible session can therefore process different message projections and bypass duplicate-open suppression when actions originate from different surfaces.

**User effect:** Opening the same artifact quickly from two parts of one chat can hand it to the operating system twice, and artifact availability can drift between the transcript and adjacent viewer, rail, or composer surfaces.

**Recommended fix:** Mount one ArtifactPolicyProvider per rendered session boundary, placing transcript, composer, viewer, auto-open mount, and artifact widgets beneath it. Keep standalone transcript support through an explicit wrapper rather than implicit nested ownership.

**Test:** Add a provider-boundary test proving one policy owner covers all full-chat consumers and one covers both canvas transcript and composer; verify opening one path across two consumer surfaces shares the same debounce lifecycle.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the follow-up fixes both prior blocking issues, but the artifact viewer's layout still reads open state under the requested session while the panel now renders under the effective session, and multiple policy providers give one visible session separate artifact/open lifecycles. Supplied visual evidence is present. All captured GitHub check runs passed; combined status remained pending, so required checks still govern merge readiness.

badideasforsale:
🤖 Both round-2 findings addressed.

**P1 · Viewer layout reads stale session** → fixed in `3887555`. `timelineSessionId` is now derived immediately after `effectiveSession` — *before* the viewer-store lookup — and `isArtifactViewerOpen` reads `useOpenArtifact(timelineSessionId)`, so the panel, providers, and every viewer-driven layout decision (row occlusion, gap, conversation min-width) describe one store entry. All `useOpenArtifact` call sites in ChatView were enumerated (`rg`) — this was the only remaining raw-`sessionId` reader. The reconciliation test now covers layout per your spec: viewer state seeded only under the effective session must reserve the conversation's viewer-open floor, and stale requested-session state must not (fails against the previous head).

**P2 · Artifact policy has duplicate owners** → taken in this PR rather than deferred, in `5f1a940`. `ChatTranscriptSurface` no longer mounts its own provider; the surface composing a session boundary owns exactly one: ChatView's chat row (transcript + viewer panel + right rail + auto-open) and `ChatCanvasCard` (transcript + composer, replacing the composer-only provider). Both requested tests added:
- provider-boundary tests assert one owner covers all consumers on both surfaces (each fails against the nested structure), and
- a cross-surface debounce test opens the same path from the viewer's ⋯ menu and the artifacts widget under one real provider and asserts a single OS hand-off.

`ChatTranscriptSurface` consumers audited (`rg`): ChatView and ChatCanvasCard are the only mounts, and both now provide the context above it — standalone transcript support is explicit wrapping by the composing surface, per your recommendation.

All chat + home suites pass (3,213 tests); `just check` clean.

src/features/chat/ui/ChatView.tsx
```diff
@@ -793,11 +800,26 @@ export function ChatView({
   });
 
   return (
-    <>
+    // The provider must wrap the whole chat row — not just the transcript —
+    // because siblings of the transcript consume the artifact context too:
+    // ArtifactViewerPanel ("Open in editor"), the right rail's
+    // ArtifactsWidget (row opens), and ArtifactAutoOpenMount (the artifact
+    // list). ChatTranscriptSurface renders its own provider for the
```
morgmart: 🤖 **P3 · Comment reverses provider ownership** (non-blocking)

The changed ChatView comment says ChatTranscriptSurface renders its own ArtifactPolicyProvider, but this same comparison removes that provider and establishes the composing session surface as the single owner. The comment now documents the opposite of the corrected architecture beside the load-bearing provider boundary.

**User effect:** A future maintainer can follow the comment and reintroduce nested providers, bringing back inconsistent artifact inventory and duplicate open-action lifecycles across one chat.

**Recommended fix:** Update the comment to state that ChatTranscriptSurface intentionally does not own artifact policy and that ChatView provides the single shared boundary for the transcript and its sibling consumers.

**Test:** No runtime test is needed for prose alone; keep the existing provider-boundary tests that enforce one owner across the composed chat surface.

morgmart (commented):
🤖 Automated code review

COMMENT: all four previously reported issues are fixed in the current comparison. One new non-blocking documentation finding remains: ChatView's changed ownership comment says ChatTranscriptSurface creates a provider even though this range removes it. Supplied visual evidence is present. GitHub evidence is structurally valid; several checks were still in progress at capture time, so required checks govern merge readiness.

badideasforsale:
🤖 P3 addressed in `2eba344` — the ChatView comment now documents the single-owner architecture (ChatView owns the policy boundary; ChatTranscriptSurface intentionally mounts no provider, enforced by the boundary tests). Audited for other stale ownership prose (`rg`): the test-file mock comment carried the same claim and is fixed in the same commit; no other hits.

morgmart (commented):
🤖 Automated code review

COMMENT: all five prior review findings are fixed in the current comparison. One new non-blocking lifecycle issue remains: the artifact-policy provider changes session identity without remounting, so provider-local debounce state can leak briefly across reconciled sessions. Supplied visual evidence is present. GitHub evidence is structurally valid; macOS Tauri checks and frontend checks were still in progress at capture time, so required checks govern merge readiness.

src/features/chat/ui/ChatView.tsx
```diff
@@ -793,11 +800,28 @@ export function ChatView({
...
+    // (the artifact list). ChatTranscriptSurface intentionally does NOT own
+    // artifact policy — a nested provider would derive a second artifact
+    // inventory and split the per-path open debounce (enforced by the
+    // provider-boundary tests). Consumers outside the provider get the
+    // inert default context and silently no-op.
+    // The identity, messages, and cwd must describe one session snapshot:
+    // timelineSessionId (the controller's effective session) rather than the
+    // raw requested sessionId, which can briefly disagree with
+    // controller.messages during session replacement or reconciliation.
+    <ArtifactPolicyProvider
```
morgmart, line 846: 🤖 **P2 · Provider state crosses session replacement** (non-blocking)

The moved ArtifactPolicyProvider receives timelineSessionId, but React reuses the same provider instance when that prop changes. Provider-local refs, including the per-path open debounce map, therefore survive when reconciliation replaces the effective session without unmounting ChatView. A path opened in the previous session remains temporarily marked as recently opened in the replacement session.

**User effect:** If someone opens the same file path immediately after a chat session is replaced or reconciled, the first open in the new session can appear to do nothing because it is suppressed by the previous session's debounce state.

**Recommended fix:** Key the ArtifactPolicyProvider by timelineSessionId, or explicitly reset all session-scoped provider state when the session identity changes, so each effective session starts with an independent artifact-action lifecycle.

**Test:** Add a reconciliation test that opens a path, changes the effective session identity without unmounting ChatView, and verifies that opening the same path in the replacement session reaches the OS boundary immediately.

badideasforsale:
🤖 Round-4 P2 addressed in `0775ff3` — session-scoped provider state (the per-path open debounce map, plus the artifact cache for lifecycle consistency) now resets when the provider's session identity changes, so a reconciled session starts with an independent artifact-action lifecycle. Chose the explicit reset over `key={sessionId}` remounting so no call site can forget the contract; session-scoped refs audited in the commit message. Added the reviewer-specified test: open a path, swap the session id without unmounting, and the first open of the same path in the new session reaches the OS boundary immediately (fails against the previous head).

Also rebased onto latest main (`183542e`, 15 commits) — clean, no conflicts; full chat+home suites pass (3,265 tests).

morgmart (approved):
🤖 Automated code review

NO PUBLICATION / RETRY: the fresh review found no publishable findings, but one lifecycle concern is the same underlying issue as an existing unresolved automation thread with no substantive human reply, so it is suppressed and approval fails closed under the duplicate policy. All earlier findings are otherwise fixed. Supplied visual evidence is present. GitHub evidence is structurally valid; macOS Tauri checks were still in progress at capture time, so required checks govern merge readiness.

PR #269, opened 2026-09-01
harden remote daemon recovery and log draining

damienrj:
**Category:** fix
**User Impact:** Remote SSH sessions recover reliably after interrupted cleanup, tolerate noisy startup logs, and let users remove broken ad-hoc hosts.
**Problem:** A crashed stale-lock reclaimer could block every future remote daemon operation, the chunked logger could backpressure Goose before it bound its port, and failed free-form hosts could become permanent settings rows. **Solution:** Reclaim lock generations atomically, drain startup logs through a native-speed bounded reader, and provide a terminal-state Forget operation that clears both backend and persisted UI state.

<details>

**src-tauri/src/commands/remote_backend.rs**
Exposes the inactive-host forget operation to the renderer.

**src-tauri/src/lib.rs**
Registers the new remote backend command.

**src-tauri/src/services/remote_backend/daemon.rs**
Adds deterministic kill-during-reclaim coverage and preserves the bounded-tail and 5 MiB pre-bind startup regressions.

**src-tauri/src/services/remote_backend/mod.rs**
Removes inactive host slots without allowing active connections to be orphaned, with focused registry tests.

**src-tauri/src/services/remote_backend/remote_daemon.sh**
Reclaims stale lock generations with one atomic rename and replaces the process-heavy chunk logger with a native-speed bounded FIFO reader.

**src/features/remoteHosts/stores/remoteHostStore.test.ts**
Covers complete cleanup of a broken host and preservation of local state when the backend rejects an active-host removal.

**src/features/remoteHosts/stores/remoteHostStore.ts**
Forgets backend state plus all persisted per-host metadata, including diagnostics, recent directories, and Goose overrides.

**src/features/remoteHosts/ui/RemoteHostsSettings.tsx**
Shows Forget for failed or disconnected ad-hoc hosts while leaving SSH-config hosts managed by SSH config.

**src/features/remoteHosts/ui/__tests__/RemoteHostsSettings.test.tsx**
Regresses the invalid free-form host case and verifies SSH-config hosts do not expose Forget.

**src/shared/api/remoteHosts.ts**
Adds the typed renderer wrapper for forgetting an inactive backend.

</details>

image

### DIFF
--- a/src-tauri/src/services/remote_backend/mod.rs
+++ b/src-tauri/src/services/remote_backend/mod.rs
@@ -118,6 +118,9 @@ struct HostSlot {
 
 struct SlotShared {
     state: RemoteBackendState,
+    /// Set before an inactive slot leaves the registry. Connect attempts that
+    /// already retained its Arc must observe this tombstone after admission.
+    forgotten: bool,
     /// Monotonic ownership token: each successful establish bumps it, and a
     /// supervisor only acts while its own generation is current. Explicit
     /// disconnects bump it to strand any racing supervisor.
@@ -141,6 +144,7 @@ impl RemoteBackendRegistry {
                 connect_lock: tokio::sync::Mutex::new(()),
                 shared: Mutex::new(SlotShared {
                     state: RemoteBackendState::Disconnected,
+                    forgotten: false,
                     generation: 0,
                     daemon: None,
                     local_port: None,
@@ -173,19 +177,34 @@ impl RemoteBackendRegistry {
     /// Remove an inactive host slot from the registry. The exact key is used
     /// deliberately so malformed inputs from a failed connect can still be
     /// forgotten instead of having to pass host parsing again.
-    pub fn forget(&self, key: &str) -> bool {
-        let mut slots = self.slots.lock().expect("remote backend registry poisoned");
-        let Some(slot) = slots.get(key) else {
+    pub async fn forget(&self, key: &str) -> bool {
+        let Some(slot) = self.existing_slot(key) else {
             return true;
         };
-        let forgettable = matches!(
-            slot.shared.lock().expect("slot poisoned").state,
-            RemoteBackendState::Disconnected | RemoteBackendState::Failed { .. }
-        );
-        if forgettable {
-            slots.remove(key);
+
+        // Serialize with every establish/reconnect owner. A connect may have
+        // retained this Arc before waiting on the lock, so removal also leaves
+        // a tombstone and advances its generation for that admitted waiter.
+        let _guard = slot.connect_lock.lock().await;
+        let mut slots = self.slots.lock().expect("remote backend registry poisoned");
+        let Some(registered) = slots.get(key) else {
+            return true;
+        };
+        if !Arc::ptr_eq(registered, &slot) {
+            return false;
         }
-        forgettable
+        let mut shared = slot.shared.lock().expect("slot poisoned");
+        if !matches!(
+            shared.state,
+            RemoteBackendState::Disconnected | RemoteBackendState::Failed { .. }
+        ) {
+            return false;
+        }
+        shared.forgotten = true;
+        shared.generation += 1;
+        drop(shared);
+        slots.remove(key);
+        true
     }
 
     /// Best-effort synchronous tunnel teardown for app exit. Daemons are left
@@ -368,6 +387,11 @@ pub async fn connect(
 
     {
         let mut shared = slot.shared.lock().expect("slot poisoned");
+        if shared.forgotten {
+            return Err(RemoteBackendError::internal(
+                "remote connection attempt was superseded",
+            ));
+        }
         if let Some(existing) = cached_connection(&shared, goose_path.as_deref()) {
             return Ok(existing);
         }
@@ -746,6 +770,7 @@ mod tests {
                 http_base_url: "http://127.0.0.1:5000".to_string(),
                 local_port: 5000,
             },
+            forgotten: false,
             generation: 1,
             daemon: Some(RemoteDaemonInfo {
                 pid: 10,
@@ -855,8 +880,8 @@ mod tests {
         assert_eq!(snapshot[0].host, "devbox");
     }
 
-    #[test]
-    fn forget_removes_inactive_slot_by_its_exact_key() {
+    #[tokio::test]
+    async fn forget_removes_inactive_slot_by_its_exact_key() {
         let registry = RemoteBackendRegistry::default();
         let spec = RemoteHostSpec::parse("devbox", &[]).unwrap();
         let slot = registry.slot(&spec);
@@ -864,18 +889,33 @@ mod tests {
             error: RemoteBackendError::internal("failed"),
         };
 
-        assert!(registry.forget("devbox"));
+        assert!(registry.forget("devbox").await);
         assert!(registry.snapshot().is_empty());
     }
 
-    #[test]
-    fn forget_preserves_active_slot() {
+    #[tokio::test]
+    async fn forget_preserves_active_slot() {
         let registry = RemoteBackendRegistry::default();
         let spec = RemoteHostSpec::parse("devbox", &[]).unwrap();
         let slot = registry.slot(&spec);
         slot.shared.lock().expect("slot poisoned").state = RemoteBackendState::Connecting;
 
-        assert!(!registry.forget("devbox"));
+        assert!(!registry.forget("devbox").await);
         assert_eq!(registry.snapshot().len(), 1);
+    }
+
+    #[tokio::test]
+    async fn pending_connect_observes_tombstone_after_forget() {
+        let registry = RemoteBackendRegistry::default();
+        let spec = RemoteHostSpec::parse("devbox", &[]).unwrap();
+        let admitted = registry.slot(&spec);
+
+        assert!(registry.forget("devbox").await);
+        let _guard = admitted.connect_lock.lock().await;
+        assert!(admitted.shared.lock().expect("slot poisoned").forgotten);
+
+        let replacement = registry.slot(&spec);
+        assert!(!Arc::ptr_eq(&admitted, &replacement));
+        assert!(!replacement.shared.lock().expect("slot poisoned").forgotten);
     }
 }
--- a/src-tauri/src/services/remote_backend/mod.rs
+++ b/src-tauri/src/services/remote_backend/mod.rs
@@ -84,6 +84,8 @@ pub enum RemoteBackendState {
 #[serde(rename_all = "camelCase")]
 pub struct RemoteBackendStatus {
     pub host: String,
+    pub incarnation: String,
+    pub generation: u64,
     #[serde(flatten)]
     pub state: RemoteBackendState,
 }
@@ -97,6 +99,9 @@ pub struct RemoteBackendConnection {
     pub local_port: u16,
     pub goose_version: String,
     pub daemon_reused: bool,
+    /// Unique identity for this registry slot. A forgotten host receives a
+    /// new incarnation even when its per-slot generation restarts.
+    pub incarnation: String,
     /// Slot generation that owns this tunnel. Callers use it to invalidate
     /// only the connection they established, never a newer replacement.
     pub generation: u64,
@@ -110,6 +115,7 @@ pub struct RemoteBackendRegistry {
 struct HostSlot {
     key: String,
     spec: RemoteHostSpec,
+    incarnation: String,
     /// Serializes establish attempts (user connects and supervisor
     /// reconnects) per host.
     connect_lock: tokio::sync::Mutex<()>,
@@ -141,6 +147,7 @@ impl RemoteBackendRegistry {
             Arc::new(HostSlot {
                 key: spec.key(),
                 spec: spec.clone(),
+                incarnation: uuid::Uuid::new_v4().to_string(),
                 connect_lock: tokio::sync::Mutex::new(()),
                 shared: Mutex::new(SlotShared {
                     state: RemoteBackendState::Disconnected,
@@ -167,9 +174,14 @@ impl RemoteBackendRegistry {
         let slots = self.slots.lock().expect("remote backend registry poisoned");
         slots
             .values()
-            .map(|slot| RemoteBackendStatus {
-                host: slot.key.clone(),
-                state: slot.shared.lock().expect("slot poisoned").state.clone(),
+            .map(|slot| {
+                let shared = slot.shared.lock().expect("slot poisoned");
+                RemoteBackendStatus {
+                    host: slot.key.clone(),
+                    incarnation: slot.incarnation.clone(),
+                    generation: shared.generation,
+                    state: shared.state.clone(),
+                }
             })
             .collect()
     }
@@ -182,9 +194,22 @@ impl RemoteBackendRegistry {
             return true;
         };
 
-        // Serialize with every establish/reconnect owner. A connect may have
-        // retained this Arc before waiting on the lock, so removal also leaves
-        // a tombstone and advances its generation for that admitted waiter.
+        // Invalidate queued connection admission before waiting on the lock.
+        // A current establish already holding the lock publishes Connecting,
+        // so only an inactive slot can reach this point; a waiter that retained
+        // the Arc must observe the tombstone when it eventually acquires.
+        {
+            let mut shared = slot.shared.lock().expect("slot poisoned");
+            if !matches!(
+                shared.state,
+                RemoteBackendState::Disconnected | RemoteBackendState::Failed { .. }
+            ) {
+                return false;
+            }
+            shared.forgotten = true;
+            shared.generation += 1;
+        }
+
         let _guard = slot.connect_lock.lock().await;
         let mut slots = self.slots.lock().expect("remote backend registry poisoned");
         let Some(registered) = slots.get(key) else {
@@ -193,16 +218,6 @@ impl RemoteBackendRegistry {
         if !Arc::ptr_eq(registered, &slot) {
             return false;
         }
-        let mut shared = slot.shared.lock().expect("slot poisoned");
-        if !matches!(
-            shared.state,
-            RemoteBackendState::Disconnected | RemoteBackendState::Failed { .. }
-        ) {
-            return false;
-        }
-        shared.forgotten = true;
-        shared.generation += 1;
-        drop(shared);
         slots.remove(key);
         true
     }
@@ -245,11 +260,12 @@ fn kill_tunnel_pid(pid: u32) {
 }
 
 fn set_state(app: &AppHandle, slot: &HostSlot, state: RemoteBackendState) {
-    {
+    let generation = {
         let mut shared = slot.shared.lock().expect("slot poisoned");
         shared.state = state.clone();
-    }
-    emit_status(app, &slot.key, &state);
+        shared.generation
+    };
+    emit_status(app, slot, generation, &state);
 }
 
 fn update_state_if_current(
@@ -277,13 +293,15 @@ fn set_state_if_current(
     if !update_state_if_current(&mut shared, generation, state.clone()) {
         return false;
     }
-    emit_status(app, &slot.key, &state);
+    emit_status(app, slot, generation, &state);
     true
 }
 
-fn emit_status(app: &AppHandle, host: &str, state: &RemoteBackendState) {
+fn emit_status(app: &AppHandle, slot: &HostSlot, generation: u64, state: &RemoteBackendState) {
     let payload = RemoteBackendStatus {
-        host: host.to_string(),
+        host: slot.key.clone(),
+        incarnation: slot.incarnation.clone(),
+        generation,
         state: state.clone(),
     };
     if let Err(error) = app.emit(REMOTE_BACKEND_STATUS_EVENT, &payload) {
@@ -318,7 +336,10 @@ fn extra_serve_args() -> Vec<String> {
     }
 }
 
-fn connection_from_shared(shared: &SlotShared) -> Option<RemoteBackendConnection> {
+fn connection_from_shared(
+    shared: &SlotShared,
+    incarnation: &str,
+) -> Option<RemoteBackendConnection> {
     let daemon = shared.daemon.as_ref()?;
     let local_port = shared.local_port?;
     if let RemoteBackendState::Ready {
@@ -334,6 +355,7 @@ fn connection_from_shared(shared: &SlotShared) -> Option<RemoteBackendConnection
             local_port,
             goose_version: daemon.goose_version.clone(),
             daemon_reused: daemon.reused,
+            incarnation: incarnation.to_string(),
             generation: shared.generation,
         })
     } else {
@@ -347,11 +369,24 @@ fn connection_from_shared(shared: &SlotShared) -> Option<RemoteBackendConnection
 fn cached_connection(
     shared: &SlotShared,
     requested_goose_path: Option<&str>,
+    incarnation: &str,
 ) -> Option<RemoteBackendConnection> {
     if shared.goose_path.as_deref() != requested_goose_path {
         return None;
     }
-    connection_from_shared(shared)
+    connection_from_shared(shared, incarnation)
+}
+
+async fn acquire_connect_admission(
+    slot: &HostSlot,
+) -> Result<tokio::sync::MutexGuard<'_, ()>, RemoteBackendError> {
+    let guard = slot.connect_lock.lock().await;
+    if slot.shared.lock().expect("slot poisoned").forgotten {
+        return Err(RemoteBackendError::internal(
+            "remote connection attempt was superseded",
+        ));
+    }
+    Ok(guard)
 }
 
 fn advance_generation_if_current(shared: &mut SlotShared, expected: u64) -> Option<u64> {
@@ -383,16 +418,19 @@ pub async fn connect(
     let goose_path = goose_path.map(daemon::normalize_goose_path).transpose()?;
     let slot = registry.slot(&spec);
 
-    let _guard = slot.connect_lock.lock().await;
+    let _guard = acquire_connect_admission(&slot).await?;
 
     {
         let mut shared = slot.shared.lock().expect("slot poisoned");
+        // Forget can tombstone an inactive slot after this connect acquires
+        // admission but before it publishes Connecting.
         if shared.forgotten {
             return Err(RemoteBackendError::internal(
                 "remote connection attempt was superseded",
             ));
         }
-        if let Some(existing) = cached_connection(&shared, goose_path.as_deref()) {
+        if let Some(existing) = cached_connection(&shared, goose_path.as_deref(), &slot.incarnation)
+        {
             return Ok(existing);
         }
         if shared.goose_path.as_deref() != goose_path.as_deref() {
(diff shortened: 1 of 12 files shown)


src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -166,31 +159,25 @@ terminate_uncommitted_daemon() {
...
-  fi
-  reclaim_held=1
-
   if lock_owner_is_current; then
-    release_reclaim_lock
     return 1
   fi
 
   claimed_lock="$STATE_DIR/.daemon.lock.reclaimed.$$.$lock_attempt"
   if mv "$LOCK_DIR" "$claimed_lock" 2>/dev/null; then
```
kalvinnchau: 🤖 [P1] The stale decision and this rename are not tied to the same lock generation. A reclaimer can observe stale generation G0 and pause; another process removes G0; a successor creates G1, publishes its live owner, waits 10 ms, verifies ownership, and enters the critical section; then the first reclaimer resumes and moves/deletes G1 here. That permits another caller to acquire G2 while the G1 owner is still mutating `daemon.record` or daemon lifecycle state.

I reproduced this deterministically at this commit by pausing a reclaimer after `lock_owner_is_current`, allowing a live successor to pass the acquisition check and enter the critical section, then resuming the reclaimer. The successor's `daemon.lock/owner` disappeared while it remained alive. The 10 ms delay narrows the race but cannot synchronize across scheduler stalls.

Please make reclamation generation-safe so the object renamed is provably the generation inspected, and add a regression that pauses the stale observer before this `mv` while a successor holds the lock.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -325,45 +326,45 @@ trim_log_if_needed() {
...
+  writer_cat_pid=""
+  trap 'kill -TERM "$writer_cat_pid" 2>/dev/null || true; wait "$writer_cat_pid" 2>/dev/null || true; rm -f "$writer_tmp"; exit 143' TERM
+  trap 'kill -TERM "$writer_cat_pid" 2>/dev/null || true; wait "$writer_cat_pid" 2>/dev/null || true; rm -f "$writer_tmp"; exit 129' HUP
+  trap 'kill -TERM "$writer_cat_pid" 2>/dev/null || true; wait "$writer_cat_pid" 2>/dev/null || true; rm -f "$writer_tmp"; exit 130' INT
+
   while :; do
-    rm -f "$writer_chunk"
-    dd bs="$LOG_WRITE_CHUNK_BYTES" count=1 of="$writer_chunk" 2>/dev/null || true
-    writer_bytes="$(wc -c <"$writer_chunk" 2>/dev/null | tr -d ' ')"
+    (ulimit -S -f "$LOG_MAX_KIB" && exec cat <&3 >>"$writer_log") 2>/dev/null &
```
kalvinnchau: 🤖 [P2] Restarting `cat` after `RLIMIT_FSIZE` does not preserve all FIFO bytes. `cat` can read records beyond the remaining file allowance into userspace; when its append crosses the limit and the process dies, those already-consumed bytes are no longer in the FIFO for the next `cat`.

I verified this at this commit with 80,000 sequential 64-byte records and a one-byte preseeded log. The retained log contained IDs `32768…65535`, then `65537…79999`: record `65536` was lost exactly at rotation. This can discard the startup diagnostic that matters at the boundary even though the file remains bounded.

Please use a rotation writer that carries overflow/read-ahead into the next file, and make the regression assert that retained sequence records are contiguous rather than only checking the size and final marker.

damienrj: 🤖 Addressed in e7c9f0ad. The logger no longer kills a reader at a file-size boundary. It drains batches into a chunk file, then carries every consumed byte through bounded tail rotation before reading again, so userspace read-ahead cannot disappear. The regression now uses the reported one-byte preseed plus 80,000 fixed-width 64-byte sequence records and asserts every retained adjacent ID is contiguous through 79,999, while preserving the strict 4 MiB cap and pre-bind startup requirement.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Addressed in e7c9f0ad. The shared reusable lock pathname is gone: each invocation now publishes a UUID-scoped Lamport bakery ticket, and stale cleanup deletes only that exact never-reused ticket generation. Legacy `daemon.lock` migration happens before ticket publication and cannot target a live ticket. The new regression pauses a legacy observer before its move, lets a successor hold a ticket, resumes the observer, and verifies the successor owner remains intact; a separate kill-during-migration regression verifies future operations still recover.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -166,92 +161,151 @@ terminate_uncommitted_daemon() {
...
+    exit 46
+    ;;
+  esac
+  our_identity="$(process_identity "$$")" || {
+    emit "ERR state-dir"
+    exit 46
+  }
+  our_lock_owner="$$ $(b64 "$our_identity")"
+
+  legacy_attempt=0
```
kalvinnchau: 🤖 [P1] The compatibility check only waits out a legacy `daemon.lock` that already exists. Once this invocation publishes its ticket, an older Berd client can create `daemon.lock`; that client does not observe `daemon.locks/ticket.*`, and this client never checks the legacy path again. Both can therefore enter the `daemon.record` mutation section during a normal mixed-version rollout.

I reproduced this deterministically at `f87ea471`: I paused this script after ticket acquisition, then ran the previous-head (`6f72c8cf`) script and paused it after legacy-lock acquisition. Both critical-section markers existed simultaneously, with one live ticket and a live `daemon.lock`. Mutual exclusion had failed.

Please use a migration protocol observed by both versions—for example, retain the legacy lock as the compatibility gate while ticket clients exist—and add cross-version tests in both acquisition orders.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -325,45 +379,58 @@ trim_log_if_needed() {
...
+  writer_tmp="$STATE_DIR/.goose-serve.log.rotate.$$"
+  trap 'rm -f "$writer_chunk" "$writer_tmp"; exit 143' TERM
+  trap 'rm -f "$writer_chunk" "$writer_tmp"; exit 129' HUP
+  trap 'rm -f "$writer_chunk" "$writer_tmp"; exit 130' INT
+
   while :; do
     rm -f "$writer_chunk"
-    dd bs="$LOG_WRITE_CHUNK_BYTES" count=1 of="$writer_chunk" 2>/dev/null || true
-    writer_bytes="$(wc -c <"$writer_chunk" 2>/dev/null | tr -d ' ')"
+    dd bs="$LOG_READ_BLOCK_BYTES" count="$LOG_READ_BLOCK_COUNT" of="$writer_chunk" 2>/dev/null || true
```
kalvinnchau: 🤖 [P2] This `dd` does not publish a chunk until it completes 64 input reads or reaches EOF. A long-lived Goose process that writes a short startup diagnostic and then keeps stderr open leaves `dd` blocked, so `goose-serve.log` remains absent or stale for the daemon lifetime. The bytes are neither visible in the log nor yet flushed to the chunk file.

I reproduced this at `f87ea471` with a Goose shim that wrote one `IMPORTANT-STARTUP-DIAGNOSTIC` line, successfully bound its port, and slept with stderr open. `ensure` returned `READY`, but after one second `goose-serve.log` did not exist and the chunk file remained zero bytes. The diagnostic was trapped upstream of the advertised log.

Please stream bounded data into the visible log without waiting for a read-count/EOF boundary, and add a regression asserting one startup line becomes visible while its producer remains alive.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -170,6 +170,24 @@ impl RemoteBackendRegistry {
...
 
+    /// Remove an inactive host slot from the registry. The exact key is used
+    /// deliberately so malformed inputs from a failed connect can still be
+    /// forgotten instead of having to pass host parsing again.
+    pub fn forget(&self, key: &str) -> bool {
+        let mut slots = self.slots.lock().expect("remote backend registry poisoned");
+        let Some(slot) = slots.get(key) else {
+            return true;
+        };
+        let forgettable = matches!(
```
kalvinnchau: 🤖 [P1] `Disconnected`/`Failed` does not prove that no operation still owns this slot. `connect()` obtains an `Arc<HostSlot>` before awaiting `connect_lock`; while it waits, `forget()` can see the current state as disconnected, remove the registry entry, and return success. The pending connect then resumes on its orphaned `Arc`, can establish a daemon/tunnel, and publish status, while subsequent registry lookups create a different slot that cannot disconnect or supervise that transport.

I verified the ownership fracture at `f87ea471` with a regression that retains the pre-forget slot as a pending-connect owner, calls `forget`, advances that retained slot to `Connecting`, and then obtains the same host again: the registry snapshot omitted the active old slot and returned a distinct `Arc`. The full remote-backend package passed 76/76 with that reproducer included; the existing tests only cover state at the instant of removal.

Please serialize forget with connection admission or tombstone/invalidate the slot so every outstanding initializer must fail its generation check before transport setup. Add a concurrent forget/connect regression.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -325,45 +379,58 @@ trim_log_if_needed() {
...
+    if [ "$chunk_bytes" -ge "$LOG_MAX_BYTES" ]; then
+      if ! tail -c "$LOG_MAX_BYTES" "$writer_chunk" >"$writer_tmp" 2>/dev/null; then
+        rm -f "$writer_tmp"
+        break
+      fi
+    elif [ $((writer_bytes + chunk_bytes)) -le "$LOG_MAX_BYTES" ]; then
+      cat "$writer_chunk" >>"$writer_log"
+      continue
+    else
+      carry_bytes=$((LOG_MAX_BYTES - chunk_bytes))
```
kalvinnchau: 🤖 [P3] After the log first reaches the cap, every later nonempty chunk rewrites the entire 4 MiB tail. This branch retains `LOG_MAX_BYTES - chunk_bytes` and then appends the chunk, leaving `writer_log` exactly at `LOG_MAX_BYTES`; the next chunk necessarily enters the same rotation branch. Because pipe reads can produce chunks well below 4 MiB, a continuously chatty daemon repeatedly reads and replaces the full log for comparatively little new output.

The previous implementation rotated down to `LOG_RETAIN_BYTES`, preserving 2 MiB of headroom. Retaining that bounded tail here would preserve record continuity while avoiding full-file rewrite amplification on every subsequent batch. Please add a sustained-output check that constrains rotation frequency or resulting write amplification.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Fixed in 742fbb54. Ticket clients now acquire and retain daemon.lock after ticket ordering, so legacy and current clients share the same compatibility gate for the full mutation. Added regressions for both acquisition orders: a live legacy holder blocks a ticket client, and a live ticket holder blocks a legacy client.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Fixed in 742fbb54. The FIFO now streams through tee directly into goose-serve.log while head acts only as the segment byte counter, so partial startup output is visible immediately without waiting for EOF or a full batch. Added a regression that asserts IMPORTANT-STARTUP-DIAGNOSTIC becomes visible while the producer remains alive.

src-tauri/src/services/remote_backend/mod.rs
damienrj: 🤖 Fixed in 742fbb54. Forget now serializes on connect_lock, tombstones the slot, and advances its generation before registry removal. Any connect that retained the old Arc must observe the tombstone after admission and fail before transport setup. Added a pending-connect/tombstone regression.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Fixed in 742fbb54. Logging now rotates from a 2 MiB retained tail only after roughly 2 MiB of additional streamed output, with 64 KiB reserved for the final in-flight pipe read. The sustained 5 MiB regression now constrains full-tail rewrites to at most two while preserving a contiguous bounded tail.

src/features/remoteHosts/stores/remoteHostStore.ts
```diff
@@ -290,14 +291,36 @@ export const useRemoteHostStore = create<RemoteHostStore>((set, get) => ({
...
       const manualHosts = state.manualHosts.filter(
         (candidate) => candidate !== host,
       );
+      const statusByHost = { ...state.statusByHost };
+      const doctorByHost = { ...state.doctorByHost };
+      const doctorPendingByHost = { ...state.doctorPendingByHost };
+      const doctorErrorByHost = { ...state.doctorErrorByHost };
+      const recentDirsByHost = { ...state.recentDirsByHost };
+      const goosePathByHost = { ...state.goosePathByHost };
+      delete statusByHost[host];
```
morgmart, line 520: 🤖 **P1 · Stale updates restore forgotten hosts** (blocking)

forgetHost removes the host only after the backend accepts Forget, but it does not invalidate frontend work that was admitted earlier. A listRemoteBackends request already in flight can resolve afterward and write the old entry back through syncBackendSnapshot; likewise, a queued backend status callback can run after the deletion and applyStatusEvent recreates the key unconditionally. That leaves the frontend lifecycle behind the backend registry lifecycle introduced by this change.

**User effect:** After choosing Forget, people can see the removed host return as a disconnected or failed row. They may then act on a host Berd has already forgotten, and the recovery action appears unreliable.

**Recommended fix:** Track a host lifecycle/incarnation or Forget tombstone in the store. Advance it when Forget succeeds, and admit snapshot entries and status events only when they still belong to the current incarnation; clear the tombstone only when a new user-initiated connection intentionally starts a new incarnation.

**Test:** Add discriminating store tests where a snapshot request starts before Forget and resolves afterward, and where a pre-Forget status event is delivered afterward; in both cases the host must stay absent. Also verify an intentional later reconnect can create a new incarnation.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
```diff
@@ -282,12 +280,14 @@ function RemoteHostRow({
...
           </Button>
-          {isManualHost && !isConnected ? (
+          {!isConfigHost && (state === "failed" || state === "disconnected") ? (
             <Button
               type="button"
               variant="ghost"
               size="sm"
-              onClick={() => removeManualHost(host)}
+              onClick={() => {
+                void forgetHost(host).catch(() => {});
```
morgmart, line 299: 🤖 **P2 · Forget failures have no feedback** (non-blocking)

The click handler catches and discards every forgetHost rejection. The backend is designed to reject Forget when the host is active or a lifecycle operation wins the race, so this is an expected outcome rather than an unreachable error.

**User effect:** When Forget cannot complete, the row simply stays in place and the button appears broken; people receive no explanation or next step.

**Recommended fix:** Represent a host-scoped Forget pending/error state and show the established inline or notification feedback while preserving the row. Explain when the host must be disconnected or when the action can be retried, and disable repeated Forget clicks while the request is pending.

**Test:** Add a component test that rejects forgetHost and verifies visible localized feedback and retained row state, plus a pending-state assertion that prevents duplicate submissions.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the new Forget lifecycle can be undone by already-admitted frontend status work, allowing a forgotten host to reappear, and expected Forget refusals have no visible feedback. Supplied GitHub evidence is structurally available and all listed checks completed successfully, but required checks independently govern merge readiness.

src/features/remoteHosts/stores/remoteHostStore.ts
damienrj: 🤖 Fixed in 639e59e7 after rebasing onto main. The store now advances a per-host lifecycle on successful Forget and explicit reconnect, keeps a Forget tombstone until user-initiated Connect, ignores status events for tombstoned hosts, and rejects snapshot entries whose admission lifecycle is stale. Added regressions for a pre-Forget snapshot resolving after Forget, the stronger snapshot-after-reconnect case, late status events, and intentional reconnect admission.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
damienrj: 🤖 Fixed in 639e59e7 after rebasing onto main. Forget now exposes host-scoped pending/error state, deduplicates submissions, disables the button with a localized “Forgetting…” label, and keeps the row with localized retry/disconnect guidance when the backend refuses the operation. Added store and component coverage for rejection, retained row state, visible feedback, and duplicate prevention.

src/features/remoteHosts/stores/remoteHostStore.ts
```diff
@@ -249,7 +387,7 @@ export const useRemoteHostStore = create<RemoteHostStore>((set, get) => ({
     set((state) => ({
       statusByHost: {
         ...state.statusByHost,
-        [host]: { state: "disconnected" },
+        [host]: { ...state.statusByHost[host], state: "disconnected" },
```
morgmart: 🤖 **P1 · Stale mutations replace current state** (blocking)

The changed store introduces lifecycle, incarnation, and generation checks for snapshots, events, connect, and Forget, but disconnect and shutdown cross an asynchronous boundary and then unconditionally publish disconnected state. disconnect also omits the already-supported expectedGeneration argument, allowing the backend command to act on whichever generation currently occupies the host slot. If either operation overlaps a replacement connection, its late completion can disconnect or overwrite the replacement lifecycle.

**User effect:** A newly reconnected host can unexpectedly become disconnected or appear disconnected because an older Disconnect or Stop operation finishes late.

**Recommended fix:** Use one lifecycle-aware mutation pattern for Disconnect and Stop: capture the current local lifecycle, incarnation, and generation at admission; pass the supported generation to Disconnect; and commit completion only if the same lifecycle still owns the row. Ensure a backend identity mismatch is distinguishable from successful completion.

**Test:** Add adversarial tests that hold Disconnect and Stop responses, establish a replacement connection with a new generation and incarnation, then release the old responses. Verify the old Disconnect sends its expected generation and neither stale completion can affect or overwrite the replacement.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
```diff
@@ -282,20 +286,30 @@ function RemoteHostRow({
...
             {t("remoteHosts.actions.check")}
           </Button>
-          {isManualHost && !isConnected ? (
+          {!isConfigHost && (state === "failed" || state === "disconnected") ? (
             <Button
               type="button"
               variant="ghost"
               size="sm"
-              onClick={() => removeManualHost(host)}
+              disabled={forgetPending}
```
morgmart: 🤖 **P1 · Forget hides destructive scope** (blocking)

The new Forget action is presented as lightweight list cleanup, but successful Forget also permanently removes the host's saved Goose binary override and recent directories. The changed interface provides no confirmation or explanatory copy before clearing those preferences, while the separate destructive Stop backend styling makes Forget appear lower consequence.

**User effect:** People trying to remove a stale host row can unexpectedly lose remembered directories and custom Goose configuration they expected to reuse later.

**Recommended fix:** Either preserve host preferences and limit Forget to removing the remembered host, or use the existing confirmation-dialog pattern and clearly explain that the host, saved binary path, and recent directories will be removed.

**Test:** Add a component test proving the destructive scope is disclosed and explicitly confirmed before forgetHost runs, or a store test proving preferences remain when Forget is intentionally limited to list removal.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: asynchronous disconnect and shutdown completions can still overwrite or act on a replacement host lifecycle, and the graphical Remote SSH hosts settings change has no screenshots or short recording. The supplied GitHub evidence is structurally available and all listed checks completed successfully; required checks independently govern merge readiness.

Deterministic publication result: 2 blocking and 0 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 1 blocking screenshot-evidence requirement(s) in this review body.

src/features/remoteHosts/stores/remoteHostStore.ts
damienrj: 🤖 Fixed in 1477161d. Disconnect and Stop now capture the admitted frontend lifecycle plus backend incarnation/generation, pass the observed generation through IPC, and only publish Disconnected when that same lifecycle still owns the row. Rust rejects generation mismatches with a typed daemon-changed error, shutdown checks after acquiring the connection lock, and disconnect state publication is generation-gated. Added adversarial store tests that replace the lifecycle while each operation is pending and verify the old completion cannot overwrite it.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
damienrj: 🤖 Fixed in 1477161d by limiting Forget to host/list lifecycle cleanup. It no longer deletes the saved Goose binary override or recent remote directories, so removing a stale row is non-destructive and those preferences are available if the host is added again. The store regression now explicitly verifies both preferences remain in memory and persisted storage.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -655,7 +738,12 @@ pub async fn shutdown(
     // until shutdown releases this lock, so ensure_daemon cannot recreate the
     // daemon after shutdown_daemon stops it.
     let _guard = slot.connect_lock.lock().await;
-    disconnect(app, registry, &spec.key());
+    if !disconnect_generation(app, registry, &spec.key(), expected_generation) {
```
morgmart, line 757: 🤖 **P1 · Failed shutdown leaves false ready** (blocking)

shutdown disconnects the local tunnel and advances its generation before attempting remote daemon shutdown. If shutdown_daemon then fails, such as when the daemon identity changed, the command rejects after the tunnel was already removed. The frontend applies disconnected state only on success, and its generation checks can reject the emitted newer disconnected status, leaving the old ready state visible.

**User effect:** Berd can report that stopping the backend failed while still showing the host as connected, even though its tunnel is gone and subsequent remote work cannot use it.

**Recommended fix:** Make shutdown ordering transactional while holding the connection lock: validate and stop the expected daemon before committing local tunnel disconnection, or return authoritative post-failure status that the frontend must reconcile without rejecting its newer generation.

**Test:** Add a backend/store integration test where a ready host's daemon changes before shutdown, shutdown returns DaemonChanged after the local path is exercised, and the UI cannot remain ready after the tunnel is gone.

src/features/remoteHosts/stores/remoteHostStore.ts
```diff
@@ -290,14 +454,83 @@ export const useRemoteHostStore = create<RemoteHostStore>((set, get) => ({
...
+        },
+      }));
+      throw error;
+    }
     set((state) => {
-      if (!state.manualHosts.includes(host)) return state;
       const manualHosts = state.manualHosts.filter(
         (candidate) => candidate !== host,
       );
+      const statusByHost = { ...state.statusByHost };
```
morgmart, line 499: 🤖 **P1 · Late Forget erases new connect** (blocking)

forgetHost does not capture the host lifecycle when the request begins. After awaiting the backend, its success reducer unconditionally deletes status and connect-pending state and increments the current lifecycle. Because Connect remains available while Forget is pending, a newer connection can start before the older Forget response returns and then be erased by that late completion.

**User effect:** Someone who reconnects while Forget is finishing can see the new connection disappear, leaving Settings and the actual backend connection out of agreement.

**Recommended fix:** Capture the lifecycle at Forget admission and apply success or failure only if that lifecycle still owns the host, or make forgetting an explicit row-level fence that rejects or disables new connection attempts until it settles.

**Test:** Add an out-of-order test for Forget starts, a newer Connect starts, then Forget completes; verify the older result cannot delete or supersede the newer connection lifecycle.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
morgmart: 🤖 **P2 · Pending label shifts action row** (non-blocking)

The button manually replaces “Forget” with the wider “Forgetting…” label inside a wrapping action group. At constrained widths this can move the control or adjacent actions onto another line while the operation is pending, instead of using the shared Button async-feedback behavior that preserves geometry.

**User effect:** The host row can jump or rewrap immediately after someone chooses Forget, making the waiting state feel unstable and moving nearby controls.

**Recommended fix:** Use the shared Button feedbackState, loadingLabel, loadingVisual, and preserveWidth API for the pending state rather than swapping the children directly.

**Test:** Add a component assertion that the pending Forget action uses the shared width-preserving loading state; retain responsive visual verification through the requested screenshot or recording evidence.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: shutdown failure can leave a killed tunnel displayed as ready, and a late Forget completion can erase a newer connection lifecycle. The graphical Remote SSH hosts settings change also lacks screenshots or a short recording. Supplied check evidence is structurally valid; Transcript virtualization is in progress, while the other listed checks succeeded. Required checks independently govern merge readiness.

Deterministic publication result: 2 blocking and 1 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 1 blocking screenshot-evidence requirement(s) in this review body.

src-tauri/src/services/remote_backend/mod.rs
damienrj: 🤖 Fixed in 7a5c5bd2. Shutdown now validates the admitted local generation while holding the connection lock, attempts the generation-checked remote daemon stop first, and only tears down the tunnel after that succeeds. A daemon identity/generation rejection therefore leaves the still-authoritative ready tunnel intact. Added Rust coverage proving a generation mismatch preserves Ready, generation, and tunnel PID, plus store coverage proving a rejected shutdown cannot falsely publish Disconnected.

src/features/remoteHosts/ui/RemoteHostsSettings.tsx
damienrj: 🤖 Fixed in 7a5c5bd2. The Forget action now uses Button feedbackState, loadingLabel, loadingVisual="text", and preserveWidth instead of swapping children. Component coverage asserts the loading feedback metadata, busy/disabled state, and retained hidden idle layers that preserve geometry.

src/features/remoteHosts/stores/remoteHostStore.ts
damienrj: 🤖 Fixed in 7a5c5bd2. Forget now captures the frontend lifecycle at admission and gates both success and failure reducers on that lifecycle. A newer explicit Connect can take ownership while Forget is pending; the late Forget result clears only its pending marker and cannot erase the replacement status, persistence, or lifecycle. Added the requested out-of-order regression.

morgmart (approved):
🤖 Automated code review

APPROVE: no new publishable findings remain in the exact PR comparison. The previously reported lifecycle, shutdown, Forget-scope, and pending-button issues are fixed or protected by resolved threads with substantive human replies. Visual evidence is present for the changed Remote SSH hosts settings surface. All supplied checks completed successfully; required checks independently govern merge readiness.

PR #274, opened 2026-09-02
add SSH environments from the composer

damienrj:
**Category:** improvement
**User Impact:** Users can add and select a new SSH environment directly from the chat composer.

**Problem:** The environment picker only lists SSH hosts Berd already knows about, so adding a new host requires leaving the chat flow and finding the remote-host settings. **Solution:** Add a separated “Add SSH environment” action that opens an accessible connection dialog, accepts an SSH config alias or `user@host`, selects the environment after a successful connection, and keeps loading and error feedback in context. This PR is stacked on #269 and should merge after it.

<details>

**src/features/chat/ui/AddRemoteHostDialog.tsx**
Adds the focused connection dialog, including validation, pending-state protection, backend error feedback, and selection after a successful connection.

**src/features/chat/ui/RemoteHostSelector.tsx**
Adds the separated “Add SSH environment” menu action and preserves focus while transitioning from the environment menu into the dialog.

**src/features/chat/ui/__tests__/RemoteHostSelector.test.tsx**
Covers the empty-host menu, successful add-and-select flow, and retained dialog feedback when connection fails.

Adds localized copy and accessible labels for the new environment action and dialog.

</details>

### DIFF
--- a/src/features/chat/ui/AddRemoteHostDialog.tsx
+++ b/src/features/chat/ui/AddRemoteHostDialog.tsx
@@ -64,10 +64,14 @@ export function AddRemoteHostDialog({
     setPending(true);
     const attempt = ++attemptRef.current;
     try {
-      const accepted = await useRemoteHostStore
+      const outcome = await useRemoteHostStore
         .getState()
         .ensureHostConnected(host);
-      if (attemptRef.current !== attempt || !accepted) return;
+      if (attemptRef.current !== attempt) return;
+      if (outcome === "superseded") {
+        setError(t("toolbar.remoteHost.add.superseded"));
+        return;
+      }
       onConnected(host);
       handleOpenChange(false);
     } catch (connectionError) {
--- a/src/features/chat/ui/AddRemoteHostDialog.tsx
+++ b/src/features/chat/ui/AddRemoteHostDialog.tsx
@@ -1,0 +1,132 @@
+import { useState, type FormEvent } from "react";
+import { useTranslation } from "react-i18next";
+import { useRemoteHostStore } from "@/features/remoteHosts/stores/remoteHostStore";
+import { isRemoteBackendError } from "@/shared/api/remoteHosts";
+import { Button } from "@/shared/ui/button";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/shared/ui/dialog";
+import { Input } from "@/shared/ui/input";
+import { Label } from "@/shared/ui/label";
+
+interface AddRemoteHostDialogProps {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  onConnected: (host: string) => void;
+}
+
+export function AddRemoteHostDialog({
+  open,
+  onOpenChange,
+  onConnected,
+}: AddRemoteHostDialogProps) {
+  const { t } = useTranslation("chat");
+  const [hostDraft, setHostDraft] = useState("");
+  const [error, setError] = useState<string | null>(null);
+  const [pending, setPending] = useState(false);
+
+  const handleOpenChange = (nextOpen: boolean) => {
+    if (!nextOpen && pending) return;
+    onOpenChange(nextOpen);
+    if (!nextOpen) {
+      setHostDraft("");
+      setError(null);
+    }
+  };
+
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (pending) return;
+
+    const host = hostDraft.trim();
+    if (!host) {
+      setError(t("toolbar.remoteHost.add.emptyHost"));
+      return;
+    }
+
+    setError(null);
+    setPending(true);
+    try {
+      await useRemoteHostStore.getState().ensureHostConnected(host);
+      onConnected(host);
+      onOpenChange(false);
+      setHostDraft("");
+    } catch (connectionError) {
+      setError(
+        isRemoteBackendError(connectionError)
+          ? connectionError.message
+          : String(connectionError),
+      );
+    } finally {
+      setPending(false);
+    }
+  };
+
+  return (
+    <Dialog open={open} onOpenChange={handleOpenChange}>
+      <DialogContent
+        size="md"
+        closeLabel={t("toolbar.remoteHost.add.close")}
+        showCloseButton={!pending}
+      >
+        <form className="contents" onSubmit={handleSubmit}>
+          <DialogHeader>
+            <DialogTitle>{t("toolbar.remoteHost.add.title")}</DialogTitle>
+            <DialogDescription>
+              {t("toolbar.remoteHost.add.description")}
+            </DialogDescription>
+          </DialogHeader>
+          <div className="space-y-2">
+            <Label htmlFor="add-ssh-environment-host">
+              {t("toolbar.remoteHost.add.hostLabel")}
+            </Label>
+            <Input
+              id="add-ssh-environment-host"
+              value={hostDraft}
+              placeholder={t("toolbar.remoteHost.add.hostPlaceholder")}
+              disabled={pending}
+              aria-invalid={error ? true : undefined}
+              aria-describedby={error ? "add-ssh-environment-error" : undefined}
+              onChange={(event) => {
+                setHostDraft(event.target.value);
+                if (error) setError(null);
+              }}
+            />
+            {error ? (
+              <p
+                id="add-ssh-environment-error"
+                className="text-sm text-destructive"
+                role="alert"
+              >
+                {error}
+              </p>
+            ) : null}
+          </div>
+          <DialogFooter>
+            <Button
+              type="button"
+              variant="outline"
+              disabled={pending}
+              onClick={() => handleOpenChange(false)}
+            >
+              {t("toolbar.remoteHost.add.cancel")}
+            </Button>
+            <Button
+              type="submit"
+              feedbackState={pending ? "loading" : "idle"}
+              loadingLabel={t("toolbar.remoteHost.status.connecting")}
+              preserveWidth
+            >
+              {t("toolbar.remoteHost.add.connect")}
+            </Button>
+          </DialogFooter>
+        </form>
+      </DialogContent>
+    </Dialog>
+  );
+}
--- a/src/features/chat/ui/AddRemoteHostDialog.tsx
+++ b/src/features/chat/ui/AddRemoteHostDialog.tsx
@@ -1,4 +1,4 @@
-import { useState, type FormEvent } from "react";
+import { useEffect, useRef, useState, type FormEvent } from "react";
 import { useTranslation } from "react-i18next";
 import { useRemoteHostStore } from "@/features/remoteHosts/stores/remoteHostStore";
 import { isRemoteBackendError } from "@/shared/api/remoteHosts";
@@ -29,13 +29,24 @@ export function AddRemoteHostDialog({
   const [hostDraft, setHostDraft] = useState("");
   const [error, setError] = useState<string | null>(null);
   const [pending, setPending] = useState(false);
+  const attemptRef = useRef(0);
+
+  useEffect(
+    () => () => {
+      attemptRef.current += 1;
+    },
+    [],
+  );
 
   const handleOpenChange = (nextOpen: boolean) => {
-    if (!nextOpen && pending) return;
     onOpenChange(nextOpen);
     if (!nextOpen) {
+      // The backend connection may still finish, but closing the dialog must
+      // keep that late result from changing the composer's selected host.
+      attemptRef.current += 1;
       setHostDraft("");
       setError(null);
+      setPending(false);
     }
   };
 
@@ -51,29 +62,29 @@ export function AddRemoteHostDialog({
 
     setError(null);
     setPending(true);
+    const attempt = ++attemptRef.current;
     try {
       await useRemoteHostStore.getState().ensureHostConnected(host);
+      if (attemptRef.current !== attempt) return;
       onConnected(host);
-      onOpenChange(false);
-      setHostDraft("");
+      handleOpenChange(false);
     } catch (connectionError) {
+      if (attemptRef.current !== attempt) return;
       setError(
         isRemoteBackendError(connectionError)
           ? connectionError.message
           : String(connectionError),
       );
     } finally {
-      setPending(false);
+      if (attemptRef.current === attempt) {
+        setPending(false);
+      }
     }
   };
 
   return (
     <Dialog open={open} onOpenChange={handleOpenChange}>
-      <DialogContent
-        size="md"
-        closeLabel={t("toolbar.remoteHost.add.close")}
-        showCloseButton={!pending}
-      >
+      <DialogContent size="md" closeLabel={t("toolbar.remoteHost.add.close")}>
         <form className="contents" onSubmit={handleSubmit}>
           <DialogHeader>
             <DialogTitle>{t("toolbar.remoteHost.add.title")}</DialogTitle>
@@ -111,7 +122,6 @@ export function AddRemoteHostDialog({
             <Button
               type="button"
               variant="outline"
-              disabled={pending}
               onClick={() => handleOpenChange(false)}
             >
               {t("toolbar.remoteHost.add.cancel")}
--- a/src/features/chat/ui/AddRemoteHostDialog.tsx
+++ b/src/features/chat/ui/AddRemoteHostDialog.tsx
@@ -64,8 +64,10 @@ export function AddRemoteHostDialog({
     setPending(true);
     const attempt = ++attemptRef.current;
     try {
-      await useRemoteHostStore.getState().ensureHostConnected(host);
-      if (attemptRef.current !== attempt) return;
+      const accepted = await useRemoteHostStore
+        .getState()
+        .ensureHostConnected(host);
+      if (attemptRef.current !== attempt || !accepted) return;
       onConnected(host);
       handleOpenChange(false);
     } catch (connectionError) {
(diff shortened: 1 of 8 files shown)


src/features/remoteHosts/stores/remoteHostStore.ts
```diff
@@ -187,31 +208,54 @@ export const useRemoteHostStore = create<RemoteHostStore>((set, get) => ({
...
-      statusByHost: {
-        ...state.statusByHost,
-        [payload.host]: {
-          state: payload.state,
-          ...(payload.attempt !== undefined
-            ? { attempt: payload.attempt }
-            : {}),
-          ...(payload.error ? { error: payload.error } : {}),
+    set((state) => {
+      if (state.forgottenHosts[payload.host]) return state;
```
morgmart: 🤖 **P1 · Stale lifecycle updates win** (blocking)

A new connection clears the forget tombstone and increments lifecycleByHost, but applyStatusEvent accepts every later event while the tombstone is clear, and ensureHostConnected applies success or failure without checking that its captured lifecycle is still current. A delayed event or completion from an older connection can therefore overwrite a replacement lifecycle or restore a host after it was forgotten.

**User effect:** A newly reconnected SSH environment can suddenly appear failed or disconnected, and a host the user forgot can reappear after an older connection finishes.

**Recommended fix:** Carry the backend connection generation or instance identity through status events and connection results, then apply renderer updates only when they match the host's current lifecycle and the host is not tombstoned.

**Test:** Add discriminating tests that complete two connection lifecycles out of order, deliver a stale status event after reconnect, and complete an admitted connection after Forget; assert that only the newest live lifecycle remains visible and persisted.

src-tauri/src/services/remote_backend/remote_daemon.sh
```diff
@@ -166,92 +167,179 @@ terminate_uncommitted_daemon() {
...
-  return 0
+    if lock_owner_is_current "$LEGACY_LOCK_OWNER"; then
+      stale_observations=0
+    else
+      stale_observations=$((stale_observations + 1))
+      # A mkdir winner publishes its owner immediately. Repeated observations
+      # distinguish that window from a process that died before publication.
+      if [ "$stale_observations" -ge 20 ]; then
+        legacy_claim="$STATE_DIR/.daemon.lock.legacy.$NONCE.$$.$compat_attempt"
+        if mv "$LEGACY_LOCK_DIR" "$legacy_claim" 2>/dev/null; then
```
morgmart: 🤖 **P1 · Legacy lock reclaims successor** (blocking)

After repeated stale-owner observations, the compatibility path renames the shared daemon.lock pathname without proving it is still the lock generation that was observed. A live process can replace or recreate that directory between the last observation and mv, causing this process to move and delete the successor's live lock.

**User effect:** Two remote-backend operations can run as if each owns the lock, producing unreliable connection state or corrupting the daemon's shared records.

**Recommended fix:** Make legacy reclamation generation-specific, or perform an atomic claim with an identity check that cannot cross pathname reuse; never remove the shared path based only on earlier observations.

**Test:** Add a race test that replaces the observed stale legacy lock with a live successor immediately before reclamation and proves the successor lock remains intact and exclusive.

src/features/chat/ui/AddRemoteHostDialog.tsx
```diff
@@ -0,0 +1,132 @@
...
+      setHostDraft("");
+      setError(null);
+    }
+  };
+
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (pending) return;
+
+    const host = hostDraft.trim();
```
morgmart, line 57: 🤖 **P1 · SSH aliases collide with actions** (blocking)

The add dialog accepts any non-empty SSH host, while the selector reserves literal string values for the local destination and Add action in the same namespace used by host values. A valid SSH config alias matching either reserved value is interpreted as a command instead of a host.

**User effect:** People with a valid alias named __local__ or __add_ssh_environment__ cannot select that remote machine; Berd silently chooses the local computer or reopens the add dialog instead.

**Recommended fix:** Represent selector entries as discriminated host and action values so command identity cannot collide with user-controlled host names. If the component cannot support that, validate and clearly reject reserved aliases at first ingress.

**Test:** Add selector tests for both reserved-looking aliases and verify each can be added and selected as a host without triggering local selection or the Add action.

src-tauri/src/services/remote_backend/mod.rs
```diff
@@ -836,4 +879,43 @@ mod tests {
         assert_eq!(snapshot.len(), 1);
         assert_eq!(snapshot[0].host, "devbox");
     }
+
+    #[tokio::test]
+    async fn forget_removes_inactive_slot_by_its_exact_key() {
```
morgmart, line 976: 🤖 **P2 · Race test is vacuous** (non-blocking)

The test named pending_connect_observes_tombstone_after_forget retains an Arc, calls forget, and inspects the forgotten flag, but never starts a connect, blocks it on connect_lock, or verifies that an admitted pending connection is superseded without establishment side effects. It can pass even if the pending-connect guard is removed or misplaced.

**User effect:** A future regression could let a forgotten host finish connecting again while the test suite continues to report that the race is protected.

**Recommended fix:** Exercise the real ordering: hold the connection lock, admit a connect, forget the host, release the lock, and assert the connect is superseded and no tunnel or ready state is established.

**Test:** Rewrite this test as the described controlled pending-connect race and assert both the returned outcome and absence of connection side effects.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. The remote-host lifecycle can be overwritten by stale connection completions and status events, the legacy lock fallback can reclaim a live successor lock, and reserved selector values collide with valid SSH aliases. The changed tests do not honestly exercise one pending-connect race. This is a meaningful UI change, but the supplied PR evidence contains no screenshots or screen recording; please add screenshots or a short screen recording. GitHub check evidence is present and currently pending, so required checks still govern merge readiness.

Deterministic publication result: 3 blocking and 1 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 1 blocking screenshot-evidence requirement(s) in this review body.

src/features/remoteHosts/stores/remoteHostStore.ts
damienrj: 🤖 Fixed in 96523a27. Backend status now carries a per-slot incarnation plus generation, and the store rejects stale snapshots, events, and out-of-order connect completions. Added regressions for concurrent connects, Forget racing a connect, and stale events after replacement.

src-tauri/src/services/remote_backend/remote_daemon.sh
damienrj: 🤖 Fixed in 96523a27. Legacy reclamation now pins the exact observed lock-directory generation with an in-directory guard and rechecks ownership before moving it. Added a regression that pauses reclamation while a live successor takes over the shared pathname.

src/features/chat/ui/AddRemoteHostDialog.tsx
damienrj: 🤖 Fixed in 93a98e0e. Host values are now encoded in a dedicated host namespace, separate from selector actions. Added coverage proving aliases named __local__ and __add_ssh_environment__ are selected as SSH hosts.

src-tauri/src/services/remote_backend/mod.rs
damienrj: 🤖 Fixed in 96523a27. The regression now holds the connection mutex, queues real connect admission, starts Forget, verifies the tombstone before releasing the mutex, and asserts establishment side effects are never reached.

damienrj:
> 🤖 Automated code review
> 
> REQUEST_CHANGES. The remote-host lifecycle can be overwritten by stale connection completions and status events, the legacy lock fallback can reclaim a live successor lock, and reserved selector values collide with valid SSH aliases. The changed tests do not honestly exercise one pending-connect race. This is a meaningful UI change, but the supplied PR evidence contains no screenshots or screen recording; please add screenshots or a short screen recording. GitHub check evidence is present and currently pending, so required checks still govern merge readiness.
> 
> Deterministic publication result: 3 blocking and 1 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 1 blocking screenshot-evidence requirement(s) in this review body.
> 
> 🤖 **Blocking · Screenshots needed**
> 
> This PR changes Berd’s graphical interface. Please add screenshots or a short screen recording so the visual result can be reviewed. Screenshots are review evidence; they do not replace accessibility, responsive, theme, localization, or behavior validation.

image

src/features/chat/ui/AddRemoteHostDialog.tsx
```diff
@@ -0,0 +1,132 @@
...
+  onOpenChange,
+  onConnected,
+}: AddRemoteHostDialogProps) {
+  const { t } = useTranslation("chat");
+  const [hostDraft, setHostDraft] = useState("");
+  const [error, setError] = useState<string | null>(null);
+  const [pending, setPending] = useState(false);
+
+  const handleOpenChange = (nextOpen: boolean) => {
+    if (!nextOpen && pending) return;
```
morgmart: 🤖 **P1 · Pending connection traps dialog** (blocking)

When pending is true, handleOpenChange rejects every close request, the close button is hidden, and Cancel is disabled. Because ensureHostConnected has no presentation timeout or cancellation in this dialog, an SSH attempt that never settles leaves the modal permanently inescapable.

**User effect:** If connecting to an SSH host stalls, people cannot close the dialog or return to their chat and must restart or otherwise abandon the current app flow.

**Recommended fix:** Keep the dialog dismissible while connection work continues, and guard the async completion so a dismissed dialog cannot later select the host unexpectedly; alternatively provide an explicit cancellation path that reliably ends the attempt.

**Test:** Add a component test with a connection promise that never resolves, dismiss the dialog through a supported route, and verify the composer becomes usable and a later completion does not select the host after dismissal.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. While an SSH connection is pending, the new modal disables or suppresses every dismissal route, so a connection attempt that never completes can trap the user and block the chat workflow indefinitely. The supplied screenshot evidence covers this meaningful graphical UI change. The four prior review issues are fixed in the current comparison and their resolved threads contain substantive human-account replies. GitHub check evidence is structurally valid and currently pending; required checks still govern merge readiness.

src/features/chat/ui/AddRemoteHostDialog.tsx
damienrj: 🤖 Fixed in e6869b26. The dialog remains dismissible while connection work continues; closing invalidates the attempt so a late success cannot select the host. Cancel and close remain available. Added a deferred-connect regression that dismisses while pending, resolves later, and verifies onHostChange is not called.

src/features/chat/ui/AddRemoteHostDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+    if (!host) {
+      setError(t("toolbar.remoteHost.add.emptyHost"));
+      return;
+    }
+
+    setError(null);
+    setPending(true);
+    const attempt = ++attemptRef.current;
+    try {
+      await useRemoteHostStore.getState().ensureHostConnected(host);
```
morgmart: 🤖 **P1 · Stale connection still selected** (blocking)

ensureHostConnected can resolve after its store update was rejected because the host was forgotten or its lifecycle was superseded. The dialog treats every resolved promise as an accepted connection and calls onConnected as long as this local dialog attempt was not dismissed, so the UI can select a host that the store deliberately declined to make current.

**User effect:** If an SSH environment is forgotten or replaced while connecting, the composer can still switch to that obsolete host after the old attempt finishes.

**Recommended fix:** Make ensureHostConnected return an accepted-current result or reject with a normalized stale/cancelled outcome, and call onConnected only when that exact lifecycle became the active ready connection.

**Test:** Add a regression that submits a connection, forgets or supersedes the host before completion, resolves the old connection, and verifies the composer does not select it.

src/features/remoteHosts/stores/remoteHostStore.ts
```diff
@@ -272,6 +272,20 @@ export const useRemoteHostStore = create<RemoteHostStore>((set, get) => ({
...
       !current.forgottenHosts[host]
     ) {
+      // A manually entered host can already be ready when it was restored
+      // from the backend snapshot. Remember it even though no new connect is
+      // required, otherwise it disappears from the selector after restart.
+      if (
+        !current.configHosts.includes(host) &&
+        !current.manualHosts.includes(host)
+      ) {
+        const manualHosts = [host, ...current.manualHosts].slice(
```
morgmart: 🤖 **P2 · Concurrent ready hosts disappear** (blocking)

The already-ready branch derives manualHosts from the state snapshot captured at function entry, persists that list, and then replaces store state directly. Two close additions can both derive from the same old list, allowing the later write to discard the earlier host in memory and localStorage.

**User effect:** A manually added SSH environment can disappear from the selector after another ready environment is added at nearly the same time.

**Recommended fix:** Move the membership check and list derivation into a functional set update based on current state, and persist the list produced by that serialized update.

**Test:** Add a regression that admits two already-ready manual hosts from the same initial state and verifies both remain in the store and persisted storage.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A connection that is forgotten or superseded while the add dialog waits can still be selected in the composer when its promise resolves, and the already-ready persistence path can lose concurrent manual-host additions by writing from a stale snapshot. The supplied evidence includes a screenshot for this meaningful graphical UI change. All five prior automated findings are resolved, fixed in the current comparison, and have substantive human-account replies. GitHub check evidence is structurally valid; required checks still govern merge readiness.

src/features/chat/ui/AddRemoteHostDialog.tsx
damienrj: 🤖 Fixed in 103d1c2c. `ensureHostConnected` now returns whether its exact lifecycle became the current ready connection, and the dialog selects only an accepted current result. Superseded success or failure returns `false`. A two-lifecycle deferred regression verifies that when the replacement resolves first, the older completion cannot select the composer.

src/features/remoteHosts/stores/remoteHostStore.ts
damienrj: 🤖 Fixed in 103d1c2c. The already-ready path now rechecks and derives membership inside a functional store update from current state, persists that derived list, and returns its accepted status. A regression adds two already-ready manual hosts concurrently and verifies both remain in the store and localStorage.

morgmart (commented):
🤖 Automated code review

COMMENT. The implementation is clear across product flow, graphical UX, architecture, and engineering, and the seven prior automated findings are fixed with substantive human-account replies in their resolved threads. One product vocabulary decision remains: the new flow calls the existing SSH host concept an SSH environment while adjacent surfaces continue to call it a host. The supplied evidence includes a screenshot for this meaningful graphical UI change. GitHub check evidence is structurally valid; required checks still govern merge readiness.

damienrj:
🤖 Addressed in 67e7ccf3. The user-facing action, dialog title, and accessibility label now say “Add SSH host,” matching the existing “SSH hosts” terminology. “Environment” remains the broader concept of where a chat runs. The selector coverage was updated, and the focused tests plus the full frontend gate pass.

src/features/chat/ui/AddRemoteHostDialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    setError(null);
+    setPending(true);
+    const attempt = ++attemptRef.current;
+    try {
+      const accepted = await useRemoteHostStore
+        .getState()
+        .ensureHostConnected(host);
+      if (attemptRef.current !== attempt || !accepted) return;
```
morgmart: 🤖 **P1 · Superseded connection fails silently** (blocking)

When ensureHostConnected returns false because this lifecycle was forgotten or replaced, the submit handler returns without closing the dialog or setting feedback. The finally block only ends the loading state, so a meaningful connection outcome is reduced to an unexplained no-op.

**User effect:** If another connection replaces this attempt or the host is forgotten while it runs, Connect appears to do nothing and people cannot tell whether they should retry, close the dialog, or trust another connection.

**Recommended fix:** Return a named connected or superseded outcome from the store and show a clear recoverable message for the superseded case while preserving the rule that stale attempts never select the host.

**Test:** Extend the superseded-lifecycle component test to assert visible feedback and an available recovery action, in addition to verifying that the stale host is not selected.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A superseded or forgotten connection returns false to the add-host dialog, which silently stops loading and leaves the unchanged dialog open without explaining what happened or how to recover. The supplied evidence includes screenshots for this meaningful graphical UI change. The seven prior automated findings are resolved, fixed in the current comparison, and have substantive human-account replies. GitHub check evidence is structurally valid; required checks still govern merge readiness.

src/features/chat/ui/AddRemoteHostDialog.tsx
damienrj: 🤖 Fixed in 6146d859. `ensureHostConnected` now returns the named outcomes `connected` or `superseded`. The add-host dialog handles `superseded` by showing “This SSH connection changed while connecting. Try again or cancel.” while keeping Connect and Cancel available, and it still never selects the stale host. The component regression now asserts the visible alert and both recovery actions; the expanded remote-host suite passes 77 tests.

morgmart (approved):
🤖 Automated code review

APPROVE. No new publishable findings remain after the full static review. The shared connection outcome can still be silently ignored by the existing Settings caller, but that is the same underlying superseded-connection feedback issue as the protected resolved thread and is therefore suppressed. The eight prior automated threads are resolved with substantive human-account replies. Supplied evidence includes screenshots for this meaningful graphical UI change. GitHub check evidence is structurally valid; required checks still govern merge readiness.

**Pending checks:** 3 check(s) are not complete.

PR #283, opened 2026-09-03
feat(voice): add OpenAI Realtime Expert and Spokesperson mode

johnmatthewtennant:
**Category:** feature
**User Impact:** Users can choose OpenAI Realtime voice for a responsive spoken interface backed by the full Berd coding agent.

**Problem:** Chained speech-to-text and text-to-speech cannot provide the same low-latency, interruptible conversation as the OpenAI Realtime API, while a Realtime voice model alone cannot perform the full work available to a Berd agent.

**Solution:** Add an experimental Realtime voice mode that presents one seamless assistant using a fast Spokesperson for live conversation and the normal Berd Expert for tools and longer-running work. The two roles exchange ordered handoffs and results behind the scenes, with explicit debug rendering available for development. Chained voice remains the fresh-install default.

<details>
<summary>Key changes</summary>

- Add the Realtime WebRTC voice lifecycle, server VAD, barge-in, transcript correction, typed-message forwarding, pause/resume replay, and cross-window ownership.
- Add ordered Expert/Spokesperson handoffs with cursors, bounded reminders, SAY/context delivery, dismissal, and debug coordination bubbles.
- Preserve normal Expert thinking and tool output while rendering spoken turns consistently with Berd voice conversations.
- Add Voice settings for Realtime model, transcription, voice, speed, turn detection, VAD, noise reduction, reasoning, and subtle/debug presentation.
- Reuse the Berd-owned OpenAI voice credential and keep the long-lived key in Tauri while exposing only an ephemeral Realtime client secret to the renderer.
- Keep ordinary text sends and text-only transcript projection on their existing fast paths; bound cross-window Realtime status checks so voice coordination cannot hold prompt ownership indefinitely.
- Add berdctl routing commands plus deterministic text-driven evaluation scenarios and lifecycle regression coverage.

</details>

## Validation

- Independent Opus risk review cleared the final isolation changes after checking shared chat, transcript, release-gate, and cross-window paths.
- Focused Realtime/chat/projection tests: 26 passed. Release-script tests: 95 passed.
- Full pre-push gate passed: formatting, frontend checks, Tauri checks, and Clippy.


src/features/chat/hooks/useMessageQueue.ts
```diff
@@ -290,7 +290,10 @@ export function useMessageQueue(
 
         if (accepted === false) {
           let retryPayload = latestQueuedMessage.payload;
-          if (retryPayload.showInComposer === false) {
+          if (
+            retryPayload.showInComposer === false &&
+            retryPayload.sendOptions?.userMessageMetadata?.userVisible !== false
```
morgmart, line 303: 🤖 **P1 · Surface exhausted hidden coordination** (blocking)

This change keeps transport-only Realtime coordination hidden when a queued send is rejected by exempting records with `userVisible === false` from the existing reveal-on-failure path. The same retry path stops automatically after `MAX_CONSECUTIVE_REJECTIONS`, but its recovery assumption is that the record was forced visible. Under the new exception, the failed coordination record remains first in the queue with `showInComposer: false`, no retry timer, and no user-removal path. Queue ordering then prevents later messages from dispatching, and the matching persistence change restores the same blocking record hidden after restart.

**User effect:** If an Expert handoff or transcript cannot be delivered after the automatic retries, the chat can appear normal but stop sending every message behind that invisible item—even after the app restarts—with no way for the person to see or remove what is blocking it.

**Recommended fix:** Keep transport-only coordination hidden during ordinary retries, but give exhaustion an explicit terminal path: surface a localized actionable error and either make the failed record removable without exposing private coordination text, or remove or quarantine it and fail the Realtime conversation so later user messages can proceed. Do not persist a permanently hidden blocking head with no recovery action.

**Test:** Add a discriminating queue test that rejects a hidden `userVisible: false` record through the full retry budget, then enqueues a normal user message. Assert the terminal failure is visible and actionable and that the later message can eventually dispatch. Add the equivalent persistence and restart case to prove a hidden exhausted head cannot strand the queue.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. Corrected the completed static Engineering review against the exact three-dot comparison 5c31189a4d55437888bb8d0f921efada827c7d86...09f6b1b1f8318a389ac73df2e076f9559ff463f7 and the supplied current changed-line anchors. One blocking lifecycle finding remains: failed transport-only Realtime coordination can exhaust its retry budget while remaining an invisible, persistent queue head that blocks every later message. The earlier startup-cancellation finding was removed after re-checking the complete flow: stop invalidates the run, cleans resources, and late microphone completion is rejected by run identity, so the claimed stuck user state was not supported. Final self-check covered every changed user flow; accessibility; i18n/localization completeness; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system/project rules; duplicate/overlapping findings; and concrete evidence plus user effect for blocking severity. No findings fall outside the exact PR comparison or supplied anchors. The supplied GitHub evidence was inspected, records no local execution, and is structurally valid for the exact base and head; all ten supplied check runs completed successfully. Required checks still independently govern merge readiness.

src/features/chat/hooks/useMessageQueue.ts
johnmatthewtennant: 🤖 Fixed in 9fd66436. Hidden transport-only coordination now has an explicit terminal exhaustion path: Berd shows a localized generic error without exposing the coordination payload, removes and persists removal of the blocking queue head, stops only the matching Realtime session, and immediately lets the next user message drain. The regression test exhausts all five attempts, queues a normal message behind the hidden record, verifies that message sends, and verifies the restart cache cannot restore the exhausted record.

morgmart (approved):
🤖 Automated code review

APPROVE. Completed a fresh static Engineering review of the exact three-dot comparison 5c31189a4d55437888bb8d0f921efada827c7d86...9fd66436217a6e1b14b801f804ae37b7916cb2f5 after explicitly applying the project code-review skill and the user wes-review skill. No publishable findings remain. The previously reported hidden-coordination queue failure is fixed at this head: after the bounded retry budget, the implementation emits localized generic feedback without exposing private coordination text, removes and persists removal of the blocking queue head, stops only the matching Realtime session, and allows the next queued message to drain; the new regression test covers exhaustion, subsequent user-message delivery, payload privacy, and restart persistence. The resolved authoritative thread has a substantive human reply and was not re-raised. The review traced Realtime startup and teardown, microphone and credential boundaries, typed and spoken transcript delivery, Expert and Spokesperson handoffs and reminders, cross-window bridge routing, queue persistence and recovery, session promotion, navigation and consent guards, CLI contracts, release gates, accessibility, localization, and changed-test honesty. Final self-check covered every changed user flow; accessibility; i18n/localization completeness; navigation and consent guards; async, error, never-completes, lifecycle, and race behavior; test honesty; design-system/project rules; duplicate and overlapping findings; and concrete evidence plus user effect for blocking severity. Unsupported claims were removed. The supplied GitHub evidence was inspected, records no local execution, is structurally valid for the exact base and head, and reports ten successful check runs. Required checks still independently govern merge readiness.

PR #287, opened 2026-09-03
feat(chat): add starred models to the agent model picker

NickTitle:
## What

Star/unstar any model from the agent model picker. Starred models surface in a Starred section pinned at the top of the picker lists (collapsed and expanded views), with a divider after the starred block.

## How

- New `src/features/chat/lib/starredModels.ts`: localStorage persistence (key `goose:starredModels:v1`), keyed per `(scopeId, modelId)`, with a window event to sync pickers across windows.
- New `useStarredModels` hook wrapping the lib for React components.
- `AgentModelPickerLists`: starred models are grouped first in both the shortlist and the sorted/browse views; recents and recommended entries exclude starred models so nothing appears twice. Starred items integrate with the existing recents/shortlist logic.
- `AgentModelPicker` rows get a star toggle button with en/es strings ("Star {{model}}" / "Unstar {{model}}").
- `Separator` moves into the design-system explorer used-components list (it is now rendered by the picker).

## Testing

- `AgentModelPicker.test.tsx` extended for star/unstar, starred-section ordering, and persistence; 51 tests pass.
- `pnpm typecheck` and `design-system:coverage --strict` pass.

Co-authored-by: Goose <opensource@block.xyz>
Screenshot 2026-09-02 at 1 42 25 PM


src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -199,30 +228,44 @@ export const RecommendedModelList = forwardRef<
...
         (m) =>
-          !recent.some((r) => r.id === m.id && r.providerId === m.providerId),
+          !recent.some((r) => r.id === m.id && r.providerId === m.providerId) &&
+          !starredKeys.has(modelStarKey(m.providerId ?? selectedAgentId, m.id)),
       );
     const shortlist = [...recent, ...rec];
     if (
       currentModelId &&
-      shortlist.length > 0 &&
+      starred.length + shortlist.length > 0 &&
```
morgmart, line 412: 🤖 **P1 · Keep saved stars visible** (blocking)

The new compact-list branch at this condition returns the current model plus the recent/recommended shortlist when the current model is outside those groups, but it omits the separately computed starred group. For example, after starring one non-recommended model and selecting a different non-recommended model, reopening the compact picker hides the saved star even though it remains persisted.

**User effect:** A model someone deliberately starred can disappear from the picker exactly when they switch to another less-common model, making the saved shortcut look lost and forcing them back into View more or search.

**Recommended fix:** Compose every compact-list branch from the same deduplicated groups, preserving the intended starred/current/recent/recommended ordering instead of returning before starred models are included.

**Test:** Add a regression test with one non-recommended starred model and a different non-recommended current model; reopening the compact picker must still show the starred model in the starred group.

src/features/chat/ui/AgentModelPicker.tsx
```diff
@@ -500,7 +500,7 @@ export function AgentModelPicker({
           // gated single-column layout has no dead vertical space below the
           // model list.
           "flex max-h-[min(24rem,50vh)] flex-col overflow-hidden p-1 transition-[width] duration-[240ms] ease-[cubic-bezier(0.2,0,0,1)]",
-          isWidePicker ? "w-[37.25rem]" : "w-[26.25rem]",
+          isWidePicker ? "w-[39.25rem]" : "w-[28.25rem]",
```
morgmart: 🤖 **P1 · Fit the supported window** (blocking)

The expanded picker is now fixed at 39.25rem (628px), while Berd's configured minimum window width is 608px. Radix collision positioning can move the popover but this surface has no width cap or narrow layout, so the three-column picker cannot fit inside a supported viewport even before collision padding.

**User effect:** At a valid narrow Berd window, part of the picker can be clipped off-screen, leaving model details, star actions, or another column difficult or impossible to inspect and use.

**Recommended fix:** Cap the popover to the collision-aware available width and make its columns adapt at narrow widths rather than applying an unconditional 628px width.

**Test:** Add a component/layout regression at the 608px supported minimum viewport that asserts the expanded picker stays within the available horizontal bounds and all columns remain reachable.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -380,33 +448,67 @@ export const RecommendedModelList = forwardRef<
...
+                        </div>
+                      </div>
+                      {isSelected ? (
+                        <IconCheck className="size-4 shrink-0 text-muted-foreground" />
+                      ) : null}
+                    </PickerItem>
+                    <button
+                      type="button"
+                      onClick={() => toggleStar(scopeId, model.id)}
+                      className="flex size-7 shrink-0 items-center justify-center rounded-sm text-muted-foreground/50 transition-colors hover:bg-accent hover:text-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
```
morgmart: 🤖 **P1 · Use a perceptible star action** (blocking)

The new raw button hand-authors size, radius, color, hover, and focus behavior instead of using the required shared Button primitive. Its idle icon uses text-muted-foreground at 50% opacity, which resolves to roughly 2.1:1 contrast on the light popover and 2.5:1 on the dark popover, below the 3:1 non-text contrast requirement for an interactive graphical control.

**User effect:** Low-vision users may not perceive that models can be starred until they happen to hover or focus the faint icon, and the control behaves visually unlike Berd's other icon actions.

**Recommended fix:** Render this through the shared Button icon API and use an approved semantic idle/pressed treatment that reaches at least 3:1 contrast in both themes; extend the shared component with a semantic selected state if the current API cannot express it.

**Test:** Add component/design-system coverage confirming the star action uses the shared Button contract and that its idle foreground token meets non-text contrast against the popover in light and dark themes.

src/features/chat/lib/starredModels.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  window.dispatchEvent(new CustomEvent(STARRED_MODELS_CHANGED_EVENT));
+}
+
+export function getStarredModelKeys(): StarredModelSet {
+  return readStarredModels();
+}
+
+export function toggleModelStar(scopeId: string, modelId: string): void {
+  const next = readStarredModels();
```
morgmart: 🤖 **P2 · Preserve concurrent star changes** (blocking)

Each toggle reads the complete shared set and writes a replacement set. Two Berd windows starting from the same snapshot can star different models before either receives the other's storage event; the last full-set write then silently removes the other change. Event-based refresh synchronizes the resulting value but does not make the read-modify-write transaction safe.

**User effect:** When someone updates starred models in two open Berd windows around the same time, a star saved in one window can unexpectedly vanish from both.

**Recommended fix:** Give mutations a single serialization owner or persist independent per-model states so unrelated updates cannot replace the entire collection, then reconcile external storage changes through that owner.

**Test:** Add a storage-layer test with two independent window snapshots that interleave different star operations from the same initial state and verify both changes survive.

src/features/chat/lib/starredModels.ts
```diff
@@ -0,0 +1,69 @@
...
+  try {
+    if (models.size === 0) {
+      window.localStorage.removeItem(STARRED_MODELS_STORAGE_KEY);
+    } else {
+      window.localStorage.setItem(
+        STARRED_MODELS_STORAGE_KEY,
+        JSON.stringify([...models]),
+      );
+    }
+  } catch {
```
morgmart, line 67: 🤖 **P2 · Report failed star saves** (non-blocking)

The persistence catch discards local-storage failures and still dispatches the changed event. The hook then rereads the unchanged storage value, so the action appears to do nothing without exposing that the preference could not be saved.

**User effect:** If preference storage is unavailable, pressing the star gives no confirmation or explanation; people cannot tell whether the click was missed or Berd cannot retain the choice.

**Recommended fix:** Return a persistence result from the storage boundary and let the picker show a concise failure outcome, or keep a clearly defined in-memory fallback while explaining that it will not persist.

**Test:** Add a regression test where localStorage.setItem/removeItem throws and verify the user receives a visible failure outcome rather than an unexplained unchanged control.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the exact PR comparison has four blocking findings and one non-blocking finding. The core starred-model flow can hide saved stars, the expanded picker exceeds Berd's supported minimum window width, the new idle star control fails non-text contrast and bypasses the shared Button system, and concurrent window updates can overwrite saved intent. A failed local-storage write also produces no user-facing outcome. Supplied GitHub evidence is structurally valid for the exact head; some checks remain in progress, so required checks still govern merge readiness.

Deterministic publication result: 4 blocking and 1 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

morgmart:
I’m aligned with the feature and would like to keep it, but the current visual treatment feels too heavy. Please refine the interaction and styling:

- Show the outline star only when its row is hovered or receives keyboard focus, rather than displaying it at all times. This keeps the list from feeling overloaded while preserving the existing affordance.
- Make the favorited state less visually dominant. Use a lighter semantic treatment rather than the current harsh filled-black star.
- The selected treatment should cover the entire row, including the trailing action area—not stop before the star.
- Avoid showing competing selection and favorite icons. Either:
- Let the checkmark take precedence over the favorited state, while still revealing the outline star on hover or keyboard focus so a selected model can be favorited; or
- Remove the checkmark from all model-picker columns and rely on the selected-row fill state. I’m comfortable with this option as a designer, but I’ll leave the final choice to the PR owner.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found one new non-blocking localization issue. Four still-present issues are suppressed because unresolved review threads already cover them, and the previously reported compact-list omission is fixed. Supplied GitHub evidence is structurally valid for the exact head and all captured checks are successful.

Deterministic publication result: 0 blocking and 1 non-blocking inline finding(s) publishable; 4 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/shared/i18n/locales/es/chat.json
```diff
@@ -621,7 +621,9 @@
         "agent-speaking": "El agente está hablando…",
         "error": "Error de conversación de voz: {{error}}"
       }
-    }
+    },
+    "starModel": "Destacar {{model}}",
+    "unstarModel": "Quitar destaque de {{model}}"
```
morgmart: 🤖 **P2 · Use idiomatic Spanish unstar copy** (non-blocking)

The new icon-only action uses “Quitar destaque de {{model}}” as its accessible name. That construction is awkward and non-idiomatic Spanish, making the action less clear than its English equivalent.

**User effect:** People using Berd in Spanish, especially with a screen reader, hear an unnatural description when they try to remove a model from their starred choices.

**Recommended fix:** Use an idiomatic label aligned with the chosen Spanish term, such as “Quitar {{model}} de destacados” or “Dejar de destacar {{model}}”.

**Test:** Update the locale assertion or accessibility-name coverage to expect the revised Spanish label with the model name interpolated.

NickTitle:
🤖 **P1 #3 resolved** — `d8b07c59` renders the star action through the shared `Button` primitive (`variant="ghost"`, `size="icon-xs"`) and extends `Button` with a ghost-only `selected` flag for pressed/added-state toggles:

- Idle (unstarred) star rests on the ghost icon contract's `muted-foreground`: ≈5.7:1 on the light popover, ≈6.1:1 on the dark popover — above the 3:1 WCAG non-text contrast bar for icon-only controls.
- Favorited rows soften to `text-foreground/80` (≈8:1 light, ≈11:1 dark) via the new `selected` flag, keeping the hover-reveal and keyboard focus behavior.
- Coverage: `globals.test.ts` now enforces the `muted-foreground`/`popover` pairing ≥3:1 in both themes; the picker test asserts the shared Button contract (`data-slot="button"`), idle/selected treatments, and `aria-pressed` state.
- Verification: 52 picker tests + related suites (237 total) pass; `pnpm typecheck`, `biome check`, and the full design-system check chain (manifest, tokens, audit, coverage) pass.

**Review-infra note:** the Builderbot review loop was dispatched for this SHA, but the routine run failed at GitHub submission (`403 Resource not accessible by integration` on create-review); re-dispatch deduped to the same failed run, so findings could not be retrieved. Proceeding per the review-outage policy.

**Still open** (next commits):
1. P2 storage: per-key storage entries for `toggleModelStar` (cross-window last-write-wins on the whole-set read-modify-write).
2. P2 storage: surface `persistStarredModels` quota failures to the user instead of swallowing them.

src/features/chat/hooks/useStarredModels.ts
```diff
@@ -0,0 +1,66 @@
...
+  cachedSnapshot = null;
+}
+
+function getSnapshot(): Set<string> {
+  if (cachedSnapshot === null) {
+    cachedSnapshot = getStarredModelKeys();
+  }
+  return cachedSnapshot;
+}
+
```
morgmart, line 24: 🤖 **P1 · Refresh stars after remount** (blocking)

The module-level cached snapshot survives after the final subscriber unmounts, but the storage listener is removed. If another Berd window changes a star during that gap, no event invalidates the cache; remounting subscribes again but getSnapshot can keep returning the old Set because it only reads storage when the cache is null.

**User effect:** After changing favorites in another Berd window, reopening the model picker can continue showing the old starred choices instead of the current saved state.

**Recommended fix:** Close the missed-event window by invalidating and rereading the snapshot when subscription begins after listeners are attached, or compare a stable storage fingerprint whenever the snapshot is requested.

**Test:** Add a regression that mounts and reads the store, unmounts its final subscriber, mutates a star as an external window would, then remounts and verifies the new stored state appears.

src/features/chat/lib/starredModels.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/**
+ * One-shot migration from the pre-#287 aggregate format (one array under a
+ * single key). Idempotent and safe to race across windows: entry writes are
+ * identical values, and removing the legacy key twice is a no-op.
+ */
+function migrateLegacyStarredModels(): void {
+  const legacy = window.localStorage.getItem(LEGACY_STARRED_MODELS_STORAGE_KEY);
+  if (legacy === null) {
```
morgmart: 🤖 **P1 · Commit migration after writes** (blocking)

The legacy aggregate key is removed before its JSON is parsed and before every replacement entry is written. If parsing succeeds but a later setItem fails because storage is unavailable or full, the original preference has already been deleted and the migration leaves an empty or partial result with no safe retry source.

**User effect:** On upgrade, people can permanently lose some or all of their saved model favorites when device storage encounters a migration failure.

**Recommended fix:** Parse and validate the legacy value first, write every replacement entry successfully, and remove the legacy key only as the final commit step. Preserve the legacy value on any failure so migration can retry.

**Test:** Add a migration test where one replacement setItem throws and verify the legacy aggregate remains intact and can be migrated successfully on a later read.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -380,33 +485,78 @@ export const RecommendedModelList = forwardRef<
...
+                    data-model-key={modelKey}
+                    data-starred={starred || undefined}
+                  >
+                    <PickerItem
+                      onClick={() => {
+                        onModelSelect(model);
+                        resetView();
+                      }}
+                      selected={isSelected}
+                      className="w-auto flex-1 justify-between"
```
morgmart, line 817: 🤖 **P2 · Unify the selected model row** (blocking)

The selected background is still owned only by PickerItem while the star action is a sibling outside that surface. The trailing area and gap remain visually detached, and interacting with a selected starred model can reveal the favorite icon alongside the checkmark rather than establishing one clear status hierarchy.

**User effect:** A selected model looks like two adjacent controls instead of one coherent row, and selection competes visually with favorite state.

**Recommended fix:** Let the complete outer row own the selected surface, including the trailing action, and apply one of the accepted icon hierarchies: give the checkmark precedence while keeping a quiet interactive star affordance, or remove picker checkmarks and rely on the unified fill state.

**Test:** Add component coverage for selected starred and unstarred rows that verifies one continuous selected surface and the chosen single-icon hierarchy at rest, hover, and keyboard focus.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh review of the exact three-dot comparison found three blocking issues. Star state can remain stale after a picker remount, the new legacy migration can delete saved preferences before replacement writes succeed, and the selected model treatment still excludes the trailing star action. Two still-present issues are suppressed by unresolved review threads. Supplied GitHub evidence is structurally valid for the exact head; several checks remain in progress, so required checks still govern merge readiness.

Deterministic publication result: 3 blocking and 0 non-blocking inline finding(s) publishable; 2 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/lib/starredModels.ts
```diff
@@ -0,0 +1,132 @@
...
+  return JSON.stringify([scopeId, modelId]);
+}
+
+/** localStorage key of the single entry that records one starred model. */
+export function starredModelStorageKey(starKey: string): string {
+  return STARRED_MODELS_ENTRY_PREFIX + encodeURIComponent(starKey);
+}
+
+/**
+ * One-shot migration from the pre-#287 aggregate format (one array under a
```
morgmart: 🤖 **P2 · Remove the unshipped migration** (non-blocking)

The exact base contains no starred-model storage owner or aggregate goose:starredModels:v1 format; that format existed only in an earlier revision of this PR. Treating it as legacy turns an abandoned implementation detail into a durable compatibility contract and adds destructive migration machinery for user state no released base created.

**User effect:** People gain no compatibility benefit, while future maintenance must preserve and reason about an extra storage format and its failure modes indefinitely.

**Recommended fix:** Remove the legacy aggregate key, migration function, and migration tests unless an intermediate build carrying this format was actually distributed. If it was distributed, document that requirement and retain a transaction-safe migration.

**Test:** Remove the obsolete migration test with the code, or, if compatibility is required, replace it with a documented regression that proves the distributed format migrates transactionally without deleting its source on failure.

src/shared/ui/button.tsx
```diff
@@ -296,6 +306,16 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> &
      * on ghost; other variants ignore it (dev builds warn).
      */
     flush?: boolean;
+    /**
+     * On/pressed state for toggle controls that lean on Button as their
```
morgmart, line 310: 🤖 **P2 · Unify toggle state ownership** (non-blocking)

The shared Button API documents selected as its on/pressed state and uses it for visual treatment, but semantic state remains a separate aria-pressed attribute supplied by each caller. These independent inputs can diverge, allowing a control to look active while assistive technology announces it as inactive, or vice versa.

**User effect:** People using assistive technology can be told a different toggle state from the one sighted users see when a caller does not manually synchronize both properties.

**Recommended fix:** Give the shared toggle state one owner: prefer a pressed contract that applies both styling and aria-pressed, or derive aria-pressed from the supported selected prop. Alternatively, use or extend the existing Toggle primitive for this control.

**Test:** Add shared-primitive coverage that supplies one state input and verifies both visual styling and accessible pressed state for true, false, omitted, and unsupported-variant cases.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found two new non-blocking architecture/design-system issues. The PR retains migration machinery for an aggregate format absent from the exact base, and the shared Button visual toggle state can drift from its separately supplied accessible pressed state. Five still-present issues are suppressed by unresolved review threads. Supplied GitHub evidence is structurally valid for the exact head and all captured checks are successful.

Deterministic publication result: 0 blocking and 2 non-blocking inline finding(s) publishable; 5 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -149,9 +193,83 @@ export const RecommendedModelList = forwardRef<
...
+    modelId: string;
+    state: StarAnimation;
+  } | null>(null);
   const [query, setQuery] = useState("");
+  useEffect(() => {
+    if (!starAnimation || prefersReducedMotion) {
+      return;
+    }
+    if (starAnimation.state.phase === "out") {
+      const timer = window.setTimeout(() => {
```
morgmart: 🤖 **P1 · Commit stars before animating** (blocking)

The click only records animation state; the actual preference mutation waits 240ms inside this effect. If the picker closes, switches agents, or otherwise unmounts during that interval, effect cleanup cancels the timer and the action is never saved. The delayed call also toggles whatever storage state exists later rather than committing the captured target, so an intervening cross-window change can produce the opposite result. A single shared animation guard additionally drops attempts to star another row while any animation is active.

**User effect:** A person can click a star and immediately continue or close the picker, then return to find the model was never favorited. Quickly favoriting multiple models can silently lose later clicks.

**Recommended fix:** Persist the explicit requested target state synchronously in the click handler, then animate the already-committed result. Keep animation state row-local and non-blocking so presentation lifecycle cannot cancel or globally lock user intent.

**Test:** Add regressions that click Star and immediately close or unmount before 240ms, inject an external same-entry change during the visual transition, and click two different rows quickly; every requested final state must persist.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh review of the exact three-dot comparison found one blocking issue. The new decorative star animation delays persistence into a cancellable component effect, so closing the picker can discard the click and concurrent state changes can invert the requested result. Seven still-present issues are suppressed by unresolved review threads. Supplied GitHub evidence is structurally valid for the exact head; some checks remain in progress, so required checks still govern merge readiness.

Deterministic publication result: 1 blocking and 0 non-blocking inline finding(s) publishable; 7 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPicker.tsx
```diff
@@ -366,10 +368,28 @@ export function AgentModelPicker({
...
 
-  const handleModelSelect = (model: ModelOption) => {
-    recordModelSelection(selectedAgentId, model);
+  const pendingCrossAgentModelRef = useRef<{
+    agentId: string;
+    model: ModelOption;
+  } | null>(null);
+  const handleModelSelect = (model: ModelOption, agentId: string) => {
+    if (agentId !== selectedAgentId) {
+      pendingCrossAgentModelRef.current = { agentId, model };
```
morgmart: 🤖 **P1 · Apply one cross-agent target** (blocking)

A foreign-agent favorite is split into two independently owned transitions: the picker stores the model in a ref and first calls onAgentChange, whose controller path can apply that agent’s preferred model or recreate and navigate the session; only after selectedAgentId changes does this effect call onModelChange for the clicked model. Queued work can observe the intermediate target, navigation can unmount and lose the ref, and a failed first transition can leave stale intent that applies during a later unrelated switch.

**User effect:** Choosing a favorite from another agent can move the chat to that agent’s default model instead of the model clicked, briefly run queued work on the wrong target, or unexpectedly apply the old choice during a later switch.

**Recommended fix:** Emit one agent-and-model selection intent from the picker and let the session/controller owner apply that final execution target atomically through its existing versioning, recreation, navigation, and rollback machinery. Do not bridge session transitions with component-local ref state.

**Test:** Cover a started session with queued work, a recreation/navigation path, a failed or rejected agent switch, and a later unrelated switch. The click must produce one final agent/model target, never expose an intermediate preferred model, and never leave stale pending intent.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -368,45 +595,181 @@ export const RecommendedModelList = forwardRef<
...
+                    <Separator
+                      className="my-1"
+                      data-testid="starred-models-divider"
+                    />
+                  </motion.div>
+                );
+              }
+
+              const { model } = item;
+              const modelAgentId = modelAgentIds.get(model) ?? selectedAgentId;
```
morgmart, line 666: 🤖 **P2 · Scope selection to the agent** (blocking)

Cross-agent favorites now share one model list, but selected state is still computed only from model ID and provider ID. When currentModelProviderId is absent, the existing fallback matches by model ID alone, so a favorite from another agent with the same model ID can receive the selected background, checkmark, and selected ordering.

**User effect:** The picker can show multiple models as selected at once, including one owned by an inactive agent, making it unclear which agent and model will answer the next message.

**Recommended fix:** Require modelAgentId to equal selectedAgentId before applying model/provider selection matching. Keep providerless model-ID fallback scoped to the active agent only.

**Test:** Render same-ID providerless models for the selected agent and another agent, then verify only the active agent’s row receives selected styling, checkmark, and selected ordering.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh review of the exact three-dot comparison found two blocking issues in the new cross-agent favorites flow. Selecting a foreign-agent favorite coordinates two separate session transitions through component-local state, and same-ID models from different agents can both render as selected. Eight still-present issues are suppressed by unresolved review threads. Supplied GitHub evidence is structurally valid for the exact head and all captured checks are successful. A product decision is also needed on whether cross-agent favorites should bypass the existing gated agent-switch flow.

Deterministic publication result: 2 blocking and 0 non-blocking inline finding(s) publishable; 8 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -368,45 +647,231 @@ export const RecommendedModelList = forwardRef<
...
```
morgmart, line 666: 🤖 **P2 · Name cross-agent favorite owners** (non-blocking)

The combined favorites list knows each row’s owning agent, but communicates it only through an optional provider icon and native title. getProviderIcon returns null for unknown or custom agents, and the model-selection button’s accessible name remains only the model name. Two agents exposing the same model name therefore produce indistinguishable controls, even though selecting a foreign row changes the execution target.

**User effect:** People may not be able to tell which agent owns a favorite before selecting it, and screen-reader users can hear identical model choices that lead to different agents.

**Recommended fix:** Include the owning agent’s localized label as compact visible metadata for foreign-agent favorites and in the model-selection control’s accessible name. Use an existing secondary-text row pattern instead of relying only on branded icons.

**Test:** Render same-named favorites for the active agent, a known foreign agent, and a custom agent without an icon. Verify every foreign row visibly names its agent and has a unique accessible name.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found one new non-blocking accessibility and clarity issue: cross-agent favorite rows do not reliably expose their owning agent. Nine still-present issues are suppressed by unresolved review threads. Supplied GitHub evidence is structurally valid for the exact head; some checks remain in progress, so required checks still govern merge readiness. Favoriting and cross-agent favorites are accepted product direction; no escalation is needed.

Deterministic publication result: 0 blocking and 1 non-blocking inline finding(s) publishable; 9 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed. Compact-list construction now includes the starred group in every branch and deduplicates starred/current/shortlist rows. Regression coverage keeps a non-recommended favorite visible when a different non-recommended model is current.

src/features/chat/ui/AgentModelPicker.tsx
NickTitle: 🤖 Fixed. Compact and expanded widths are capped with `min(..., calc(100vw - 1.5rem))`, so the picker fits the supported narrow viewport while retaining reachable columns.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `d8b07c59`. The star uses the shared `Button` ghost icon contract. Idle and selected colors meet the required contrast, and design-system plus picker coverage enforce the contract.

src/features/chat/lib/starredModels.ts
NickTitle: 🤖 Fixed in `af9ffaa1`. Each favorite is persisted under an independent per-model key, so concurrent changes to different models no longer replace a shared aggregate set. Storage-event sync now watches the entry prefix.

src/features/chat/lib/starredModels.ts
NickTitle: 🤖 Fixed in `34d27b32`. Failed localStorage writes return failure, show a localized error toast, and leave the rendered state aligned with persisted state. Tests cover failed add and remove writes.

src/shared/i18n/locales/es/chat.json
NickTitle: 🤖 Fixed in `d63418cd`. The Spanish accessible label is now “Quitar {{model}} de destacados,” which is idiomatic and preserves model-name interpolation.

src/features/chat/lib/starredModels.ts
NickTitle: 🤖 Fixed in `d63418cd`. Removed the aggregate key, migration function, export, and migration test because that format was never shipped.

src/features/chat/hooks/useStarredModels.ts
NickTitle: 🤖 Fixed in `d63418cd`. `subscribe` now attaches both listeners and invalidates the module snapshot, so a final-subscriber remount rereads storage changes made while no listener existed.

src/shared/ui/button.tsx
NickTitle: 🤖 Fixed in `d63418cd`. For supported ghost toggles, `selected` now derives both visual state and `aria-pressed`. Shared Button tests cover true, false, omitted, and unsupported variants.

src/features/chat/lib/starredModels.ts
NickTitle: 🤖 Resolved by removing the never-shipped aggregate migration in `d63418cd`; there is no destructive migration path or source key to lose.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed and covered in `6126f3c5`. The complete outer row owns the continuous selected background, including the trailing star area. The model checkmark was removed, leaving the row fill as the selection signal while the quiet star remains independently interactive.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `6126f3c5`. Selected state now requires the row owner to equal the active agent before model/provider matching. A same-ID cross-agent regression verifies exactly one selected row.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `6126f3c5`. Foreign favorites now show the owning agent as quiet secondary text and include it in the model button accessible name. Labels come from picker agent metadata and fall back for custom agents without icons.

morgmart (commented):
🤖 Automated code review

No new publishable findings were identified in the fresh review of the exact three-dot comparison. Two still-present blocking issues are suppressed by their existing unresolved threads, which have no substantive human replies; approval therefore fails closed and no publication/retry is recommended. The supplied GitHub evidence is structurally valid for the exact head. The frontend check is failed and still governs merge readiness. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 0 blocking and 0 non-blocking inline finding(s) publishable; 2 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `bbd315b9`. The click handler now persists an explicit target state synchronously; animation is presentation-only and cannot cancel or invert the saved intent. Animation state is per row, so rapid clicks on different rows no longer block each other. Regressions cover immediate unmount, an external same-entry change during animation, and two rapid row clicks.

src/features/chat/ui/AgentModelPicker.tsx
NickTitle: 🤖 Fixed in `97e01fbf`. A foreign favorite now emits one model selection carrying its owning agent. `useResolvedAgentModelPicker` builds and publishes one complete execution target through the existing versioning, pending-target, recreation, navigation, and rollback owner. The component-local ref/effect bridge is removed, provider fallback is scoped to the requested agent, and same-ID no-op checks cannot drop foreign choices. Coverage includes pending and started sessions; the existing recreation, failure, supersession, and dispatch suites remain green.

src/features/chat/hooks/useResolvedAgentModelPicker.ts
```diff
@@ -505,21 +506,24 @@ export function useResolvedAgentModelPicker({
           console.error("Failed to update ACP session provider:", error);
         });
     },
-    onModelSelected: (model) => {
+    onModelSelected: (model, requestedAgentId) => {
+      const targetAgentId = requestedAgentId ?? selectedAgentId;
```
morgmart, line 510: 🤖 **P1 · Enforce favorite agent readiness** (blocking)

The direct cross-agent model path accepts requestedAgentId and builds a session target without passing through the existing provider-change owner that rejects agents absent from readyAgentIds. The favorite inventory can still contain cached models for agents marked not_ready or not_installed, so selecting one can publish a target intent before its agent is connected or installed.

**User effect:** A saved favorite for a disconnected or unavailable agent can appear selectable and then change the draft or session toward an agent that cannot run, instead of taking the person through Connect or Install.

**Recommended fix:** Validate cross-agent favorite readiness through the same owner used for normal agent changes before publishing the combined agent-and-model intent. Keep the favorite saved, but route selection to setup or leave it unavailable until setup succeeds.

**Test:** Seed a cached favorite for a foreign agent marked not_ready and not_installed, select it from both a pending draft and an existing session, and verify no target intent, token reset, or backend transition occurs before setup succeeds.

src/features/chat/ui/AgentModelPicker.tsx
```diff
@@ -701,11 +692,15 @@ export function AgentModelPicker({
                   key={selectedAgentId}
                   ref={modelListRef}
                   models={displayedModels}
+                  favoriteModels={favoriteModels}
```
morgmart, line 707: 🤖 **P2 · Show favorites from unavailable agents** (blocking)

RecommendedModelList is rendered only when displayedModels for the currently selected agent is non-empty. If that agent has no available models, the component takes the empty-state branch without considering valid favorites supplied from other agents.

**User effect:** Saved favorites disappear exactly when the active agent has no usable models, so people see “No models available” and lose the direct shortcut that could move them to a working agent.

**Recommended fix:** Render the model list whenever either the active agent has displayed models or the supplied favorite inventory contains a live favorite. Preserve the empty state only when neither source can provide a selectable row.

**Test:** Render an active agent with an empty model catalog and a live favorite from another ready agent, then verify the favorite is visible and selectable instead of the no-models empty state.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -368,45 +645,271 @@ export const RecommendedModelList = forwardRef<
...
+                        // muted-foreground — ≈5.7:1 light / ≈6.1:1 dark against
+                        // the popover, above the 3:1 WCAG non-text bar
+                        // (enforced in globals.test.ts) — and favorited rows
+                        // soften to foreground/80 via the selected flag.
+                        className={cn(
+                          "shrink-0",
+                          activeStarAnimation?.phase === "moving" &&
+                            "pointer-events-none",
+                        )}
+                        aria-label={t(
```
morgmart: 🤖 **P2 · Disambiguate cross-agent star controls** (non-blocking)

Foreign model-selection rows now include the owning agent in visible metadata and their accessible names, but the adjacent star toggle still announces only Star or Unstar plus the model name. Same-named favorites from different agents therefore expose duplicate, indistinguishable star controls to assistive technology.

**User effect:** Screen-reader users can hear multiple identical favorite actions and cannot tell which agent’s model they are about to add or remove.

**Recommended fix:** Include the foreign agent label in a localized star and unstar accessible-name variant while retaining the shorter label for models owned by the active agent.

**Test:** Render same-named favorites from two agents and verify each foreign star toggle has a unique localized accessible name containing its owning agent.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh review of the exact three-dot comparison found two blocking behavior issues and one non-blocking accessibility issue in the new cross-agent favorites flow. Foreign favorites can bypass agent readiness, all foreign favorites disappear when the active agent has no models, and duplicate cross-agent star controls omit the owning agent from their accessible names. Supplied GitHub evidence is structurally valid for the exact head; some checks remain in progress, so required checks still govern merge readiness. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 2 blocking and 1 non-blocking inline finding(s) publishable; 0 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/shared/ui/GlobalComposerPill.tsx
```diff
@@ -1513,6 +1513,12 @@ export function GlobalComposerPill({
             }
             currentModelName={effectiveModelSelection?.modelName ?? null}
             availableModels={availableModels}
+            favoriteModels={pickerAgents.flatMap((agent) =>
```
morgmart, line 1512: 🤖 **P1 · Route global favorites to their agent** (blocking)

The Global Composer supplies favorites from every agent to AgentModelPicker, whose cross-agent callback now passes the owning agent as a third argument. This composer’s onModelSelected handler still accepts only the model and derives both the selection and execution target from the currently selected agent, discarding that owner. A foreign favorite can therefore be paired with the wrong harness or provider.

**User effect:** Choosing a favorite from another agent in the Global Composer can start a chat with the current agent and an incompatible or unintended model instead of the agent/model pair the person selected.

**Recommended fix:** Accept the optional owning agent in the Global Composer’s model-selection handler and build the provider override, model selection, and execution target from that explicit agent/model pair, using the same readiness guard already added to the shared picker state.

**Test:** In the Global Composer, select a ready favorite owned by another agent and verify the resulting execution target, provider override, selected provider, and model override all use that favorite’s agent and model rather than the previously active agent.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh review of the exact three-dot comparison found one blocking integration bug. The Global Composer exposes cross-agent favorites but its model-selection callback ignores the owning agent, so it can build the wrong execution target. Three still-present issues are suppressed by unresolved threads. Supplied GitHub evidence is structurally valid for the exact head; only three checks were captured, and required checks still govern merge readiness. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 1 blocking and 0 non-blocking inline finding(s) publishable; 3 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/hooks/useResolvedAgentModelPicker.ts
NickTitle: 🤖 Fixed in `1c31bd03`. Cross-agent favorite selection now checks `readyAgentIds` in the shared picker-state owner before emitting a combined target intent. Saved favorites remain unchanged. Regression tests cover `not_ready` and `not_installed` agents in the composer, pending drafts, and started sessions, with no target or backend change.

src/features/chat/ui/AgentModelPicker.tsx
NickTitle: 🤖 Fixed in `1c31bd03`. The picker now renders its model list when the supplied inventory contains a live favorite, even if the active agent has no displayed models. A regression test verifies that a foreign favorite remains visible and selectable with an empty active catalog.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `1c31bd03`. Foreign star and unstar controls now include the owning agent in their accessible names, with English and Spanish translations. Active-agent controls keep the shorter label. Regression coverage checks agent-specific labels for same-named models. Validation for this commit: 117 focused tests and `just check` passed. Builderbot returned 0 findings, but posting its status failed with HTTP 403.

src/shared/ui/GlobalComposerPill.tsx
NickTitle: 🤖 Fixed in `02147651`. Global Composer now accepts the favorite’s owning agent and uses that agent/model pair for the provider override, model selection, execution target, selected provider, and model override. The shared readiness guard remains in place. Two new regression cases cover ready foreign favorites with and without an explicit model provider; both fail against the original implementation and pass with this fix. All 119 focused tests and the checks and push hooks passed. The full suite retains three pre-existing GlobalComposerPill test failures, reproduced before this fix. Builderbot returned 0 findings, but posting its commit status failed with HTTP 403.

src/features/chat/hooks/useChatSessionController.ts
```diff
@@ -1418,12 +1420,13 @@ export function useChatSessionController({
         delete pendingDefaultReasoningEffortBySessionRef.current[sessionId];
       }
       useChatStore.getState().resetTokenState(stateSessionId);
-      handleModelChange(modelId, model);
+      handleModelChange(modelId, model, agentId);
```
morgmart: 🤖 **P1 · Reject before resetting chat state** (blocking)

The controller wrapper clears pending default reasoning effort and resets token state before calling the model-selection owner. That downstream owner now rejects cross-agent favorites whose agent is not ready, but the wrapper has already applied these side effects. The selection is blocked while unrelated chat state is still discarded.

**User effect:** Clicking a saved model whose agent has become unavailable can leave the current chat’s model unchanged but silently reset its token information and pending reasoning choice.

**Recommended fix:** Validate and accept the complete agent/model intent before any context-reset side effects. Prefer one selection owner that returns whether the intent was accepted, then clear reasoning and token state only for an accepted transition.

**Test:** Exercise the controller-level callback with a foreign favorite whose agent is not ready and verify the model target, pending reasoning state, and token state all remain unchanged. Also confirm an accepted ready-agent selection still resets them.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -368,45 +645,285 @@ export const RecommendedModelList = forwardRef<
...
+                          >
+                            {providerIcon}
+                          </span>
+                        ) : null}
+                        <div className="flex min-w-0 flex-1 items-baseline gap-1.5 overflow-hidden">
+                          <span className="min-w-0 truncate">
+                            {getModelDisplayName(model)}
+                          </span>
+                          {foreignAgentLabel ? (
+                            <span className="shrink-0 text-xs text-muted-foreground">
```
morgmart: 🤖 **P2 · Protect model names from agent labels** (non-blocking)

The foreign-agent label is secondary metadata but uses shrink-0 with no width or truncation bound. In a narrow picker, a long custom or localized agent name can consume the row’s remaining width while the primary model name shrinks away.

**User effect:** People can see the agent name but lose the model name they need to identify the choice, especially with long names or narrow windows.

**Recommended fix:** Constrain and truncate the secondary agent label while reserving usable space for the model name, using the existing responsive row and typography patterns.

**Test:** Render a narrow picker with a long foreign-agent label and long model name, then verify both remain present and the secondary label truncates without crowding the model name out.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the fresh review of the exact three-dot comparison found one blocking state-consistency issue and one non-blocking responsive-label issue. A readiness-rejected cross-agent favorite still clears reasoning and token state before rejection, and long foreign-agent labels can crowd out the primary model name. Supplied GitHub evidence is structurally valid for the exact head; only three checks were captured, and required checks still govern merge readiness. Favoriting remains accepted product direction, with no escalation.

morgmart (commented):
🤖 Automated code review

No new publishable findings were identified in the fresh review of the exact three-dot comparison. Two still-present issues are suppressed by their existing unresolved threads, which have no substantive human replies; approval therefore fails closed and no publication/retry is recommended. Supplied GitHub evidence is structurally valid for the exact head. The frontend check is failed and still governs merge readiness. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 0 blocking and 0 non-blocking inline finding(s) publishable; 2 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/hooks/useChatSessionController.ts
NickTitle: 🤖 Fixed in `3aa9dcd6`. Model selection now reports synchronous acceptance through the existing selection owners. The controller clears pending default reasoning and resets tokens only after acceptance. Rejected and no-op selections preserve state; backend preparation and rollback retain their existing async behavior. Three controller regression cases use the real picker with `not_ready`, `not_installed`, and `ready` agents. All 203 focused tests, `just check`, and push hooks passed. Builderbot returned 0 findings, but posting its status failed with HTTP 403. The three known PR-related GlobalComposerPill test failures remain outside this fix.

morgmart (commented):
🤖 Automated code review

No new publishable findings were identified in the fresh review of the exact three-dot comparison. One still-present non-blocking issue is suppressed by its existing unresolved thread, which has no substantive human replies; approval therefore fails closed and no publication/retry is recommended. Supplied GitHub evidence is structurally valid for the exact head. The frontend check is failed and still governs merge readiness. Favoriting remains accepted product direction, with no escalation.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -251,36 +469,92 @@ export const RecommendedModelList = forwardRef<
...
+        !liveStarredKeys.has(modelStarKey(getModelScopeId(model), model.id)),
+    );
+    const browsableModels = [...favoriteRows, ...regularRows];
     const normalizedQuery = query.trim().toLowerCase();
     if (!normalizedQuery) {
-      return models;
+      return browsableModels;
     }
-    return models.filter(
+    return browsableModels.filter(
```
morgmart, line 482: 🤖 **P2 · Search visible agent names** (non-blocking)

Cross-agent favorite rows now display the owning agent label, but the search filter checks only model and provider fields. The visible agent label comes from agentLabels and can differ from providerName or providerId, so searching text the row visibly presents can return no match.

**User effect:** A person can see that a favorite belongs to an agent such as “Research,” then be unable to find that favorite by searching for “Research.”

**Recommended fix:** Include the row’s resolved owning-agent display label in search matching for cross-agent favorites, using the same label derivation that renders the row.

**Test:** Render a cross-agent favorite whose visible agent label differs from its provider metadata, search for the visible agent name, and verify the row remains in the results.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found one new non-blocking search-consistency issue. Cross-agent favorite rows visibly show their owning agent, but model search does not match that visible label. Supplied GitHub evidence is structurally valid for the exact head. The frontend check is failed and still governs merge readiness. Favoriting remains accepted product direction, with no escalation.

src/features/chat/ui/AgentModelPicker.tsx
```diff
@@ -373,48 +374,34 @@ export function AgentModelPicker({
     }
   };
 
-  const handleModelSelect = (model: ModelOption) => {
-    recordModelSelection(selectedAgentId, model);
-    onModelChange?.(model.id, model);
+  const handleModelSelect = (model: ModelOption, agentId: string) => {
+    recordModelSelection(agentId, model);
```
morgmart, line 378: 🤖 **P2 · Record only accepted selections** (non-blocking)

The picker records model recency before the downstream selection owner accepts or rejects the agent/model target, and the list resets immediately regardless of acceptance. A cached favorite for an unready agent is rejected by the readiness guard but still becomes recent and closes the current picker flow.

**User effect:** Choosing an unavailable favorite appears to do nothing, yet it changes later model ordering and dismisses the list as though the choice succeeded.

**Recommended fix:** Return the existing acceptance result through AgentModelPicker’s callback, then record recency and reset the list only when the target was accepted.

**Test:** Select a favorite owned by an unready agent and verify recency, ordering, and the open picker view remain unchanged; verify an accepted favorite records recency and resets normally.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -368,48 +645,291 @@ export const RecommendedModelList = forwardRef<
...
+                !favoriteModelKeys || favoriteModelKeys.has(modelKey);
+              const activeStarAnimation = rowAnimation?.state ?? null;
+              const idleStarVisible =
+                starred ||
+                hoveredModelKey === modelKey ||
+                focusedModelKey === modelKey;
+              const handleStarClick = () => {
+                if (rowAnimation) return;
+                const targetStarred = !starred;
+                if (!setStarred(scopeId, model.id, targetStarred)) return;
```
morgmart, line 709: 🤖 **P2 · Restore focus after removing favorites** (non-blocking)

Unstarring a foreign favorite can remove its only row immediately because the model is absent from the selected agent’s catalog. If its star button holds keyboard focus, the focused element unmounts with no explicit focus destination; reduced-motion mode makes this removal immediate.

**User effect:** Keyboard users can lose their place in the picker after removing a foreign favorite and must navigate back into the list from an unclear position.

**Recommended fix:** Before removing the row, move focus to the nearest remaining model row or a stable picker control, preserving list-navigation continuity in both animated and reduced-motion paths.

**Test:** Focus the star action for a foreign-only favorite, unstar it with reduced motion on and off, and verify focus lands on the nearest remaining row or the designated stable control.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found two new non-blocking interaction-consistency issues. Rejected cross-agent choices still alter recency and close/reset the picker, while immediately removing a keyboard-focused foreign favorite can lose focus. Two existing issues are suppressed by unresolved threads. Supplied GitHub evidence is structurally valid for the exact head. The frontend check is failed and still governs merge readiness. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 0 blocking and 2 non-blocking inline finding(s) publishable; 2 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `833bfea1` using the existing RowButton two-line typography pattern within PickerItem. The model name sits above the muted foreign-agent label. Both truncate independently across the full text area; the star remains outside it. Removed the earlier 40% cap.

Verified the production component in Chrome at a 608 × 720 CSS-pixel viewport with long model and agent names. In the narrow expanded harness-plus-reasoning layout, each text line now has 114 px, compared with 62 px for the model under the earlier cap. Both lines share the same left edge, and the star retains a 28 px target. Full accessible names and foreign-agent selection passed. Long names still truncate naturally. This check covers Chrome component rendering, not the full Tauri shell.

All 71 picker tests, checks, and push hooks passed. Builderbot returned 0 findings, but posting its status failed with HTTP 403. The three known PR-related GlobalComposerPill test failures and a separate Goose reasoning-control clipping issue remain outside this fix.

morgmart (commented):
🤖 Automated code review

No new publishable findings were identified in the fresh review of the exact three-dot comparison. Three still-present issues are suppressed by existing unresolved threads with no substantive human replies; approval therefore fails closed and no publication/retry is recommended. Supplied GitHub evidence is structurally valid for the exact head and all captured checks are successful. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 0 blocking and 0 non-blocking inline finding(s) publishable; 3 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPickerLists.tsx
```diff
@@ -368,48 +658,312 @@ export const RecommendedModelList = forwardRef<
...
+                            </span>
+                          ) : null}
+                        </div>
+                      </div>
+                    </PickerItem>
+                    {existsInCatalog ? (
+                      <Button
+                        variant="ghost"
+                        size="icon-xs"
+                        selected={starred}
```
morgmart, line 884: 🤖 **P2 · Navigate from the focused star row** (non-blocking)

Star buttons are keyboard-focusable, but the picker’s Up and Down handler indexes only model-selection buttons marked as navigation items. When focus is on a star, its index is always missing, so ArrowDown jumps to the first model and ArrowUp jumps to the last rather than moving relative to the star’s owning row.

**User effect:** Keyboard users who favorite or unfavorite a model can press an arrow and unexpectedly jump across the list instead of continuing to the adjacent model.

**Recommended fix:** When focus is inside a model row, resolve that row’s model-selection button as the navigation origin before calculating the next item. Keep star actions outside the navigation item set while preserving their row context.

**Test:** Focus a star action in a middle model row, press ArrowDown and ArrowUp, and verify focus moves to the next and previous model rows rather than the first and last rows.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found one new non-blocking keyboard-navigation issue. Arrow navigation from a focused star action loses the row context and jumps to an edge of the model list. Supplied GitHub evidence is structurally valid for the exact head; two checks remain in progress, so required checks still govern merge readiness. Favoriting remains accepted product direction, with no escalation.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `c5108ab4`. Search and rendering now share the resolved foreign-agent display label. A regression uses the visible label “Research” with different provider metadata and verifies the favorite remains in search results.

src/features/chat/ui/AgentModelPicker.tsx
NickTitle: 🤖 Fixed in `c5108ab4`. The controller and picker callbacks now preserve the selection acceptance result. Rejected selections do not record recency or reset the picker view. A regression uses the real readiness hook to reject an unavailable foreign favorite, then accept it when ready. Existing void callbacks remain compatible.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Fixed in `c5108ab4`. Removing a focused foreign favorite moves focus to an adjacent model row, with search or the stable picker dialog as fallbacks. Regression tests cover normal motion, reduced motion, and removal of the last row. Four Chrome checks with production picker components passed across both motion modes, with and without a remaining row.

Validation for this commit: 7,770 tests passed, one skipped; checks and commit/push hooks passed. CI and the clean-room install completed successfully: https://github.com/block/berd/actions/runs/34615598380 . Builderbot computed zero review comments, but its status-post step failed with HTTP 403.

src/features/chat/ui/AgentModelPicker.tsx
```diff
@@ -572,7 +553,14 @@ export function AgentModelPicker({
                 "button[data-picker-nav-item]:not(:disabled)",
               ),
             );
-            const idx = items.indexOf(document.activeElement as HTMLElement);
+            const activeElement = document.activeElement as HTMLElement;
```
morgmart: 🤖 **P2 · Preserve star row across columns** (non-blocking)

The new vertical arrow handler resolves a focused star action to its owning model-selection button before finding the row index. The Left and Right handler still indexes document.activeElement directly against navigation-item buttons, so a focused star has no index and falls back to the first item in the adjacent column.

**User effect:** Keyboard users moving sideways from a star in a middle model row unexpectedly land at the top of the neighboring column instead of staying on the corresponding row.

**Recommended fix:** Reuse the focused star’s owning model-row navigation item as the horizontal navigation origin before calculating currentIdx, matching the vertical path.

**Test:** Focus a star action in a middle model row, press ArrowLeft and ArrowRight with adjacent columns visible, and verify focus lands on the corresponding row rather than the first item.

morgmart (commented):
🤖 Automated code review

COMMENT: the fresh review of the exact three-dot comparison found one new non-blocking keyboard-navigation issue. Vertical movement now resolves a focused star to its model row, but horizontal movement still loses that row position and jumps to the first item in the adjacent column. Supplied GitHub evidence is structurally valid for the exact head; several checks remain in progress, so required checks still govern merge readiness. Favoriting remains accepted product direction, with no escalation.

morgmart (commented):
🤖 Automated code review

No new publishable findings were identified in the fresh review of the exact three-dot comparison. Two prior keyboard-navigation issues are suppressed by existing unresolved threads with no substantive human replies; approval therefore fails closed and no publication/retry is recommended. Supplied GitHub evidence is structurally valid for the exact head. The frontend check is failed and still governs merge readiness. Favoriting remains accepted product direction, with no escalation.

Deterministic publication result: 0 blocking and 0 non-blocking inline finding(s) publishable; 2 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/chat/ui/AgentModelPicker.tsx
NickTitle: 🤖 Addressed in `cea884f`. The same row-resolved navigation origin now drives Arrow Left and Arrow Right, preserving the focused model row across picker columns. Added three-column regression coverage for both directions.

src/features/chat/ui/AgentModelPickerLists.tsx
NickTitle: 🤖 Addressed in `cea884f`. The picker now resolves a focused star button to its owning model row before calculating Arrow Up and Arrow Down destinations. Added regression coverage for both directions from a middle-row star.

morgmart (approved):
🤖 Automated code review

APPROVE: the fresh review of the exact three-dot comparison found no new publishable findings and no unresolved duplicate suppressions. Favoriting is accepted product direction and the implementation is coherent across persistence, cross-agent selection, responsive UI, accessibility, localization, lifecycle, and tests. Supplied GitHub evidence is structurally valid for the exact head and all captured checks are successful. No escalation is needed.

PR #289, opened 2026-09-03
feat(memory): require review for memory proposals

delkc:
## Summary

Require an explicit review decision before an agent suggestion becomes durable or recallable memory.

- Keep agent suggestions in a local pending queue until the person approves them
- Provide editable approval and decline controls in Settings → Memory
- Show Review and Decline notifications, including session-safe handling for MCP proposals
- Use salted, content-free suppression fingerprints for declines and unambiguous manual deletions
- Serialize and idempotently apply approval in Rust, with credential validation and scoped destination routing

### Related issue

N/A. Depends on #288.

### Testing

- Proposal parsing, review, suppression, and notification tests
- Backend approval retry, credential rejection, and queue tests
- TypeScript typecheck, i18n, Biome, and design-system checks

Generated with Goose


shellz-n-stuff (commented):
This needs to strip zero-width chars and other such tricks that hide content from users whilst still being there.

If the security control is review it's vital that everything is stripped to human readable text

PR #292, opened 2026-09-03
fix(voice): defer OpenAI Keychain access

johnmatthewtennant:
## Summary

- Read the shared OpenAI voice credential only when the selected Voice settings need it, while keeping OpenAI TTS selectable before that read.
- Coalesce concurrent mount-time status requests and force fresh status reads after credential changes.
- Use only Berd’s shared `berd-openai-voice` / `api-key` credential, without legacy account lookup or migration.

- Open Voice settings with the default on-device chained configuration. Berd does not inspect the OpenAI Keychain credential.
- Select OpenAI transcription, OpenAI playback, or OpenAI Realtime. The corresponding settings load the shared credential status.
- Save or clear the shared key. Every mounted OpenAI voice surface refreshes to the new status, while simultaneous mount reads share one in-flight request.

### DIFF
--- a/src/features/voice-conversation/hooks/useOpenAiVoiceSetup.ts
+++ b/src/features/voice-conversation/hooks/useOpenAiVoiceSetup.ts
@@ -10,7 +10,11 @@ export function useOpenAiVoiceSetup(enabled = true) {
   const [error, setError] = useState<string | null>(null);
 
   useEffect(() => {
-    if (!enabled) return;
+    if (!enabled) {
+      setStatus(null);
+      setError(null);
+      return;
+    }
     let active = true;
     let refreshGeneration = 0;
     let unsubscribe: (() => void) | null = null;
--- a/src/features/voice-conversation/hooks/useOpenAiVoiceSetup.ts
+++ b/src/features/voice-conversation/hooks/useOpenAiVoiceSetup.ts
@@ -16,7 +16,7 @@ export function useOpenAiVoiceSetup(enabled = true) {
     let unsubscribe: (() => void) | null = null;
     const refresh = () => {
       const generation = ++refreshGeneration;
-      void getOpenAiVoiceStatus().then(
+      void getOpenAiVoiceStatus({ coalesce: true }).then(
         (next) => {
           if (active && generation === refreshGeneration) {
             setStatus(next);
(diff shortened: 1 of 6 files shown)


src/features/voice-conversation/hooks/useOpenAiVoiceSetup.ts
```diff
@@ -10,13 +10,17 @@ export function useOpenAiVoiceSetup(enabled = true) {
...
+    if (!enabled) {
+      setStatus(null);
+      setError(null);
+      return;
+    }
     let active = true;
     let refreshGeneration = 0;
     let unsubscribe: (() => void) | null = null;
-    const refresh = () => {
+    const refresh = (coalesce = false) => {
```
morgmart, line 21: 🤖 **P1 · Credential changes can reuse stale status** (blocking)

The Tauri event listener invokes its callback with an event object. Because `refresh` accepts its first argument as the `coalesce` flag and is passed directly to `listenToOpenAiVoiceSettings`, a settings-changed event makes that flag truthy. If the mount-time status read is still pending, the post-write refresh can therefore join that older request instead of starting a fresh read. The test invokes the captured callback without an event argument, so it does not reproduce the production contract.

**User effect:** After saving or removing an OpenAI voice key, people can still be told that the old key state is active and may be prompted to save or remove it again until another refresh happens.

**Recommended fix:** Pass a zero-argument wrapper to the event listener that starts an uncoalesced refresh, and reserve coalescing for mount-time reads.

**Test:** Dispatch the captured settings listener with a representative Tauri event object while an older status request is unresolved; assert that a second fresh request starts and that only its result becomes visible.

morgmart (changes requested):
🤖 Automated code review

Engineering found one blocking credential-refresh race. The supplied GitHub evidence is structurally valid: Semgrep, zizmor, DCO, clean-room install, and Linux Rust checks passed, while frontend and macOS/Windows Rust checks were in progress when captured; required checks still govern merge readiness.

src/features/voice-conversation/hooks/useOpenAiVoiceSetup.ts
johnmatthewtennant: 🤖 Accepted and fixed in 0f876f70. The listener now uses a zero-argument wrapper so Tauri event payloads cannot enable coalescing, and the regression test dispatches a representative event while the mount read is unresolved and verifies that the fresh result wins.

morgmart (approved):
🤖 Automated code review

Engineering completed a fresh static review of the full exact PR comparison and found no publishable issues. The previously reported credential-refresh race is fixed at this head and its resolved thread has a substantive human reply. Supplied GitHub evidence is structurally valid, and all captured check runs completed successfully; required checks still govern merge readiness.

PR #293, opened 2026-09-03
Add access controls to bb apps

nathan-thillairajah:
## Summary

`bb apps` can manage the app lifecycle, but it cannot inspect or update who may view a deployed app. This adds:

- `bb apps access get <APP_ID> [--environment <ENVIRONMENT>]`
- `bb apps access set <APP_ID> --visibility organization|restricted [--viewer <IDENTITY>...] [--clear-viewers] [--environment <ENVIRONMENT>]`

These commands map to the control plane GET and PUT access endpoints through the existing allowlisted client. `set` replaces the full policy. Restricted policies require at least one `--viewer` or an explicit `--clear-viewers` confirmation, while the owner and approved publishers remain effective viewers. Viewer help points to the exact case-sensitive `caller` subject returned by `bb apps list --json`. The control plane remains authoritative for identity validation and owner-only mutation authorization, and responses are preserved unchanged.

### Related issue

N/A. No matching open issue or PR found.

### Testing

At commit `b89c2ebe`:

- `cd bb-cli && source ../bin/activate-hermit && just check` (153 unit, 81 bb E2E, 22 CLI E2E, 6 acceptance tests)
- repository pre-commit and pre-push hooks (root frontend checks, Tauri checks, formatting, and Clippy)

Live access mutation was not run because the local Builderlab CLI session remains intentionally logged out. The request and response contract was verified against the live control-plane contract and current Compose source, then covered with mocked client and process tests.

### DIFF
--- a/bb-cli/src/bb/apps.rs
+++ b/bb-cli/src/bb/apps.rs
@@ -366,9 +366,10 @@ pub fn command() -> Command {
                         .about("Replace an app's visibility and explicit viewer list")
                         .long_about(
                             "Replace an app's complete access policy. For restricted visibility, \
-                             repeat --viewer for each explicit viewer. Omitting --viewer clears the \
-                             explicit viewer list; the owner and approved publishers remain effective \
-                             viewers. Only the original owner may update access.",
+                             repeat --viewer for each explicit viewer, or pass --clear-viewers to \
+                             explicitly clear the list. The owner and approved publishers remain \
+                             effective viewers. Only the original owner may update access. Ask each \
+                             intended viewer to copy the exact caller value from `bb apps list --json`.",
                         )
                         .arg(
                             Arg::new("app-id")
@@ -389,7 +390,20 @@ pub fn command() -> Command {
                                 .long("viewer")
                                 .value_name("IDENTITY")
                                 .action(clap::ArgAction::Append)
-                                .help("Explicit viewer identity; repeat for each viewer"),
+                                .help(
+                                    "Exact case-sensitive Apps Platform user subject (for example, \
+                                     auth0|...); ask the viewer to copy `caller` from `bb apps list \
+                                     --json`; repeat for each viewer",
+                                ),
+                        )
+                        .arg(
+                            Arg::new("clear-viewers")
+                                .long("clear-viewers")
+                                .action(clap::ArgAction::SetTrue)
+                                .conflicts_with("viewer")
+                                .help(
+                                    "Confirm replacing the explicit viewer list with an empty list",
+                                ),
                         )
                         .arg(
                             Arg::new("environment")
@@ -664,12 +678,17 @@ fn run_access_set(config: &SkillsConfig, matches: &ArgMatches) -> Result<()> {
     let visibility = matches
         .get_one::<String>("visibility")
         .context("expected access visibility")?;
-    let viewers = matches
+    let viewers: Vec<&str> = matches
         .get_many::<String>("viewer")
         .into_iter()
         .flatten()
         .map(String::as_str)
         .collect();
+    if visibility == "restricted" && viewers.is_empty() && !matches.get_flag("clear-viewers") {
+        anyhow::bail!(
+            "restricted visibility requires at least one --viewer or explicit --clear-viewers confirmation"
+        );
+    }
     let request = AccessRequest {
         visibility,
         viewers,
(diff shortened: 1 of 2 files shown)


bb-cli/src/bb/apps.rs
```diff
@@ -334,6 +334,71 @@ pub fn command() -> Command {
...
+                                .value_name("ENVIRONMENT")
+                                .help("Optional Compose environment override"),
+                        ),
+                ))
+                .subcommand(control_plane_args(
+                    Command::new("set")
+                        .about("Replace an app's visibility and explicit viewer list")
+                        .long_about(
+                            "Replace an app's complete access policy. For restricted visibility, \
+                             repeat --viewer for each explicit viewer. Omitting --viewer clears the \
```
morgmart: 🤖 **P1 · Require explicit viewer clearing** (blocking)

The new command defines omission of every --viewer argument as replacing a restricted policy's explicit viewer list with an empty list. Because omission is indistinguishable from forgetting or dropping the repeated arguments, a full access revocation can pass argument validation without any affirmative clearing intent. Help text documents the behavior, but the command does not guard the destructive case before authentication and network dispatch.

**User effect:** An owner who forgets the viewer arguments can unintentionally remove access for every explicitly invited person while believing they only changed the app's visibility setting.

**Recommended fix:** For restricted visibility, require either at least one --viewer or a separately named --clear-viewers confirmation, make those paths unambiguous, and reject the omitted case before reading credentials or making a request.

**Test:** Add a discriminating command test proving that restricted visibility with neither --viewer nor --clear-viewers fails before authentication or network activity, and that the explicit clearing path serializes an empty viewer list.

bb-cli/src/bb/apps.rs
```diff
@@ -334,6 +334,71 @@ pub fn command() -> Command {
...
+                                .value_parser(["organization", "restricted"])
+                                .required(true)
+                                .help("Who may view the app"),
+                        )
+                        .arg(
+                            Arg::new("viewer")
+                                .long("viewer")
+                                .value_name("IDENTITY")
+                                .action(clap::ArgAction::Append)
+                                .help("Explicit viewer identity; repeat for each viewer"),
```
morgmart: 🤖 **P2 · Explain viewer identity discovery** (non-blocking)

The new --viewer option is described only as an explicit viewer identity. The implementation and tests use opaque values such as auth0|bob, but the CLI does not state the accepted human-facing identifier format or identify a command or system where an owner can obtain it. Server-side validation can reject bad values, but it does not make the successful input discoverable before submission.

**User effect:** An owner can find the access command yet still be unable to invite a new person without undocumented, out-of-band knowledge of that person's identity value.

**Recommended fix:** Update the option and command help to name the accepted identifier format and where owners can obtain it; if only opaque internal IDs are accepted, connect the workflow to an identity lookup or listing surface rather than leaving ID discovery implicit.

**Test:** Extend the help test to pin the accepted identifier guidance and its discovery path, while keeping authoritative identity validation at the control-plane boundary.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static review of the exact three-dot comparison c5058ae95a5baee7dd6645f69a6f1bf1fe94fcb7...532017e9b265dac5e4cf2cfc25eb38a22e19eece found one blocking consent risk and one non-blocking discoverability gap in the new access-policy command: omitting every viewer silently clears restricted access, and the CLI does not explain which viewer identity to supply or where to find it. The project code-review and user wes-review skills were explicitly loaded and applied as distinct Engineering passes. Final self-check covered the access help, get, and full-policy replacement flows; accessibility; i18n/localization completeness; navigation and consent guards; bounded network waits, error and never-completes behavior, stateless lifecycle and race behavior; test honesty; design-system and project rules; duplicate and overlapping findings; and concrete evidence plus user effect for blocking severity. The supplied GitHub evidence was inspected: seven reported checks completed successfully, while Tauri Rust checks (macOS) and Frontend checks and unit tests were in progress; required checks independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 addressed in b89c2ebe. Restricted visibility now rejects an omitted viewer list before credential lookup or network dispatch unless --clear-viewers is explicitly present; --clear-viewers also conflicts with --viewer. Added E2E coverage for the early rejection and a process test proving the confirmed path serializes viewers: [].

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 addressed in b89c2ebe. The option help now identifies the value as the exact case-sensitive Apps Platform user subject and tells the intended viewer to copy caller from bb apps list --json, with auth0|... as an example. The help E2E test pins that discovery path while the control plane remains authoritative for validation.

morgmart (commented):
🤖 Automated code review

COMMENT. A fresh static review of the full exact three-dot comparison c5058ae95a5baee7dd6645f69a6f1bf1fe94fcb7...b89c2ebe7a43420882fc1256cf09751eac7d24a2 found no publishable findings. The prior restricted-viewer clearing risk is fixed by affirmative --clear-viewers consent plus pre-credential rejection of accidental omission, and viewer identity discovery is now connected to the exact caller subject from bb apps list --json; both fixes have discriminating coverage. The project code-review and user wes-review skills were explicitly loaded and applied as distinct Engineering passes. Final self-check covered access discovery, get, organization and restricted replacement, explicit clearing, invalid and conflicting arguments, environment selection, authorization and control-plane errors, bounded waits and never-completes behavior, retry/idempotence, lifecycle and concurrency, accessibility, i18n/localization, navigation and consent guards, test honesty, design-system and project rules, duplicate overlap, and concrete evidence plus user effect for blocking severity. No unsupported implementation finding survived. Atlas retains one product-policy escalation about the original-owner and approved-publisher access model, so the recommendation is COMMENT rather than APPROVE. All ten supplied GitHub check runs completed successfully; the supplied combined commit status remains pending with no status contexts, so required checks still independently govern merge readiness.

morgmart (approved):

PR #297, opened 2026-09-04
Require explicit Apps Platform identity in bb

nathan-thillairajah:
## Summary

Apps Platform owns the final app identity and the decision to initialize it. `bb apps create` needs to preserve both decisions without accepting a substitute ID, mutating from a partial plan, or leaving an interrupted create permanently stranded.

Create now requires a non-empty `--app-id` or `--name`, plans the identity, and reserves the exact planned ID before optional initialization. It rejects an ID substitution returned by plan, reserve, or initialize. Before reserving anything, it also requires both `initialize.required` and `initialize.recommended` to be present booleans; a partial or malformed decision fails before mutation.

Every valid plan reaches the reservation endpoint, including static and other reserve-only plans. The CLI forwards the plan settings to reserve and calls initialize only when the authoritative plan requests it, so recovery cannot turn a reserve-only plan into an initialized app.

Interrupted creates remain resumable without weakening the server create boundary. If reserve fails or returns an unreadable response, the CLI performs an authenticated lookup for the exact app. It resumes only when the response explicitly proves a matching caller-owned, revision-1 idle reservation with the same name, environment, persistence, and runtime class, no active or deleted state, and a present, well-typed, empty versions array. Other-owner, publisher-only, malformed, mismatched, deleted, or established responses retain the original collision failure.

For dynamic apps, the companion Compose change keeps initialization required in the plan until the server records the runner variant. That server-authoritative marker lets a retry resume initialization after a partial failure without causing an intentional reserve-only app to initialize.

If the reservation result and follow-up inspection are both inconclusive, the CLI returns `reservation_outcome_unknown` with the exact `bb apps get` command to inspect the app. Initialization failures explain that the ID is already reserved and that repeating the same create command will reconcile and continue.

[squareup/blockplat-compose#259](https://github.com/squareup/blockplat-compose/pull/259) supplies the authoritative create-only endpoint, collision semantics, and initialization-state plan behavior. Deploy that server change first, then include this client change in a subsequent `bb` release.

### Related issue

N/A.

### Testing

- `cd bb-cli && source ../bin/activate-hermit && just check` (166 unit tests, 81 `bb` end-to-end tests, 22 CLI end-to-end tests, and 6 acceptance-contract tests)
- Repository pre-commit hook
- Repository pre-push hooks: `fmt-check`, `tauri-check`, `clippy`, and `check`
- Create coverage includes partial and malformed initialization decisions failing before reserve, app-ID substitutions at every response stage, initialization failure followed by a server-plan-driven retry, reserve-only reconciliation without initialization, unreadable committed reservation responses, fail-closed version evidence, unresolved ambiguous outcomes, and non-owner collisions

### DIFF
--- a/bb-cli/src/bb/apps.rs
+++ b/bb-cli/src/bb/apps.rs
@@ -35,7 +35,7 @@ use super::display::{print_json, terminal_safe_text, Style};
 use super::runner;
 #[cfg(test)]
 use super::skills_api::failure_info;
-use super::skills_api::{exit_codes, failure};
+use super::skills_api::{exit_codes, failure, CliFailure};
 use super::skills_config::SkillsConfig;
 
 const APPS_BASE_URL_ENV_VAR: &str = "BB_APPS_CONTROL_PLANE_URL";
@@ -137,7 +137,9 @@ pub fn command() -> Command {
                     "Plan an app identity through Apps Platform, reserve that exact ID, then \
                      initialize it only when the returned plan marks initialization as required \
                      or recommended. Supply a descriptive DNS-safe app ID or a human-readable \
-                     name from which the control plane can derive one.",
+                     name from which the control plane can derive one. If creation is interrupted \
+                     after reservation, repeat the same command to reconcile the caller-owned \
+                     idle reservation and continue initialization.",
                 )
                 .group(
                     ArgGroup::new("app-identity")
@@ -543,15 +545,22 @@ fn run_create(config: &SkillsConfig, matches: &ArgMatches) -> Result<()> {
             "Apps Platform plan response did not include initialize.required or initialize.recommended"
         );
     }
-    let should_initialize =
+    let plan_requests_initialize =
         initialize_required.unwrap_or(false) || initialize_recommended.unwrap_or(false);
     let mutation_request = mutation_request_from_plan(&plan);
-    let reservation = client.reserve(&credential, &app_id, &mutation_request)?;
-    let reserved_app_id =
-        required_response_string(&reservation, "app_id", "Apps Platform reserve")?;
-    require_exact_app_id("reserve", &app_id, reserved_app_id)?;
+    let (reservation, reservation_reconciled) =
+        reconcile_create_reservation(&client, &credential, &app_id, &plan, &mutation_request)?;
+    let should_initialize = plan_requests_initialize
+        || (reservation_reconciled
+            && plan.get("runtime_profile").and_then(Value::as_str) != Some("static"));
     let initialize = if should_initialize {
-        let response = client.initialize(&credential, &app_id, &mutation_request)?;
+        let response = client
+            .initialize(&credential, &app_id, &mutation_request)
+            .with_context(|| {
+                format!(
+                    "Apps Platform reserved app_id {app_id:?}, but initialization did not complete. Retry the same `bb apps create` command to reconcile the reservation and continue initialization."
+                )
+            })?;
         let initialized_app_id =
             required_response_string(&response, "app_id", "Apps Platform initialize")?;
         require_exact_app_id("initialize", &app_id, initialized_app_id)?;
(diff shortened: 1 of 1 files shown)


bb-cli/src/bb/apps.rs
```diff
@@ -531,35 +545,51 @@ fn run_create(config: &SkillsConfig, matches: &ArgMatches) -> Result<()> {
     }
     let should_initialize =
         initialize_required.unwrap_or(false) || initialize_recommended.unwrap_or(false);
+    let mutation_request = mutation_request_from_plan(&plan);
+    let reservation = client.reserve(&credential, &app_id, &mutation_request)?;
```
morgmart: 🤖 **P1 · Create cannot resume partial success** (blocking)

The command reserves the planned app ID before optional initialization, but it only emits the successful result after initialization completes. If initialization fails, or if the reservation commits but its response is lost, the command exits after the durable reservation. Repeating the same create flow calls reserve again, while this change's collision behavior stops on an existing reservation, and the CLI exposes no separate initialize or resume path.

**User effect:** Someone can see app creation fail even though their requested ID has already been reserved, then have every retry blocked by a collision while the app remains unusable and uninitialized.

**Recommended fix:** Make the operation resumable through one server-authoritative mechanism: have an identical request idempotently return the caller's matching reservation and continue initialization, use a supported idempotency key, or expose a supported resume/initialize operation. For an ambiguous post-mutation transport failure, report that the outcome is unknown and provide a reliable reconciliation path rather than treating it as an ordinary failed create.

**Test:** Cover reserve success followed by initialize failure, then repeat the identical create request and assert that it resumes the same app and completes initialization. Also simulate reserve committing while its response is dropped and assert that retry reconciles the existing reservation instead of ending in an unrecoverable collision.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the new create sequence can durably reserve an app and then fail without a supported way to resume, leaving the requested identity stranded. The supplied GitHub evidence is structurally available and all listed check runs completed successfully; required checks still independently govern merge readiness.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in de8ef04e. `bb apps create` now reconciles a failed or ambiguous reserve through an authenticated exact-app lookup and resumes only a caller-owned, revision-1 idle reservation whose name, environment, persistence, and runtime class match and which has no published versions. Non-static recovery runs the idempotent initialize operation even when the retry plan no longer recommends it. Initialization failures now tell the user to retry the same create command; if both reserve and inspection are inconclusive, the CLI returns `reservation_outcome_unknown` with the exact `bb apps get` command. Added process tests for initialize failure followed by an identical successful retry, an unreadable committed reservation response, unresolved ambiguity, and non-owner collisions. `cd bb-cli && source ../bin/activate-hermit && just check` plus the repository pre-commit and pre-push hooks pass.

bb-cli/src/bb/apps.rs
```diff
@@ -529,37 +545,189 @@ fn run_create(config: &SkillsConfig, matches: &ArgMatches) -> Result<()> {
             "Apps Platform plan response did not include initialize.required or initialize.recommended"
         );
     }
-    let should_initialize =
+    let plan_requests_initialize =
         initialize_required.unwrap_or(false) || initialize_recommended.unwrap_or(false);
+    let mutation_request = mutation_request_from_plan(&plan);
+    let (reservation, reservation_reconciled) =
+        reconcile_create_reservation(&client, &credential, &app_id, &plan, &mutation_request)?;
+    let should_initialize = plan_requests_initialize
```
morgmart: 🤖 **P1 · Recovery overrides reserve-only plans** (blocking)

For any reconciled reservation, the CLI initializes whenever runtime_profile is not exactly static, even when the current authoritative plan explicitly sets both initialize.required and initialize.recommended to false. A non-static reserve-only create whose reservation response is lost can therefore be reconciled and initialized solely from its runtime profile, contradicting the plan and the supported artifact-only reserve flow.

**User effect:** Repeating or recovering a reserve-only app creation can unexpectedly initialize resources that Apps Platform explicitly said should remain uninitialized.

**Recommended fix:** Obey the plan's initialize decision during reconciliation. If an interrupted initialization must resume after a later plan returns false, obtain a server-authoritative marker that initialization was previously requested or started rather than inferring that state from a non-static runtime profile.

**Test:** Simulate a non-static plan with initialize.required and initialize.recommended both false, a committed reservation with an unreadable response, and successful inspection. Assert reconciliation succeeds without calling initialize. Separately cover resumed initialization using an authoritative server state that distinguishes it from an intentional reserve-only app.

bb-cli/src/bb/apps.rs
```diff
@@ -529,37 +545,189 @@ fn run_create(config: &SkillsConfig, matches: &ArgMatches) -> Result<()> {
...
+        return false;
+    };
+    if app.get("app_id").and_then(Value::as_str) != Some(app_id)
+        || app.get("role").and_then(Value::as_str) != Some("owner")
+        || app.get("route_status").and_then(Value::as_str) != Some("idle")
+        || app.get("route_revision").and_then(Value::as_u64) != Some(1)
+        || value_has_text(app.get("active_version_id"))
+        || value_has_text(app.get("version_id"))
+        || value_has_text(app.get("deleted_at"))
+        || response
```
morgmart, line 657: 🤖 **P1 · Recovery accepts missing version evidence** (blocking)

The incomplete-reservation predicate rejects versions only when the response field is an array containing entries. If versions is absent, null, or malformed, the condition evaluates as though an authoritative empty array was returned, allowing recovery to continue despite the stated invariant that the app has no published versions.

**User effect:** A retry can mistake an established app for this interrupted reservation and initialize or otherwise alter it instead of preserving the collision.

**Recommended fix:** Fail closed by requiring versions to be present, be an array, and be empty before treating the app as a recoverable incomplete reservation. Apply the same fail-closed rule to every response field used as affirmative proof of safe recovery.

**Test:** Return an otherwise matching revision-1 idle app with versions omitted, null or non-array, and non-empty. Assert every case preserves the original collision and never calls initialize; retain an explicit empty-array case that does reconcile.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: recovery can initialize an intentionally reserve-only non-static app against the current plan, and its reservation predicate treats missing or malformed version evidence as proof that no versions exist. The prior partial-success concern was reassessed against the new implementation; one remaining instance is suppressed to its existing unresolved thread. Supplied GitHub checks are valid but one was still in progress, so required checks continue to govern merge readiness independently.

Deterministic publication result: 2 blocking and 0 non-blocking inline finding(s) publishable; 1 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 8303bcfd. Reconciliation no longer infers initialization from a non-static runtime profile; it calls initialize only when the current server plan sets `initialize.required` or `initialize.recommended`. The companion Compose change makes the missing server-recorded runner variant keep that plan decision true after an interrupted dynamic initialization. Added a process test for an unreadable committed reservation under a non-static reserve-only plan and asserted successful reconciliation with no initialize request, while the interrupted-initialize retry test is driven by the authoritative plan.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 8303bcfd. Recovery now requires `ok: true`, an app object with exact typed identity/ownership/idle-revision fields, and `versions` to be present, an array, and empty. Missing, null, malformed, or non-empty versions all fail closed; malformed active-version, version, or deletion fields also fail closed. Added explicit coverage for every versions shape plus malformed empty-state fields. The full `bb-cli` suite passes.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Follow-up in 8303bcfd refines the recovery added in de8ef04e: initialization now always follows the current authoritative plan. The companion Compose change keeps initialization required while a reserved dynamic route lacks its server-recorded runner variant, so a failed initialization still resumes, while intentional reserve-only plans remain reserve-only. Recovery also now requires an explicit empty versions array and well-typed empty state. The full `bb-cli` check suite passes.

bb-cli/src/bb/apps.rs
```diff
@@ -529,37 +545,189 @@ fn run_create(config: &SkillsConfig, matches: &ArgMatches) -> Result<()> {
             "Apps Platform plan response did not include initialize.required or initialize.recommended"
         );
     }
-    let should_initialize =
+    let plan_requests_initialize =
         initialize_required.unwrap_or(false) || initialize_recommended.unwrap_or(false);
-    let initialize = if should_initialize {
-        let request = initialize_request_from_plan(&plan);
-        Some(client.initialize(&credential, &app_id, &request)?)
+    let mutation_request = mutation_request_from_plan(&plan);
```
morgmart, line 550: 🤖 **P1 · Require the complete initialize decision** (blocking)

The response is rejected only when both initialize.required and initialize.recommended are absent or non-boolean. If either field is missing or malformed, the other is accepted and the missing value is converted to false. The command then proceeds to the newly durable reservation even though the control plane did not provide its complete initialization decision.

**User effect:** A malformed plan can reserve an app and report reserve-only success even though Apps Platform never gave a complete answer about whether that app must be initialized.

**Recommended fix:** Require both initialize.required and initialize.recommended to be present booleans before making the reserve request. Fail before mutation when either field is missing or has the wrong type.

**Test:** Cover each initialize field independently as missing, null, and non-boolean while the other field is valid. Assert every case fails after plan and never sends reserve or initialize.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES: the create flow accepts a partial or malformed initialization decision and can durably reserve an app while silently treating the missing decision as false. The prior recovery-provenance concern remains protected by a resolved thread with substantive human replies and is suppressed. Supplied GitHub evidence is structurally valid and all listed checks completed successfully; required checks still independently govern merge readiness.

Deterministic publication result: 1 blocking and 0 non-blocking inline finding(s) publishable; 1 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

bb-cli/src/bb/apps.rs
nathan-thillairajah: 🤖 Addressed in 2e5d5317. `bb apps create` now requires both `initialize.required` and `initialize.recommended` to be present booleans before constructing or sending the reserve request; either missing, null, or non-boolean field fails immediately after plan. The process-test matrix covers all six independent malformed shapes and asserts that plan is the only control-plane request. `cd bb-cli && source ../bin/activate-hermit && just check` passes with 166 unit, 81 bb end-to-end, 22 CLI end-to-end, and 6 acceptance tests.

morgmart (approved):
🤖 Automated code review

APPROVE: the exact-ID create, reserve, initialization, and recovery flows are coherent at this SHA, and the prior blocking findings are addressed. One remaining recovery-provenance concern is protected by a resolved thread with substantive human replies and is suppressed. Supplied GitHub evidence is structurally valid and all listed checks completed successfully; the companion control-plane change must still be deployed first as documented, and required checks independently govern merge readiness.

PR #301, opened 2026-09-06
feat(voice): centralize Expert Spokesperson runtime in berd-voice

johnmatthewtennant:
## Summary

Centralizes OpenAI Realtime Expert and Spokesperson behavior in `berd-voice`. Native Berd uses the in-process managed host, while external clients such as Voice Conversation CLI use the framed session and remote PCM adapter. Both paths share the same prompt, causal delivery core, handoff lifecycle and identity derivation, transcript and interruption reducer, response coordinator, revisioned voice/rate transaction and queue, quiescence policy, and renewal/recovery decisions. A single `RealtimeHostLifecycle` owns activity tracking, renewal admission and timing, maintenance request identity, queued-settings readiness, update safety, and disconnect recovery for both hosts. The remaining host loops are transport adapters for native playback versus framed PCM acknowledgements.

The runtime emits typed Expert delivery events. Presentation stays at the client boundary, so native Berd renders conversational context and Voice Conversation CLI can render its TSV stream without putting a display format into the shared runtime. The Berd Voice management API also owns the supported OpenAI voice catalog. The renamed `live_event` contract and both framed audio directions use wire marker 3.

OpenAI server VAD owns turn detection and interruption. Berd Voice tracks host-confirmed playback frames for conservative interrupted-transcript estimates and provider context truncation. Graceful shutdown drains queued microphone audio, appends enough silence for server VAD to finalize active input, and waits for provider lifecycle completion before releasing shared protocol state. Both hosts renew idle Realtime sessions before expiry and recover quiescent disconnected sessions through the same replacement transaction while preserving buffered input.

Paired VCCLI adapter: https://github.com/squareup/voice-conversation-cli/pull/118

1. Start an OpenAI Realtime voice conversation in Berd. A routine greeting stays with the Spokesperson. A question about repository implementation or session lifecycle hands off to the Expert without a speculative answer first.
2. While a Spokesperson response is active, change its voice or playback rate. The change waits for a safe boundary, preserves the conversation transcript, and reports the new authoritative revision.
3. Interrupt a long Spokesperson response. Playback stops immediately, the provider conversation is truncated to the delivered audio duration, and the Expert receives an interrupted best-effort transcript rather than the unheard suffix.
4. Run Voice Conversation CLI against this `berd-voice` binary in Expert and Spokesperson mode. The same causal and handoff behavior is delivered as structured events through the framed session, with VCCLI responsible for TSV rendering and host PCM acknowledgements.
5. Set `BERD_VOICE_REALTIME_RENEW_AFTER_MS` to a short positive interval and leave either host quiescent. It replaces the Realtime session without changing the public voice-settings revision; a quiescent provider disconnect similarly reconnects and forwards buffered microphone input once.


src-tauri/src/commands/openai_realtime.rs
```diff
@@ -60,20 +97,483 @@ pub async fn create_openai_realtime_session() -> Result<OpenAiRealtimeSession, S
...
+    Ok(())
+}
+
+#[cfg(not(target_os = "macos"))]
+fn ensure_native_realtime_playback_supported() -> Result<(), String> {
+    Err("Native OpenAI Realtime playback is not supported on this platform".into())
+}
+
+#[tauri::command]
+pub fn send_openai_realtime_spokesperson_runtime_event(
```
morgmart, line 163: 🤖 **P1 · Enforce runtime window ownership** (blocking)

The runtime stores owner_window and push_openai_realtime_spokesperson_audio checks the invoking WebviewWindow against it, but this newly registered session-scoped command and the other runtime/protocol mutation commands retrieve state using only the caller-supplied session_id. A session ID is an identifier, not proof that the invoking renderer owns that runtime, so the ownership boundary established at startup is not enforced consistently.

**User effect:** A stale or unrelated Berd window that has another conversation's session ID can inject voice-provider events, change or dismiss handoffs, alter voice state, or stop and release that conversation, causing incorrect transcripts, unexpected agent responses, or an interrupted voice session.

**Recommended fix:** Accept WebviewWindow on every session-scoped Tauri command and centralize lookup through a helper that verifies the caller label matches NativeRealtimeRuntime.owner_window before reading, mutating, stopping, or removing the entry. Keep trusted internal window-destruction cleanup on a separate non-command path.

**Test:** Create a runtime entry for session A owned by window owner, invoke every session-scoped mutation as window other with session ID A, and assert each call returns an ownership error without changing runtime state, protocol cursor, handoffs, settings, lifecycle, or membership; then assert the same calls from owner succeed.

morgmart (changes requested):
🤖 Automated code review

REQUEST_CHANGES. A fresh static Engineering review of the exact three-dot comparison f7ce2748cc465f23f4210f892b3cae83d778f93e...01279825793c5cb49ff6a7166ed0449d2d0abdfb found one blocking ownership-boundary defect. The new native OpenAI Realtime runtime records an owner window and checks it for microphone audio, but its other session-scoped Tauri commands authorize solely by session ID, allowing another renderer window to mutate or terminate a runtime it does not own. The project code-review and wes-review passes were applied distinctly. Final self-check covered every changed voice flow; accessibility; i18n/localization completeness; navigation and consent guards; async, error, never-completes, lifecycle, renewal, shutdown, interruption, and race behavior; test honesty; design-system and project rules; duplicate and overlapping findings; and concrete evidence plus user effect for blocking severity. Unsupported and overlapping claims were removed. Supplied GitHub evidence and all supplied review threads were inspected; ten check runs at the exact head completed successfully, while the combined commit status is pending with no commit statuses. Required checks still independently govern merge readiness. No local code, tests, builds, package managers, generated binaries, hooks, or project scripts were run.

src-tauri/src/commands/openai_realtime.rs
johnmatthewtennant: 🤖 Accepted and fixed in 5ded279453632bc1ea095ce604cafd07e1a8d704. All ten previously unguarded runtime/protocol commands now use the Tauri-injected WebviewWindow label and the shared ownership predicate before accessing or mutating an existing entry, including stop and release. Trusted window-destruction cleanup remains separate. Validation: just tauri-check, four focused command-module tests, and pre-push checks passed. The new regression tests exact owner-label matching; it is not a two-window IPC integration test. A fresh correctness review is running.

morgmart (approved):
🤖 Automated code review

APPROVE. A fresh static Engineering review of the full exact three-dot comparison f7ce2748cc465f23f4210f892b3cae83d778f93e...5ded279453632bc1ea095ce604cafd07e1a8d704 found no publishable findings. The previously reported runtime ownership defect is fixed at this head: every session-scoped native OpenAI Realtime command now receives the Tauri-injected invoking window and checks its label against the stored runtime owner before access or mutation, while trusted owner-window destruction cleanup remains separate. The exact-match ownership predicate has focused regression coverage. The resolved prior thread also contains a substantive human-author reply, so it would remain protected from duplicate publication independently of the fix. Project code-review and user wes-review were applied distinctly. Final self-check covered every changed Expert/Spokesperson and voice-conversation flow; accessibility; i18n/localization completeness; navigation and consent guards; async, error, never-completes, shutdown, interruption, renewal, recovery, lifecycle, queue, and race behavior; boundary ownership and input trust; test honesty; design-system and project rules; duplicate and overlapping findings; and concrete evidence plus user effect for blocking severity. No supported finding remains after consolidation. The supplied GitHub evidence is structurally valid for the exact base and head: Semgrep OSS, zizmor, DCO, the clean-room install, and Windows Rust checks succeeded, while Linux and macOS Rust checks and frontend checks were in progress and the combined status was pending. Required checks still independently govern merge readiness. No local code, tests, builds, package managers, generated binaries, hooks, or project scripts were run.

PR #309, opened 2026-09-09
feat(voice): simplify voice settings and defaults

johnmatthewtennant:
## Summary

- Clarifies direct coding-agent and voice-assistant modes, including latency, agent behavior, and Local/Cloud labels.
- Makes Apple speech the recommended direct-mode default, migrates legacy Parakeet preferences once, and presents voice and speed controls consistently across backends.
- Adds a full reset-to-defaults flow while preserving credentials and models, with atomic, concurrency-safe native settings persistence.

### Related issue

None found.

### Testing

- `just check`
- `pnpm vitest run src/features/voice-conversation`
- `just tauri-check`
- `just _tauri-cargo-unix test --lib settings`
- `pnpm build`
- `pnpm playwright test tests/e2e/voice-settings-visual.spec.ts`


src/features/voice-conversation/lib/voiceInputPreference.ts
```diff
@@ -98,6 +100,45 @@ export function setVoiceInputBackend(backend: VoiceInputBackend): void {
...
+      "complete"
+    ) {
+      return false;
+    }
+    const stored = normalizeStored(window.localStorage.getItem(STORAGE_KEY));
+    window.localStorage.setItem(
+      PARAKEET_RECOMMENDATION_MIGRATION_KEY,
+      "complete",
+    );
+    if (stored !== "parakeet") return false;
```
morgmart, line 126: 🤖 **P1 · Migration overwrites explicit choice** (blocking)

The migration treats every stored `parakeet` value as a legacy default and immediately replaces it with `macos` once Apple speech becomes available. The stored record contains no provenance that distinguishes an old default from a person explicitly choosing Parakeet, so this destructively rewrites valid saved intent during app startup.

**User effect:** People who deliberately chose Parakeet can reopen Berd and find speech input silently switched to Apple STT without their consent.

**Recommended fix:** Only migrate preferences that can be proven to be an implicit legacy default; otherwise preserve the stored selection and let the new recommendation affect only people without a saved choice.

**Test:** Add a discriminating migration test where Parakeet is an explicit saved choice and verify startup preserves it, while an actually identifiable legacy default still migrates once.

src/features/voice-conversation/ui/VoiceSettings.tsx
```diff
@@ -165,46 +218,83 @@ export function VoiceSettings() {
...
               );
 
+  const resetAllVoiceSettings = async () => {
+    const macSpeechAvailable = Boolean(
+      macSpeechSetup.status?.supported && macSpeechSetup.status.localeSupported,
+    );
+    setResetting(true);
+    setResetError(null);
+    try {
+      await Promise.all([
```
morgmart: 🤖 **P1 · Reset can partially apply** (blocking)

The three native resets run concurrently with `Promise.all`. If one fails after either of the others succeeds, the operation rejects before the renderer-owned preferences are reset, but the successful native changes remain persisted. The dialog reports one failed reset even though part of the reset has already taken effect, with no rollback or reconciliation.

**User effect:** After a reset error, people can be left with a mixed set of old and default voice settings even though Berd presents the reset as a single all-or-nothing action.

**Recommended fix:** Make the reset transactional across settings, or explicitly reconcile every setting to a deterministic final state before reporting success or failure so the visible result cannot be mixed.

**Test:** Add a failure-path test where one native reset rejects after another succeeds and verify the resulting saved and visible settings are either fully unchanged or fully reset, never partial.

morgmart (changes requested):
🤖 Automated code review

Engineering found two blocking state-consistency issues. The one-time migration can overwrite an explicitly chosen local transcription backend, and reset can partially apply before reporting failure. Supplied GitHub evidence shows all 11 captured check runs succeeded; the combined status endpoint was still pending with no statuses.

src/features/voice-conversation/lib/voiceInputPreference.ts
johnmatthewtennant: 🤖 This behavior is intentional under the product decision for this change: migrate every pre-existing Parakeet selection to Apple STT once, even though the old preference has no provenance, while keeping Parakeet available and preserving any Parakeet selection made after the migration marker is recorded. The migration tests cover that one-time boundary, so I am keeping this behavior as designed.

src/features/voice-conversation/ui/VoiceSettings.tsx
johnmatthewtennant: 🤖 Fixed in e0bf3bae. Reset is now one native transaction: it holds the Pocket and Apple settings locks, snapshots both stores, resets Pocket then Apple then OpenAI, and restores completed stores if a later backend fails. OpenAI runs last and persists before its in-memory state changes. Renderer-owned preferences are applied only after the native transaction succeeds. Added a native later-backend failure/rollback test and a renderer failure-path test; the full 432-test voice suite, `just check`, `just tauri-check`, and `just clippy` pass.

src/features/voice-conversation/ui/VoiceSettings.tsx
```diff
@@ -165,46 +218,80 @@ export function VoiceSettings() {
...
 
+  const resetAllVoiceSettings = async () => {
+    const macSpeechAvailable = Boolean(
+      macSpeechSetup.status?.supported && macSpeechSetup.status.localeSupported,
+    );
+    setResetting(true);
+    setResetError(null);
+    try {
+      await resetAllVoiceBackendSettings();
+      await setup.refreshSettings();
```
morgmart, line 229: 🤖 **P2 · Reset hides refresh failure** (blocking)

After the native reset succeeds, this flow awaits `setup.refreshSettings()`, but that helper catches status-read failures and resolves normally. The remaining preferences are then reset and the dialog closes even though the mounted Pocket status still contains its pre-reset voice and speed. No Pocket settings event or retry reconciles that stale state.

**User effect:** People can be told the reset completed, then reopen the Pocket controls in the same settings page and see old voice or speed values that no longer match what Berd will use.

**Recommended fix:** Make the post-reset refresh propagate failure, or update the mounted Pocket status to the known defaults and retry reconciliation, so the reset cannot report success while the controls remain stale.

**Test:** Add a reset-flow test where the native transaction succeeds but the Pocket status refresh rejects, and verify the dialog does not report a completed reset with stale controls; then verify a successful retry or deterministic local update shows the defaults.

morgmart (changes requested):
🤖 Automated code review

Engineering found one blocking state-consistency issue: a failed post-reset Pocket status refresh is swallowed, so the reset reports success while mounted controls retain stale values. The earlier partial-reset issue is fixed by the native transaction. The prior migration concern is suppressed because its resolved thread has a substantive human reply. Supplied GitHub evidence is structurally valid; four captured checks passed and four were still in progress, so required checks continue to govern merge readiness.

Deterministic publication result: 1 blocking and 0 non-blocking inline finding(s) publishable; 1 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/voice-conversation/ui/VoiceSettings.tsx
johnmatthewtennant: 🤖 Fixed in 09f6fa63. Pocket settings refresh now propagates status-read failures after updating the visible error state, so reset cannot close and report success with stale mounted controls. The regression test exercises a failed refresh followed by a successful retry and verifies renderer preferences remain unchanged until reconciliation succeeds. Validation: all 434 voice-conversation tests pass; `just check`, pre-push Tauri checks, and clippy pass.

morgmart, line 229: 🤖 **P2 · Reset leaves Apple controls stale** (blocking)

The native transaction resets Siri voice and speed, but this success path only refreshes Pocket status. OpenAI refreshes through its emitted backend event, while the Siri hook listens only for a renderer-local `berd:siri-voice-settings-changed` event that this transaction never dispatches. When Apple TTS is already selected, setting the same output backend does not re-enable or remount the hook, so its existing status continues to show the pre-reset voice and speed.

**User effect:** People who reset while using Apple TTS can see their old Siri voice and playback speed remain selected on the settings page even though Berd has already restored different defaults.

**Recommended fix:** After the native transaction succeeds, explicitly refresh the mounted Siri setup or dispatch its settings-changed signal and await reconciliation before closing the reset dialog.

**Test:** Add a reset-flow test starting with Apple TTS selected and non-default Siri voice and speed, then verify the mounted controls show the native defaults after reset without requiring navigation or window focus.

morgmart (changes requested):
🤖 Automated code review

Engineering found one blocking state-consistency issue: reset refreshes Pocket and OpenAI state but leaves mounted Apple voice controls stale when Apple TTS was already selected. The earlier native partial-reset and Pocket refresh issues are fixed. The migration concern is suppressed because its resolved thread has a substantive human reply. Supplied GitHub evidence is structurally valid; seven captured checks passed and three were still in progress, so required checks continue to govern merge readiness.

Deterministic publication result: 1 blocking and 0 non-blocking inline finding(s) publishable; 1 duplicate(s) suppressed; 0 blocking screenshot-evidence requirement(s) in this review body.

src/features/voice-conversation/ui/VoiceSettings.tsx
johnmatthewtennant: 🤖 Fixed in 4bdb6f5e. The reset flow now explicitly awaits a mounted Siri settings refresh after the native transaction and before applying renderer preferences or closing the dialog. The hook exposes a refresh operation that propagates read failures, and regressions cover Apple TTS already selected, fail-closed behavior, retry, and the hook contract. Validation: 58 focused tests, all 436 voice-conversation tests, `just check`, and the full pre-push Tauri/clippy hooks pass.

morgmart (approved):
🤖 Automated code review

Engineering found no new publishable issues in the exact PR comparison. The native reset transaction, Pocket refresh failure propagation, and mounted Apple settings reconciliation address the prior reset findings with discriminating coverage. The migration concern remains suppressed because its resolved thread has a substantive human reply documenting the intentional behavior. Supplied GitHub evidence is structurally valid; nine captured checks passed and one was still in progress, so required checks continue to govern merge readiness.

PR #311, opened 2026-09-09
fix(voice): preserve first streamed TTS phrase

johnmatthewtennant:
**Category:** fix
**User Impact:** Streamed voice replies no longer pronounce an isolated first letter or word, pause, and then continue after the coding agent uses a tool or as its final ACP update settles.

**Problem:** Two ordering boundaries could split the opening phrase. Tool-only assistant phases were recorded only after they owned a speech utterance, so the first later text delta looked like a newly crossed tool boundary and was flushed alone. Separately, ACP may resolve a prompt just before dispatching its final already-read update, allowing completion observers to finish an incomplete fragment.

**Solution:** Record the observed tool count for every assistant message, while explicitly tracking only completion events already handled. At the producer boundary, keep prompt ownership for one event-loop turn before marking the response complete so the final ACP update reaches the transcript and shared TTS stream first. The fix applies to Siri, Pocket, and OpenAI TTS through the shared native speech path.

**Reviewer-reproducible examples:**
- `pnpm vitest run src/features/voice-conversation/lib/nativeAssistantSpeech.test.ts -t "does not flush the first text chunk when a tool preceded it"` simulates a completed tool-only phase followed by the first streamed word and verifies no standalone flush occurs.
- `pnpm vitest run src/features/chat/lib/sendCore.test.ts -t "applies the final ACP update before marking the response complete"` simulates prompt resolution immediately before a queued final text update and verifies the complete phrase is present before completion is published.

### DIFF
--- a/src/features/chat/hooks/__tests__/useChat.test.ts
+++ b/src/features/chat/hooks/__tests__/useChat.test.ts
@@ -1,4 +1,4 @@
-import { act, renderHook } from "@testing-library/react";
+import { act, renderHook, waitFor } from "@testing-library/react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { useAgentStore } from "@/features/agents/stores/agentStore";
 import { useChatStore } from "../../stores/chatStore";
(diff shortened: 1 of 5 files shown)


morgmart (approved):
🤖 Automated code review

APPROVE. The completed Engineering review of the exact prepared three-dot comparison found no publishable changed-line defects, duplicate threads, or escalation needs. The completion-order and tool-boundary fixes are covered by discriminating regression tests. Supplied GitHub check evidence for the exact head SHA is structurally valid and successful; required checks still independently govern merge readiness.

PR #318, opened 2026-09-10
fix(voice): keep Realtime calls alive after cancellation races

johnmatthewtennant:
## Summary

Fixes an OpenAI Realtime call ending with “Cancellation failed: no active response found” when a response finishes before its cancellation reaches the provider. Coordinator cancellations carry tracked event IDs, and the matching no-active-response error is consumed before provider events reach the host's fatal-error reducer. Unrelated errors and other cancellation failures remain visible.

Runtime regressions reproduce the cancellation failure on the base commit and verify that a subsequent speech-start event is processed after the fix. They also preserve failures for uncorrelated errors and unexpected cancellation failures. This coverage uses a local WebSocket test provider; live OpenAI audio and the macOS call UI remain unverified.

### DIFF
--- a/src-tauri/crates/berd-voice/src/openai_spokesperson.rs
+++ b/src-tauri/crates/berd-voice/src/openai_spokesperson.rs
@@ -584,7 +584,16 @@ async fn run_inner(
             }
             command = commands.recv(), if shutdown_deadline.is_none() => {
                 match command {
-                    Some(SpokespersonCommand::Provider(event)) => {
+                    Some(SpokespersonCommand::Provider(mut event)) => {
+                        if event.get("type").and_then(serde_json::Value::as_str) == Some("response.cancel") {
+                            if let Some(response_id) = string(&event, "response_id").map(str::to_owned) {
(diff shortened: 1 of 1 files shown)


johnmatthewtennant:
🤖 User-observed Realtime lifecycle bug from live testing: ending an Expert-Spokesperson call displayed `Spokesperson runtime is unavailable`. This appears to be a clean-shutdown ordering issue rather than evidence that the active call failed: the renderer intentionally enters `stopping` while final runtime events are still draining, so a late runtime failure can be surfaced after shutdown has begun. Please ensure intentional stop suppresses or classifies that late failure without hiding genuine failures before shutdown.

morgmart (approved):
🤖 Automated code review

Engineering review found no publishable findings in the exact PR comparison. The changed voice-call lifecycle, cancellation correlation, transcript discard handling, shutdown ordering, and regression tests are internally consistent under static inspection. Supplied GitHub checks are complete and successful. Recommend APPROVE.

